PR #18, opened 2022-03-27
Fix #17 - Use pomless where possible

### LINKED ISSUE #17: Use pomless where possible
Currently p2 build uses pom base tycho build even for simple bundle-projects without any special configuration.

I'd like to suggest using pomless instead where suitable so for example one only needs to change the version in the manifest.

laeubi:


mickaelistria:
Thank you!

Issue #28, opened 2022-04-04
Update to Sat4J 2.3.6


akurtakov:
There is new version after years which we would better use. https://gitlab.ow2.org/sat4j/sat4j/-/releases/2_3_6

mickaelistria:
Done with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/commit/128cc023518716dea9abb006e68a26efd223e57b

PR #30, opened 2022-04-09
Touch features to use newer sat4j (#28)

mickaelistria:


mickaelistria:
This causes 1 test to fail because of a parsing exception (was working with sat4j 2.3.5, failing with sat4j 2.3.6).

The opb file is
```
* #variable= 6 #constraint= 3  
*


explain:  x1 x2 x3 x4 x5 ;

* P 1.0.0 requires either PP 1.0.0 
-1 x2 +1 x1 >= 0;
* 1223597333557 0.0.0.1223597333557 requires either P 1.0.0 
-1 x3 +1 x2 >= 0;
 +1 x3 = 1;
```
and the error is `Parsing Error line 5, java.lang.NumberFormatException: For input string: "plain"`
@danielleberre @sat4j could it be a regression in sat4j, or maybe the opb file has been invalid and needs an adjustment?

danielleberre:
I will take a look. 

explain is a specific syntax we used for p2, when we used file based communication between p2 and sat4j.

danielleberre:
@mickaelistria Here is a fix for this testcase:
https://gitlab.ow2.org/sat4j/sat4j/-/commit/8461e3723b05df0717f02210d2ed9e205eed86c8

Note that this reader (OPBEclipseReader) was used in the first release of p2 including Sat4j where communication was based on files. It is no longer used now in p2 codebase.

Maybe @prapicault can comment on that.

I would rather remove that old test (unless some compatibility with Eclipse Ganymede is still needed) that updating to a 2.3.6.1 bugfix release of sat4j (not available yet, but I can release it if needed).

mickaelistria:
I'll remove this test as you suggest.

mickaelistria:
Replaced by #33

PR #31, opened 2022-04-09
Remove inclusion of sat4j in feature.xml (#28)

mickaelistria:
sat4j is already included transtively by Require-Bundle for
org.eclipse.equinox.p2.director. So we don't need to add it
explicitly, and that will facilitate further verison upgrades
as the version wouldn't be statically replaced.

### DIFF
--- a/features/org.eclipse.equinox.p2.core.feature/feature.xml
+++ b/features/org.eclipse.equinox.p2.core.feature/feature.xml
@@ -112,20 +112,6 @@
          unpack="false"/>
 
    <plugin
-         id="org.sat4j.core"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.sat4j.pb"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
          id="org.eclipse.equinox.frameworkadmin"
          download-size="0"
          install-size="0"
--- a/features/org.eclipse.equinox.p2.sdk/feature.xml
+++ b/features/org.eclipse.equinox.p2.sdk/feature.xml
@@ -470,20 +470,6 @@
          unpack="false"/>
 
    <plugin
-         id="org.sat4j.core"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
-         id="org.sat4j.pb"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
-         unpack="false"/>
-
-   <plugin
          id="org.eclipse.equinox.p2.operations"
          download-size="0"
          install-size="0"


HannesWell:
In the past I also wondered if it is necessary or only good practice to add Plugin requirements to a Feature. From all I knew I also came to the conclusion you expressed on the mailing-list, that it should be the duty of the dependency-manager (in this case P2) to install requirements as needed and that there is no need to add those dependencies to a feature. So I think it is good to remove this duplication.

But to add dependencies also to the Eclipse SDK's P2-repository the [`includeAllDependencies` property of the `tycho-p2-repository-plugin`](https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-repository-plugin/assemble-repository-mojo.html#includeAllDependencies) has to be set to true. Otherwise only (transitively) `included` plugins and features are added.
The default value of this property is false and I could not find any occurrence of it in the [eclipse.platform.releng.aggregator/eclipse-platform-parent/pom.xml](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse-platform-parent/pom.xml), so I wonder if this property has to be set?

The same applies for https://github.com/eclipse-pde/eclipse.pde.ui/pull/24.

mickaelistria:
> But to add dependencies also to the Eclipse SDK's P2-repository the [includeAllDependencies property of the tycho-p2-repository-plugin](https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-repository-plugin/assemble-repository-mojo.html#includeAllDependencies) has to be set to true. Otherwise only (transitively) included plugins and features are added.
The default value of this property is false and I could not find any occurrence of it in the [eclipse.platform.releng.aggregator/eclipse-platform-parent/pom.xml](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse-platform-parent/pom.xml), so I wonder if this property has to be set?

Since there are .product file used to source this repository, their transitive deps are included anyway (it's more like it `includeAllDependencies` is automatically true for .product files). So no extra setting is needed.

HannesWell:
> Since there are .product file used to source this repository, their transitive deps are included anyway (it's more like it `includeAllDependencies` is automatically true for .product files). So no extra setting is needed.

AFAICT this is not the case, even for products. But I only know this because exactly I had this problem with our product. Users were not able to update because the product's repo did not contain all dependencies and the included plug-ins had requirements for newer versions.

Maybe this is set somewhere else? Actually I would be surprised if all dependencies are listed in features, so expected it is set already. Any way, I just wanted to mention, in case there is a problem with it. :)

mickaelistria:
OK, so maybe I misinterpreted what's happening for products or looked at a wrong repo.
Would you be willing to open a PR that sets `includeAllDepndencies` to the main p2 repo? If you prefer, I can take care of it.

HannesWell:
> OK, so maybe I misinterpreted what's happening for products or looked at a wrong repo. Would you be willing to open a PR that sets `includeAllDepndencies` to the main p2 repo? If you prefer, I can take care of it.

Isn't P2 just build with the entire SDK? I did not know there is a dedicated repo for P2.
Maybe it's quicker if you do take care of it? I think you have the better overview.

mickaelistria:
> Isn't P2 just build with the entire SDK? I did not know there is a dedicated repo for P2.

Sorry, I meant SDK indeed.
I'll take care of the change on the aggregator.

Issue #36, opened 2022-04-16
Some FileReader.CancelHandler jobs never terminate


merks:
CancelHandler jobs are not always terminated:

https://github.com/eclipse-equinox/p2/blob/4cc5059736d6ee6d17f2d2acc86f2df98c3bbc1f/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/FileReader.java#L150

There's a strange `done` flag that is never set to true and is used to keep the loop running.

![image](https://user-images.githubusercontent.com/208716/163677560-ba01aa08-274a-479f-9290-0954d0e60778.png)

If you stop one of these zombie jobs in its loop, you can see that the containing FileReader instance's cancelJob has been set to null:

![image](https://user-images.githubusercontent.com/208716/163677968-dc100a3e-def2-4adc-bc2e-24fae0f3034b.png)

So the CancelHandler job must have been canceled by this line:

https://github.com/eclipse-equinox/p2/blob/4cc5059736d6ee6d17f2d2acc86f2df98c3bbc1f/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/FileReader.java#L402

So I will change the "done" test to test if the `cancelJob` field is still a reference the CancelHandler instance.

PR #37, opened 2022-04-16
Some FileReader.CancelHandler jobs never terminate #36

merks:
Terminate the job's loop when the enclosing FileReader no longer has a reference to the CancelHandler job.


merks:
The maven.yml wrong and needs to be fixed.  

The jenkins verification build don't appear to work because the tests rely on other parts that don't appear to exist but that seems unrelated to any changes I've made:
```
Need to set the "org.eclipse.equinox.p2.reconciler.tests.platform.archive" system property with a valid path to the platform binary drop or copy the archive to be a sibling of the install folder.
The file '/home/jenkins/agent/workspace/p2_PR-37/bundles/org.eclipse.equinox.p2.tests/target/../../org.eclipse.equinox.p2.tests.reconciler.product/target/products/org.eclipse.equinox.p2.reconciler-linux.gtk.x86_64.tar.gz' does not exist
```

PR #38, opened 2022-04-16
Fix maven.yml to actually invoke mvn.

merks:


merks:
I aborted the ci-build because this commit only fixes the maven.yml so no point in doing the other build.

merks:
The build gets farther, but still not quite configured correctly.

PR #42, opened 2022-04-20
Fix #6 - Deprecate Transport.download(URI, OutputStream, long,IProgressMonitor)

### LINKED ISSUE #6: Deprecate Transport.download(URI, OutputStream, long, IProgressMonitor)
It seems that the method `org.eclipse.equinox.internal.p2.repository.Transport.download(URI, OutputStream, long, IProgressMonitor)` is actually never called from P2.

As `Transport` is an abstract class I suggest to do the following:

1. mark the method deprecated
2. remove abstract modifier and implement it by throwing `OperationNotSupportedException`.

This would make implementing alternative transport much more easier.

laeubi:


bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/Transport.java
```diff
@@ -28,21 +29,28 @@ public abstract class Transport {
...
 	 * is also set on the target. An IStateful target is updated with status even if this methods
 	 * throws {@link OperationCanceledException}.
-	 * 
+	 *
 	 * @returns IStatus, that is a {@link DownloadStatus} on success.
 	 * @param toDownload URI of file to download
 	 * @param target OutputStream where result is written
 	 * @param startPos the starting position of the download, or -1 for from start
 	 * @param monitor where progress should be reported
 	 * @throws OperationCanceledException if the operation was canceled.
```
mickaelistria, line 38: We need to comment in javadoc with `@deprecated <alternative>` describing that this method isn't used anymore and what other code should be used/implemented instead.

mickaelistria (changes requested):
Just need some more doc.

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/Transport.java
laeubi: done

mickaelistria (approved):

laeubi:
@merks @vogella do you like to review?

laeubi:
No need to rush :+1:

merks:
@laeubi 

Yes this is easy to accommodate and I agree that removing functionality that is complex to implement and isn't even used makes sense.

laeubi:
Just approve/merge it when your fine with that :-)

vogella:
LGTM, if you want to have the option to delete the API in 3 years you also need to add the planned deletion to the migration guide.

@akurtakov would it be possible to generate the API to be deleted automatically so that we could get

vogella:
> > if you want to have the option to delete the API in 3 years you also need to add the planned deletion to the migration guide
> 
> This is not API this is an internal class.

Great

akurtakov:
> @akurtakov at some point it be possible to generate the API to be deleted automatically so that we could get rid of this manual step.

Some list is already visible http://help.eclipse.org/2022-03/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Fdeprecated-list.html&anchor=forRemoval not easy to guess when it's fine to remove it.

vogella:
Please merge Jenkins build was succesfull

laeubi:
I'll leave this to @merks (I can't merge it anyways) as he liked to make some additional checks.

vogella:
Looks save to me. @merks any additional concerns? If not I plan to merge early next week.

merks:
As I mentioned in https://github.com/eclipse-equinox/p2/pull/42#issuecomment-1103732728 I'm can implement this in a way that works well for old a new versions of p2.

Issue #40, opened 2022-04-20
Ignore pack200 requests


akurtakov:
As per https://www.eclipse.org/lists/cross-project-issues-dev/msg18275.html:
* Stop dealing with pack200 at all
* Clean up internals from pack200 
* No API change but deprecate and mark for removal if not done already and make the calls *noop*

akurtakov:
@laeubi Thanks for the reminder!

PR #43, opened 2022-04-20
Ignore pack200 entirely

akurtakov:
Internal implementation removed but API preserved as NO-OP.

Fixes https://github.com/eclipse-equinox/p2/issues/40


akurtakov:
@laeubi any idea what goes wrong with the build?

laeubi:
The github action seem not to be configured correctly,  the jenkins want some version bumps.

PR #44, opened 2022-04-20
Add toolchains that maps the github java

laeubi:


laeubi:
@akurtakov this should fix the GH action but I have no idea why it complains about versions not bumped...

laeubi:
As it at least now not fails with the toolchain error it might still eb a good idea to merge this already?

merks:
It's a further improvement indeed.  Thanks. 👍

Issue #45, opened 2022-04-20
The use of sat4j by p2 leaks solver instances via threads spawned by sat4j


merks:
I had the following horrible workaround in my marketplace  report generator:

![image](https://user-images.githubusercontent.com/208716/164242386-3953319b-7cf3-479f-9ddd-4d9dfc9481e1.png)

It would of course be much better to solve this properly in p2, especially given that the problem appears to have come back with the new sat4j version.

Here's are examples of such leaked objects:

![image](https://user-images.githubusercontent.com/208716/164243567-16618e23-1067-45b5-afbb-a51a0e295da4.png)

![image](https://user-images.githubusercontent.com/208716/164243750-03ed70f9-9129-46bd-9aef-1c4e8a48f0c9.png)

mickaelistria:
Would it be worth making the `Projector implements AutoCloseable` so we can later use
```java
try (Projector projector = ....) {
 ...
}
```
?

HannesWell:
> Would it be worth making the `Projector implements AutoCloseable` so we can later use

This was discussed in the linked PR: https://github.com/eclipse-equinox/p2/pull/46#discussion_r857065215

laeubi:
Yes it was discussed there and I think we even don't need to close the projector at all but the `DependencyHelper` in this specific method call, but haven't yet found the time to look into this and since everyone seems fine I'm more focused on other "issues" ,,,

PR #46, opened 2022-04-20
The use of sat4j by p2 leaks solver instances via threads spawned by sat4j

merks:
Add a close method to Projector which resets the dependency helper,
which ensures that any threads are canceled.

Call the close method after the usage of the Projector has completed.

https://github.com/eclipse-equinox/p2/issues/45

### DIFF
--- a/bundles/org.eclipse.equinox.p2.director/src/org/eclipse/equinox/internal/p2/director/Projector.java
+++ b/bundles/org.eclipse.equinox.p2.director/src/org/eclipse/equinox/internal/p2/director/Projector.java
@@ -1083,4 +1083,11 @@ public class Projector {
 	public void setUserDefined(boolean containsKey) {
 		userDefinedFunction = containsKey;
 	}
+
+	public void close() {
+		if (dependencyHelper != null) {
+			dependencyHelper.reset();
+			dependencyHelper = null;
+		}
+	}
 }
--- a/bundles/org.eclipse.equinox.p2.director/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.director/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.director;singleton:=true
-Bundle-Version: 2.5.200.qualifier
+Bundle-Version: 2.5.300.qualifier
 Bundle-ClassPath: .
 Bundle-Activator: org.eclipse.equinox.internal.p2.director.DirectorActivator
 Bundle-Vendor: %providerName
--- a/bundles/org.eclipse.equinox.p2.director/src/org/eclipse/equinox/internal/p2/director/SimplePlanner.java
+++ b/bundles/org.eclipse.equinox.p2.director/src/org/eclipse/equinox/internal/p2/director/SimplePlanner.java
@@ -459,6 +459,7 @@ public class SimplePlanner implements IPlanner {
 			case CANCEL: {
 				IProvisioningPlan plan = engine.createPlan(profile, context);
 				plan.setStatus(s);
+				projector.close();
 				return plan;
 			}
 
@@ -469,6 +470,7 @@ public class SimplePlanner implements IPlanner {
 						|| Boolean.parseBoolean(context.getProperty(EXPLANATION))))) {
 					IProvisioningPlan plan = engine.createPlan(profile, context);
 					plan.setStatus(s);
+					projector.close();
 					return plan;
 				}
 
@@ -484,6 +486,7 @@ public class SimplePlanner implements IPlanner {
 
 				IProvisioningPlan plan = engine.createPlan(profile, context);
 				plan.setStatus(plannerStatus);
+				projector.close();
 				return plan;
 			}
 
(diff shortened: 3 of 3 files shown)


mickaelistria:
Thanks Ed. Is this new with latest sat4j or was it already an issue before that?

merks:
It was already an issue before.  It generally only happens when an install fails and detailed explanation is needed.  But in general org.sat4j.minisat.core.Solver.isSatisfiable(IVecInt, boolean) method can create a timer:

```
        boolean firstTimeGlobal = false;
        if (this.timeBasedTimeout) {
            if (!global || this.timer == null) {
                firstTimeGlobal = true;
                this.undertimeout = true;
                TimerTask stopMe = new TimerTask() {
                    @Override
                    public void run() {
                        Solver.this.undertimeout = false;
                        synchronized (Solver.this) {
                            if (Solver.this.timer != null) {
                                Solver.this.timer.cancel();
                                Solver.this.timer = null;
                            }
                        }
                    }
                };
                this.timer = new Timer(true);
                this.timer.schedule(stopMe, this.timeout);

            }
        } else {
```
and this cleans it up, if it was created.
```
    public void reset() {
        if (this.timer != null) {
            this.timer.cancel();
            this.timer = null;
        }
```

I believe the inner TimerTask class is the source of the leak; certainly the hacky way to call reset did fix the problem previously. This problem was only really noticeable in the marketplace tester where I test well over 1000 marketplace listings each against more than 20 Eclipse versions.

But with the switch to the new sat4j, there appears to be perhaps another/different leak so I decided best to fix it properly in p2, i.e., always reset the solver, not only when its was used to create a detailed explanation of failure. and hope that the marketplace tester works again.   We'll see later today...

bundles/org.eclipse.equinox.p2.director/src/org/eclipse/equinox/internal/p2/director/Projector.java
```diff
@@ -1083,4 +1083,11 @@ private void rememberHostMatches(IInstallableUnitFragment fragment, List<IInstal
 	public void setUserDefined(boolean containsKey) {
 		userDefinedFunction = containsKey;
 	}
+
+	public void close() {
```
laeubi, line 1087: Maybe projector should implement Autocloseable then?

merks: If you look at where close is called in a several places for the one place that a Projector is created, it wouldn't really help make anything simpler. at least in the p2 code itself.

laeubi: But why is close called in seveveral places and not once after the user is done? AutoColeable has the advantage that eclipse warns you if you do not close it after use.

merks: The complete code is there for your inspection to answer questions like this yourself...  

The Projector is used in a rather ugly way.   In some cases the Projector is used to create and return a plan, at which point it's no longer used, and can be closed.  In other cases it's returned from the method (whose return type is Object) and must be closed by the caller.  I haven't properly covered all the uses of the call to getSolutionFor either, just the path followed by the marketplace tester.  And the marketplace tester job still dies with the newer version of sat4j which requires further investigation, but testing a job that runs for hours that maybe has a leak is not trivial...

laeubi: >  In other cases it's returned from the method (whose return type is Object) and must be closed by the caller. 

Maybe it then cloud return an Closable instead to indicate the to the caller then or convert it? This are just some ideas without having to investigate this in details, but having a "close" method that is arbitrary called really seems like code-smell ... especially if that not really "closes" but seem to reset some internal state...

If I search for Projector#dependencyHelper it seems only be used inside the projector and there is only one public method referencing this, so I think one should remove the field, and pass the object to subcalls as a parameter instead and always reset it (or just stop the job if possible) before returning would prevent the instance from being hard referenced here.

laeubi: e.g. `DependencyHelper#stopSolver()` seems a possible candidate we should always call in a finally block.

merks: Yes, the code smells.   But I really have so many other things to keep me busy that it's hard to keep track of them all.  And that's not to mention the open bugs in p2 that, in my opinion, are generally all more important than removing this particular smell, at the risk of adding this this list:

https://bugs.eclipse.org/bugs/buglist.cgi?bug_status=UNCONFIRMED&bug_status=NEW&bug_status=ASSIGNED&bug_status=REOPENED&classification=Eclipse%20Project&component=p2&list_id=21260880&order=bug_id%20DESC&product=Equinox&query_format=advanced

Perhaps you could focus on one of those because then some real live human being will see some real actual improvement in their lives.

laeubi: If my goal would be to make people happy I won't review code ;-)

It seems that this could really cause some serious problem making "real live human being" unhappy, so I think its worth to have a sustained fix for this, of course you could simply ignore that and wait until you or some else run into the same problem because there is somewhere another `close()` missing ...

Issue #23, opened 2022-03-29
release p2 more often?


laeubi:
I found the recent discussions at `tycho`/ `oomph` and sharing/contribution code changes often has one hindering, that it is too inflexible because both projects don't want to wait for the next release.

What I'm wondering is: As far as I know there is nothing forcing p2 *not* leasing more often, so why not release every month/ week / day or even on every commit or on demand?

What would be required for this would be an automatic deployment workflow e.g we can deploy to eclipse-maven (as we deploy snapshots) automatically and deploy to maven-central the sim-rel.

Maybe p2 would even be the perfect 'reactor' to get experience with an automatic maven deployment (in contrast to the sim-rel based one) and a more flexible deployment workflow?

mickaelistria:
>  even on every commit 

That's what I-Builds are about.

merks:
Indeed, Oomph uses integration builds....

laeubi:
> > even on every commit
> 
> That's what I-Builds are about.

I don't think that I builds are suitable here:

1. I-Builds are moving targets, they vanish once in a while and are therefore lost after a short while
2. They don't offer maven-artifacts
3. I once asked for using the I-Build but @akurtakov has explained some issues why we can't use them (probably because of 1.) as then Tycho can't perform release

merks:
Keep in mind too that p2 depends on other things.  During the development cycle it might well be modified to depend on things that are specific and new in its dependencies.  So if p2 were to release, it wouldn't necessarily work with the released versions of its dependencies.  This type of problem is of course why we have SimRel in general and why often downstream projects don't wont to depend on new things because then they can't be installed anywhere expect the latest release...

laeubi:
> Keep in mind too that p2 depends on other things. During the development cycle it might well be modified to depend on things that are specific and new in its dependencies.

Then one goal might be to release that dependencies also more often :-)

In fact we don't need to promote these "mid-releases" in a way that they become available in regular eclipse products, but for tycho or eclipse-installer, who cares what version/release they use internally to perform their work as long as I can press "the download button" and it works?

merks:
By p2's dependencies we're specifically talking the Eclipse UI, all of its dependencies, including the rest of Equinox.   So we're talking about most of what we currently release 4 times per year.  That seems infeasible...

laeubi:
Of course we can concentrate either of two things:

1. Find as much things why it /can't/ happen
2. Try it out and see if it /can/ happen

[...]

PR #47, opened 2022-04-22
Use even less pom.xml files

HannesWell:
This enhances the fix of https://github.com/eclipse-equinox/p2/issues/17.
@laeubi is there a certain reason you skipped these poms?

Like https://github.com/eclipse-equinox/p2/pull/18 it changes the `groupId` of multiple plug-ins from `org.eclipse.equinox` to `org.eclipse.equinox.p2`. But when looking at Maven-Central (e.g. [o.e.equinox.p2.director.app](https://mvnrepository.com/artifact/org.eclipse.platform/org.eclipse.equinox.p2.director.app/1.2.100
)) it seems that those plug-ins are published as part of `groupId` `org.eclipse.platform` that is completely unrelated to the groupIds used directly in the poms. And IIRC the publication process of those Eclipse-Platform Plug-in to Maven-Central changes the so groupdId so the change here should not be a problem.
@mickaelistria @akurtakov is that correct?

Furthermore this removes the unused org.eclipse.equinox.p2.tests-parent (at least it looks like that).


laeubi:
If we wan't a different group id you should be able to define this in the root pom.xml as pomless items get them from the parent.
Beside this, it is possible to override them in a build.properties.

About the deployment see #23

akurtakov:
Publishing to Maven central is totally different thing IIRC and doesn't use the pom.xml files from the repo but rather generates new ones. @sravanlakkimsetti knows the detail but for me this is definetely a topic that needs improvement and "real" pom.xml files are used. Unfortunately we (releng people) never found time to even look into it.

laeubi:
@akurtakov I suggested to investigate this in the context of #23 but it got not much support so I postponed it from my (long) task list for now, let me know if it becomes relevant in the future...

akurtakov:
> @akurtakov I suggested to investigate this in the context of #23 but it got not much support so I postponed it from my (long) task list for now, let me know if it becomes relevant in the future...

Way things are published to Maven Central is a pre-req for faster release cycle together with many other things. It's not like smth I (and probably others) don't support but current processes are too cumbersome to allow it. Whenever we manage to get less manual work per release we can reconsider faster release cycles. 
Note that deploying I-builds content to repo.eclipse.org as SNAPSHOT versions is totally not blocked by anything right now if someone wants to do it. We just don't want not releases stuff at Maven Central.

laeubi:
> It's not like smth I (and probably others) don't support but current processes are too cumbersome to allow it.

That's why I suggested we change this ;-)

> Note that deploying I-builds content to repo.eclipse.org as SNAPSHOT versions is totally not blocked by anything right now if someone wants to do it.

That would be a first step yes but even this was not consent as we already have the I-builds update-sites... anyway I might take a look at it sometimes later...

HannesWell:
I now know why you left the simple pom.xml files that are actually generated by Tycho-build on the file in the repo:
Those project also contain a product-file and the project is then considered a 'product' project and not a Eclipse-Plugin project. So without refactoring the project structure those pom.xml files have to stay.

Besides that I think this PR is ready. Any concerns/objections?

laeubi:
Yes there is a ticket on tycho: 
- https://github.com/eclipse/tycho/issues/36
but I'm not sure if this projects are just examples and should actually be build or not...

HannesWell:
I thought about that issue, too. But I'm relatively sure those products are just examples and are not meant to be build, at least they where not build in the past.
However if https://github.com/eclipse/tycho/issues/36 is resolved one day they mojos to build the product could be disabled to suppress product build and to only build the plug-in to keep the current behavior.
Building a product that is not published is just a waste of build-time. On the other hand, it would ensure that those products always work.

laeubi:
another option would be to move those example products in a separate folder (either in the plugin or a different source-tree) as pomless only looks at the root for product files.

HannesWell:
> another option would be to move those example products in a separate folder (either in the plugin or a different source-tree) as pomless only looks at the root for product files.

I think moving them to a new project seems to be a bit too much just for this purpose, but moving them into a subfolder of the Plug-in project is a good suggestion.

However I wonder how these products are supposed to be used. The projects I have checked do not bin.include the product-file. So is it intended that an interested user checks out the p2-sources?

laeubi:
No idea :-)

HannesWell:
Does anybody know more about these example products or can ping somebody who does.

From scamming them and the associated launch configs I think would benefit from being reworked to feature-based.
I can have a look at that later but I would like to avoid unnecessary work in case they are obsolete.

laeubi:
> I think moving them to a new project seems to be a bit too much just for this purpose,

I mean we could have one "example/products" folder where all products are moved to.

vogella:
Example which are not documented nor easy to understand should be retired, I do this all the time, if I come across such outdated material.

HannesWell:
I created https://github.com/eclipse-equinox/p2/issues/52 to discuss the future of those examples (hopefully more people notice that compared to if we discuss this in an unrelated PR :) ).
Lets keep this at the poms.

Actually I think this is ready, but there are test-failure. I looks like they are unrelated but I would like to have them fixed before we submit this.

HannesWell:
Btw. what do you think about disabling the GH workflow to build p2? It is constantly failing, which is IMHO disturbing and adds not value since we (or at least I) only relay on Jenkins therefore.

akurtakov:
Please disable. If someone is interested to figure why it fails and fixes it, he/she can reenable.

HannesWell:
> Please disable. If someone is interested to figure why it fails and fixes it, he/she can reenable.

Yes, fully agree. Disabled it.

laeubi:
It still would eb interesting to find out why it is failing, last time I checked it complained about missing version bumps, so whats different here with the jenkins?

HannesWell:
> It still would eb interesting to find out why it is failing, last time I checked it complained about missing version bumps, so whats different here with the jenkins?

Sure, that would be interesting but at least I am happy with one working build. 
That was still the error the last time I looked into the results. GH actions were also unreliably in m2e IIRC so maybe there is a more fundamental difference.

PR #49, opened 2022-04-23
Delete code of abandoned p2 plug-ins

HannesWell:
Like in https://github.com/eclipse-equinox/equinox.bundles/pull/17 I want to propose to delete the code of the following p2 plug-ins that are not build and published anymore (I don't know since when):
- `org.eclipse.equinox.p2.artifact.checksums.bouncycastle`
- `org.eclipse.equinox.p2.artifact.optimizers`
- `org.eclipse.equinox.p2.artifact.processors`
- `org.eclipse.equinox.p2.sar`
- `org.eclipse.equinox.p2.tests.optimizers`
- `org.eclipse.equinox.p2.testserver`

If any project is still relevant I suggest to include it in the Jenkins build again to ensure it continues to function (or at least compiles).


vogella:
Would be nice to finish this one. Deletion is always good. :-)

HannesWell:
I'm just waiting for approval for this one, because I don't have a deep insight into p2 if any of those projects could be useful one day. Especially for `org.eclipse.equinox.p2.testserver` I'm uncertain.

Maybe someone can judge? @tjwatson, @akurtakov, @mickaelistria, @laeubi?

laeubi:
I have no clue what these are about, but if they are not build or published I think no one uses them and source control systems are made for the case you later want to go back, so even if someone is later missing this it could be restored and/or migrated to an appropriate place.

vogella:
> I have no clue what these are about, but if they are not build or published I think no one uses them and source control systems are made for the case you later want to go back, so even if someone is later missing this it could be restored and/or migrated to an appropriate place.

+1, just delete if someone misses them, we can restore

PR #51, opened 2022-04-29
Use parseInt in RepositoryTransport instead of valueOf

vogella:
Done with the corresponding JDT cleanup

### DIFF
--- a/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
+++ b/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
@@ -292,7 +292,7 @@ public class RepositoryTransport extends Transport {
 			String value = System.getProperty(TIMEOUT_RETRY);
 			if (value != null) {
 				try {
-					int retry = Integer.valueOf(value).intValue();
+					int retry = Integer.parseInt(value);
 					if (retry > 0) {
 						Integer retryCount = null;
 						if (socketExceptionRetry == null) {


HannesWell:
PR https://github.com/eclipse-equinox/p2/pull/54 should have fixed the build failure.

mickaelistria:
@vogella Can you please rebase on top of master and push force to your branch?

mickaelistria:
@HannesWell any idea about this new failure?

HannesWell:
> @HannesWell any idea about this new failure?

Not really.
The failure is:
```
[INFO] --- tycho-p2-extras-plugin:3.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) @ org.eclipse.equinox.p2.director ---
[ERROR] Baseline and reactor have same fully qualified version, but with different content
different
   META-INF/ECLIPSE_.RSA: present in baseline only
   META-INF/ECLIPSE_.SF: present in baseline only
```
But same fully qualified version and content that differs only in the jarsinger signatures actually happens for each project in the react that has not been changed. So that is very common and in this case the unchanged project is simply replaced by the baseline.

I restarted another build, maybe it was just a hickup, maybe another rebase on top of the master helps?

mickaelistria:
Build first says
```
[INFO] --- tycho-p2-plugin:3.0.0-SNAPSHOT:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.equinox.p2.director ---
[INFO] No baseline version MavenProject: org.eclipse.equinox.p2:org.eclipse.equinox.p2.director:2.5.200-SNAPSHOT @ /home/jenkins/agent/workspace/p2_PR-51/bundles/org.eclipse.equinox.p2.director/.polyglot.META-INF
```
while baseline repo for replacement should be https://download.eclipse.org/eclipse/updates/4.24-I-builds/ , which does contain  org.eclipse.equinox.p2.director_2.5.300. So it seems like the version in this patch is older than the one in I-Builds. It needs to be rebased on top of master.

mickaelistria:
@vogella It appears your previous rebase was not correct. Did you try `git pull --rebase eclipse master`?

laeubi:
> @vogella It appears your previous rebase was not correct. Did you try `git pull --rebase eclipse master`?

This branch is 1 commit ahead, 14 commits behind eclipse-equinox/p2:master.

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
```diff
@@ -296,7 +296,7 @@ public static DownloadStatus forException(Throwable t, URI toDownload) {
 			String value = System.getProperty(TIMEOUT_RETRY);
 			if (value != null) {
 				try {
-					int retry = Integer.valueOf(value).intValue();
+					int retry = Integer.parseInt(value);
```
laeubi, line 299: I think one even can use here `java.lang.Integer.getInteger`

laeubi (changes requested):

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
mickaelistria: Is it any better?

laeubi: As it combines checking for a system property with parsing the integer the whole block could be simplified, I also would assume that the JDK always choose the most performant way to parse string > int.

mickaelistria: Thanks. So yes, let's use `Integer.getInteger()` instead.

mickaelistria:
@vogella Can you please apply suggestions?

vogella:
Suggestion is great but I would like to separate that from the usage of the JDT clean-up. I feel comfortable with apply the JDT cleanup for this case without testing this, but adjusting the logic is something I would like to see in a dedicated change.

mickaelistria:
I'm going to apply the suggested fix. I don't see much value in doing partial 1 liner optimizations in 2 different commits

vogella:
Thanks.

PR #53, opened 2022-04-30
Simplify p2.tests.reconciler product

HannesWell:
The p2.tests.reconciler product can be simplified significantly if it is converted to be feature based.
But this requires a mixed product since no feature (except for the SDK) contains the `org.eclipse.equinox.p2.installer` plug-in, which seems to be necessary.

At the moment this seems not to work (altough Tycho should support this). I will investigate it further when completing [Bug 325614](https://bugs.eclipse.org/bugs/show_bug.cgi?id=325614).

### DIFF
--- a/bundles/org.eclipse.equinox.p2.reconciler.dropins/plugin.xml
+++ b/bundles/org.eclipse.equinox.p2.reconciler.dropins/plugin.xml
@@ -31,10 +31,10 @@
          point="org.eclipse.core.runtime.products">
       <product
             application="org.eclipse.equinox.p2.reconciler.application"
-            name="p2 reconciliation application">
+            name="Reconciler test application">
          <property
                name="appName"
-               value="p2 reconciliation application">
+               value="Reconciler test application">
          </property>
       </product>
    </extension>
--- a/bundles/org.eclipse.equinox.p2.tests.reconciler.product/reconciler.product
+++ b/bundles/org.eclipse.equinox.p2.tests.reconciler.product/reconciler.product
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.5"?>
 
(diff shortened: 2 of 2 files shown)


laeubi:
@HannesWell I noticed that there is an "examples" folder, I think we should either delete or move products there it seems the products are not really build. Then there is also a tests.reconciler that also contains a product so it seems all a bit duplicated...

If products are still build, we better should place them under "products/product_name" folder

HannesWell:
> Is this still relevant?

I think so.



> @HannesWell I noticed that there is an "examples" folder, I think we should either delete or move products there it seems the products are not really build. Then there is also a tests.reconciler that also contains a product so it seems all a bit duplicated...
> 
> If products are still build, we better should place them under "products/product_name" folder

I just created https://github.com/eclipse-equinox/p2/pull/114 to move all products (except for the test-product) to a `products` folder.

akurtakov:
This one needs manual rebase if still needed.

laeubi:
@HannesWell is this still needed?

HannesWell:
> @HannesWell is this still needed?

Would be nice to have, yes to also have some 'real-life' test cases of mixed products.

Given that the build succeeds, should we wait until https://github.com/eclipse-pde/eclipse.pde/pull/291 is completed? I plan/hope to finish that soon.

akurtakov:
What is the status of this one? It needs manual rebase at least.

HannesWell:
> What is the status of this one? It needs manual rebase at least.

I will complete https://github.com/eclipse-pde/eclipse.pde/pull/291 when back from vacation and will then take care of this.

HannesWell:
Great. The build finally succeeded.

akurtakov:
Thanks @HannesWell ! I had to fix this product so many times over the years so I'm quite happy to see it so much simpler.

PR #55, opened 2022-04-30
Leverage transitive inclusion in p2.sdk feature

HannesWell:
This simplifies the p2.sdk feature by adding other p2.features (IMHO a sdk feature should just include all features of a project) and by then removing all plug-ins that are already transitively included.


mickaelistria:
This change looks sane. However which one of Jenkins or GitHub actions is right here?

HannesWell:
> This change looks sane. However which one of Jenkins or GitHub actions is right here?

Jenkins is right. GH workflows constantly struggles with the compare-version-with-baseline mojo.
I have already disabled the GH-workflow:
https://github.com/eclipse-equinox/p2/pull/47#issuecomment-1114037790

mickaelistria:
Thanks.

laeubi:
> GH workflows constantly struggles with the compare-version-with-baseline mojo

I think it is not the Gh actions but the `compare-version-with-baseline mojo` to blame here as it seems it can't scope with more recent changes on the master branch, what is really bad as you can never build an older state of the repository then...

PR #59, opened 2022-05-04
Update maven.yml

laeubi:
Don't run the API checks as no one knows why they complain on GH action.


laeubi:
replaced by https://github.com/eclipse-equinox/p2/pull/69

PR #62, opened 2022-05-04
Move equinox.server.p2 into p2 repository

HannesWell:
In https://github.com/eclipse-equinox/equinox.bundles/pull/40#issuecomment-1116670974 it was discussed to move the o.e.equinox.server.p2 feature into the p2 repository since it mainly consists of p2 plug-ins and therefore should reside in the same repository.

This the addition part of the move. The removal is performed in https://github.com/eclipse-equinox/equinox.bundles/pull/42.

### DIFF
--- a/features/org.eclipse.equinox.p2.sdk/feature.xml
+++ b/features/org.eclipse.equinox.p2.sdk/feature.xml
@@ -43,6 +43,14 @@
          id="org.eclipse.equinox.p2.user.ui.source"
          version="0.0.0"/>
 
+   <includes
+         id="org.eclipse.equinox.server.p2"
+         version="0.0.0"/>
+
+   <includes
+         id="org.eclipse.equinox.server.p2.source"
+         version="0.0.0"/>
+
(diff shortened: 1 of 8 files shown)


HannesWell:
The server.p2 feature should probably also be added to the p2.sdk to keep that complete and to continue to publish the server.p2 feature.

HannesWell:
The build times-out regularly but since other resent PRs do that as well I assume that is not related to this change (I would be surprised anyway).
Replayed the Jenkins-job with doubled timeout. At least the GH-workflow succeeds 😃

akurtakov:
I guess it's time to double the timeout by default. Would you please do so?

HannesWell:
> I guess it's time to double the timeout by default. Would you please do so?

Yes, done with https://github.com/eclipse-equinox/p2/pull/72

HannesWell:
The test-failure on Jenkins seems unrelated for adding a new feature to his repository.
Going to submit this now.

PR #68, opened 2022-05-06
Use try-with-resources to close Stream opened by Files.list in SimpleBackupStore

drsgoodall:
This is a fix for Bug 579807

https://bugs.eclipse.org/bugs/show_bug.cgi?id=579807.

The change in SimpleBackupStore to use the Java NIO API does not close the stream opened by a call to Files.list. This can cause subsequent operations on the directory to fail. Specifically in CleanupZipAction the directory can still exists after deletion but in a inaccessible state.

On machines where this issue occurs the existing test case CleanupzipActionTest#testDirectoryCleanup will fail. In my experience these are Windows Server 2016 instances on AWS.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.touchpoint.natives/src/org/eclipse/equinox/internal/p2/touchpoint/natives/SimpleBackupStore.java
+++ b/bundles/org.eclipse.equinox.p2.touchpoint.natives/src/org/eclipse/equinox/internal/p2/touchpoint/natives/SimpleBackupStore.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.nio.file.*;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.*;
+import java.util.stream.Stream;
 import org.eclipse.osgi.util.NLS;
 
 /**
@@ -271,9 +272,11 @@ public class SimpleBackupStore implements IBackupStore {
 			throw new IllegalArgumentException(NLS.bind(Messages.BackupStore_not_a_directory, file.getAbsolutePath()));
 		}
 
-		if (Files.list(path).count() > 0) {
-			throw new IllegalArgumentException(
-					NLS.bind(Messages.BackupStore_directory_not_empty, file.getAbsolutePath()));
+		try (Stream<Path> s = Files.list(path)) {
+			if (s.count() > 0) {
+				throw new IllegalArgumentException(
+						NLS.bind(Messages.BackupStore_directory_not_empty, file.getAbsolutePath()));
+			}
 		}
 
 		return moveDirToBackup(path);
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.equinox.p2.touchpoint.natives/src/org/eclipse/equinox/internal/p2/touchpoint/natives/SimpleBackupStore.java
```diff
@@ -271,9 +272,11 @@ public boolean backupDirectory(File file) throws IOException {
 			throw new IllegalArgumentException(NLS.bind(Messages.BackupStore_not_a_directory, file.getAbsolutePath()));
 		}
 
-		if (Files.list(path).count() > 0) {
-			throw new IllegalArgumentException(
-					NLS.bind(Messages.BackupStore_directory_not_empty, file.getAbsolutePath()));
+		try (Stream<Path> s = Files.list(path)) {
+			if (s.count() > 0) {
```
laeubi, line 276: I think counting all items is not necessary `s.anyMatch(()->true)` should be sufficent

merks: Note that you are asking Simon to improve code that was reviewed and approved previously (last year) which is when this suggestion should have been made.  This is all fine and good, I just point that out because if we require contributors not only to fix a bug but also improve the existing style, we may turn them away.

@drsgoodall Of course the suggestion from @mickaelistria is a very good one.  Also, you will need to change the version in
/org.eclipse.equinox.p2.touchpoint.natives/META-INF/MANIFEST.MF from 1.4.300.qualifier to 1.4.400.qualifier for the build to stop complaining so it should be easy to do both things... Thanks so much for contributing!!

laeubi: @merks as the issue is with unclosed file handles I assume that @drsgoodall is seeing issues with files being locked for this or other processes.
So I think (even though this might fix the intermediate issue as well in the current state) its worth to lock not more files as necessary and exit as early as possible (and thus closing the opened file handle).

I also think the suggestion is not that complex to argue it can't be handled as part of this change?

merks: @laeubi I said it was a good suggestion!  I see that @drsgoodall considered it as well.   Thanks!!

drsgoodall:
I found two more places using Files.list which caused additional tests to fail. Do let me know if you want any further changes

laeubi:
> I found two more places using Files.list which caused additional tests to fail. Do let me know if you want any further changes

I have just fixed the Github action so it is not your fault, you can rebase your change and then it should succeed!

laeubi (approved):
This looks good to me. Jenkins build is already successful and GH action failure is unrelated here.

mickaelistria:
Thanks!

PR #69, opened 2022-05-06
Update maven.yml

laeubi:
always check out full


mickaelistria:
Can you please use a more explicit 1st line on commit message? Something like "GitHub Action fetch history"

PR #71, opened 2022-05-11
Use getInteger in RepositoryTransport instead of valueOf/parseInt

mickaelistria:
Done with the corresponding JDT cleanup


bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
```diff
@@ -289,33 +289,26 @@ public static DownloadStatus forStatus(IStatus original, URI toDownload) {
...
-						if (retryCount != null && retryCount.intValue() <= retry) {
-							socketExceptionRetry.put(toDownload, retryCount);
-							return new DownloadStatus(IStatus.ERROR, Activator.ID, IArtifactRepository.CODE_RETRY,
-									NLS.bind(Messages.connection_to_0_failed_on_1_retry_attempt_2, new String[] {
-											toDownload.toString(), t.getMessage(), retryCount.toString() }),
-									t);
-						}
+			int retry = Integer.getInteger(TIMEOUT_RETRY, 0);
+			if (retry > 0) {
+				Integer retryCount = null;
```
laeubi, line 294: Why is an Integer required here?? All these wrapping/unwrapping of int objects look strange and could most likely be replaced maybe by a map.compute even....

Just in case you like to cleanup even more here :-)

mickaelistria: I spent a couple of minutes here, and the Integer comes from the map (that doesn't allow int) so we have to deal with nullability and so on. This code could definitely be improved, however my quick hack session tells that it can be a can of worm and that it requires more attention to be properly improved than what I'm currently able to spend.

laeubi: Sure no problem, I found map.merge useful in such cases e.g.

`int current = map.merge(key, 1, (a,b)->a+b);`

vogella:
@mickaelistria please fix your commit message, currently you copied mine "Done with the corresponding JDT cleanup" which is not true anymore

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
```diff
@@ -289,33 +289,26 @@ public static DownloadStatus forStatus(IStatus original, URI toDownload) {
...
+						retryCount = Integer.valueOf(alreadyRetryCount.intValue() + 1);
 					}
-				} catch (NumberFormatException e) {
-					// ignore
+				}
+				if (retryCount != null && retryCount.intValue() <= retry) {
+					socketExceptionRetry.put(toDownload, retryCount);
+					return new DownloadStatus(IStatus.ERROR, Activator.ID, IArtifactRepository.CODE_RETRY,
+							NLS.bind(Messages.connection_to_0_failed_on_1_retry_attempt_2,
+									new String[] { toDownload.toString(), t.getMessage(), retryCount.toString() }),
```
HannesWell, line 310: If you are touching this anyway, I think the explicit array creation can be removed (IIRC NLS.bind takes a var-arg?).

HannesWell:
Does anyone of you know why the Jenkins-build times out? I have the same problem in PR #62. For now I just replayed the build with a doubled timeout.

PR #72, opened 2022-05-12
Raise Jenkins-build timeout to 80 min

HannesWell:


HannesWell:
The last build of the master succeeded in 25min so maybe there is some runtime regression due to the Jenkins update? Or did the infra change?

mickaelistria:
Duration strongly depends on the CI node. It's not a new issue. Of increasing the timeout helps, let's just try it.

PR #77, opened 2022-05-15
Do not deal with "publishPackFilesAsSiblings" property.

akurtakov:
Pack200 is thing from the past and not supported anymore so putting
properties for it in the metadata makes no sense.


akurtakov:
Give up on this one for 4.24. To be reconsidered next stream.

laeubi:
What is the problem with this and what will change for 4.24?

akurtakov:
First of all it needs extra work which I would spend now. Second, it changes mirroring to ignore pack.gz files and I don't want to risk breaking some simrel/epp/oomph releng script by pushing it now as reaction time will have to be quite fast and late in the cycle.

PR #85, opened 2022-06-13
Replace WorkbenchJob with ProvUIJob

laeubi:
ProvUUJob is a clean room implementation that is not tied to the workbench but still provides the desired functionality.

@merks do you like to review if this potentially causes issues for Oomph? There is only a (package private) field changed so I don't expect any issue here.


merks:
To what extent does this "copy" the org.eclipse.e4.ui.progress.UIJob implementation?  Do we really need/want more copies of that same concept?

laeubi:
@merks there is also an E4-UI job I could use this if you think it is better here:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.e4.ui.progress/src/org/eclipse/e4/ui/progress/UIJob.java

but of course then needs to add a dependency to `org.eclipse.e4.ui.progress` (what might be acceptable)

The original code is here:
https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/progress/UIJob.java

but that is bound do the Workbench (or we need to set a display... of course also possible if one do not forget about it...)

I'm open to any soloution, I just found that for the P2-UI case we probably don't need much more than a thin job extension here.

merks:
@vogella @mickaelistria 

Could you guys suggest what makes the most sense here?  My least favorite solution here is to duplicate "UIJob" yet again.   It seems org.eclipse.ui.dialogs.FilteredTree was copied for this same reason and I fear that as Christoph sweeps through the code that this will crop up too (and maybe we end up with copy number 3).

image

vogella:
+1 for reusing the e4 variant. Maybe in combination of deprecating org.eclipse.ui.progress.UIJob in favor of  org.eclipse.e4.ui.progress.UIJob. Thanks @merks for watching out for additional code copies.

mickaelistria:
> Maybe in combination of deprecating org.eclipse.ui.progress.UIJob in favor of org.eclipse.e4.ui.progress.UIJob

That would be another story. We recently discussed how to best handle deprecation for popular APIs (UIJob is one of them), and there was some consensus that before deprecating, we need to ensure there is no reference left in other parts of the SDK. This usually require a more dedicated effort that needs to be tracked separately.

laeubi:
I noticed there is `org.eclipse.e4.ui.progress.feature` but this seems not be included in the update-site e.g. https://download.eclipse.org/releases/2022-06/ do I miss something? It seems `org.eclipse.e4.ui.progress` is only included transitively...

laeubi:
Using this in a plain E4 causes:
```
java.lang.NullPointerException
	at org.eclipse.e4.ui.progress.UIJob.getUiSynchronize(UIJob.java:203)
	at org.eclipse.e4.ui.progress.UIJob.run(UIJob.java:84)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
This is because one needs to enable the progress Addon manually here and the getDisplay method that should handle other cases is never called... I'll propose a patch for this, but seem to indicate this job is not widely used (yet).

mickaelistria:
I think if you use `org.eclipse.ui.progress.UIJob` with a specific Display given as parameter, you'll still need `org.eclipse.ui` as a requirement but won't need a workbench to be started for things to work as expected. So it would run well in plain e4. I would recommend that as a 1st iteration is using other class is a can of worms.

laeubi:
> I would recommend that as a 1st iteration is using other class is a can of worms.

Not really a can, but some worms we never get rid of if we won't use it...

laeubi:
With this fix: https://github.com/eclipse-platform/eclipse.platform.ui/pull/134 the (e4) `UIJob` works, out of the box, so there is no more excuse to not migrate to it :-)

laeubi:
Together with https://github.com/eclipse-equinox/p2/pull/87 I can at least use the `InstallWizard` in a plain RCP E4 application.

laeubi:
Fix for e4 is merged, so I don't see a reason to delay this further once the build succeeds.

PR #97, opened 2022-06-29
Remove the help icon from the revert profile page

vogella:
To trigger this dialog use Help -> About -> Installation Details ->
Installation History and select an older state and press the Revert
button.

Operating system Ui guidelines do not recommend the usage of the help
icon in a dialog. For example check the screenshots in the Windows
guidelines for dialogs.

Operating System UI Guidelines

Windows UI guidence
https://docs.microsoft.com/en-us/windows/apps/design/controls/dialogs-and-flyouts/dialogs

Gnome UI guidence
https://developer.gnome.org/hig/patterns/feedback/dialogs.html

Mac UI guidence
https://developer.apple.com/design/human-interface-guidelines/components/presentation/alerts/

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/p2/ui/RevertProfilePage.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/p2/ui/RevertProfilePage.java
@@ -272,15 +272,21 @@ public class RevertProfilePage extends InstallationPage implements ICopyable {
 		revertAction = new Action() {
 			@Override
 			public void run() {
-				int result = MessageDialog.open(MessageDialog.QUESTION, getShell(), ProvUIMessages.RevertDialog_Title,
-						ProvUIMessages.RevertDialog_ConfirmRestartMessage, SWT.NONE,
-						ProvUIMessages.RevertProfilePage_RevertLabel, ProvUIMessages.RevertDialog_CancelButtonLabel);
-				if (result != Window.OK)
+				PlainMessageDialog dialog = PlainMessageDialog.getBuilder(getShell(), ProvUIMessages.RevertDialog_Title)
+						.message(ProvUIMessages.RevertDialog_ConfirmRestartMessage)
+						.buttonLabels(java.util.List.of(ProvUIMessages.RevertProfilePage_RevertLabel,
+								ProvUIMessages.RevertDialog_CancelButtonLabel))
+						.build();
+
+				int result = dialog.open();
+				if (result != Window.OK) {
 					return;
+				}
 				boolean finish = revert();
 				if (finish) {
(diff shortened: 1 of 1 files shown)


laeubi:
@vogella first of all this is (according to the code) a `QUESTION` and not a `HELP` icon.

Beside that, wold it not make sense to, instead of removing this from all Dialogs, to have a flag/preference/whatever in JFace one can set to simply not show an icon for a QUESTION Dialog? I don't think that this is actually related to the the icon itself, but to indicate that a dialog of type question is to be shows, an this could then be styled according to some UX guideline, what might or might not include a specific icon....

mickaelistria:
> wold it not make sense to, instead of removing this from all Dialogs, to have a flag/preference/whatever in JFace one can set to simply not show an icon for a QUESTION Dialog? 

I agree that would be the ideal situation to push such generic UI decision to JFace directly; with some flag to force a behavior or another. Note sure whether it's affordable though.

akurtakov:
Good solution is probably to enhance SWT to allow changing the OK/Cancel/etc. labels to be custom defined and rely on SWT MessageBox directly which will give us the native LnF and thus make them look far better. E.g. on Gtk https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/MessageBox.java#L239 is a good starting pointer.

laeubi:
By the way actually JFace is quite "poor" in its options there are much more when it comes do dialogs, you can download https://www.jgoodies.com/downloads/demos/ and go to the message dialogs section to see the many variants a "simple" confirmation *might* have.

So if I express to "show a question/confirmation" this should be enough for an UX aware framework to react accordingly, I should not be forced to decide wheater to show icons or not. Even worse, now we remove the overall meta-data that this *is* a confirmation/question and degrade it to a "plain" dialog.

akurtakov:
IMHO:
```
MessageBox messageBox = new MessageBox(shell, SWT.ICON_QUESTION | SWT.OK | SWT.CANCEL);
messageBox.setText(ProvUIMessages.RevertDialog_Title);
messageBox.setMessage(ProvUIMessages.RevertDialog_ConfirmRestartMessage);
//non-existing
messageBox.setStandardButtonLabels(Map.of(SWT.OK,ProvUIMessages.RevertProfilePage_RevertLabel, SWT.CANCEL, ProvUIMessages.RevertDialog_CancelButtonLabel));
```
will create nicely looking native dialog like :
image

vogella:
Changing the defaults in JFace openQuestion dialog so that we match again the OS recommendation for user facing qustions dialogs sounds like a great idea but in the past we faced lots of resistence in changing any defaults. Maybe it is time to try again.

vogella:
More general approaches should be done via separate issues / pr.

PR #112, opened 2022-07-19
Execute the build in parallel mode

laeubi:


laeubi:
Current GH/Jenkins run time is about 25 minutes.

PR #109, opened 2022-07-16
Use Tycho-CI-Friendly-Versions

laeubi:
This will make it easier to publish correct maven-metadata and deploy to
eclipse-maven-repository.

See https://github.com/eclipse/tycho/blob/master/RELEASE_NOTES.md#enhanced-support-for-maven-ci-friendly-versions

Also enables controlling the inclusion of pom

PR #113, opened 2022-07-19
Use more Tycho-pomless to remove more poms

HannesWell:
Use more Tycho-pomless to remove more poms and simply poms that cannot be removed (yet).


HannesWell:
Does anybody have an idea, why the `org.eclipse.equinox.frameworkadmin.test` is skipped, as intended, in the GH-build:
```
 [INFO] --- tycho-surefire-plugin:3.0.0-SNAPSHOT:test (default-test) @ org.eclipse.equinox.frameworkadmin.test ---
[INFO] Skipping tests
```
But runs on Jenkins?
```
[INFO] --- tycho-surefire-plugin:3.0.0-SNAPSHOT:test (default-test) @ org.eclipse.equinox.frameworkadmin.test ---
[INFO] Could not find a java toolchain of type jdk, using java from JAVA_HOME instead (/opt/tools/java/openjdk/jdk-17/latest/bin/java)
[INFO] Executing Test Runtime with timeout 0, logs (if any) will be placed at: /home/jenkins/agent/workspace/p2_PR-113/bundles/org.eclipse.equinox.frameworkadmin.test/target/work/data/.metadata/.log
[INFO] Command line:
	[/opt/tools/java/openjdk/jdk-17/latest/bin/java, -Dosgi.noShutdown=false, -Dosgi.os=linux, -Dosgi.ws=gtk, -Dosgi.arch=x86_64, --add-modules=ALL-SYSTEM, -Dosgi.clean=true, -ea, -jar, /home/jenkins/agent/workspace/p2_PR-113/.m2/repository/p2/osgi/bundle/org.eclipse.equinox.launcher/1.6.400.v20210924-0641/org.eclipse.equinox.launcher-1.6.400.v20210924-0641.jar, -data, /home/jenkins/agent/workspace/p2_PR-113/bundles/org.eclipse.equinox.frameworkadmin.test/target/work/data, -install, /home/jenkins/agent/workspace/p2_PR-113/bundles/org.eclipse.equinox.frameworkadmin.test/target/work, -configuration, /home/jenkins/agent/workspace/p2_PR-113/bundles/org.eclipse.equinox.frameworkadmin.test/target/work/configuration, -application, org.eclipse.tycho.surefire.osgibooter.headlesstest, -testproperties, /home/jenkins/agent/workspace/p2_PR-113/bundles/org.eclipse.equinox.frameworkadmin.test/target/surefire.properties]
```

Locally I already verified that the polyglot-pom contains a property `<skipTests>true</skipTests>`. So that value is somehow overwritten?

HannesWell:
> Does anybody have an idea, why the `org.eclipse.equinox.frameworkadmin.test` is skipped, as intended, in the GH-build:
> 

The Jenkinsfile defined the system-property `skipTests=false`. Actually it was obvious that it has to be the cause, but I didn't expect that and therefore didn't check it.
Created #143 to remove that.

bundles/org.eclipse.equinox.frameworkadmin.test/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.equinox.frameworkadmin.test
-Bundle-Version: 1.3.0.qualifier
+Bundle-Version: 1.3.100.qualifier
```
mickaelistria, line 5: build.properties should be part of the jgit.ignore in the pom.xml here, because changing it doesn't affect the bundle and should require to bump the version.

bundles/org.eclipse.equinox.p2.tests.discovery/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.equinox.p2.tests.discovery;singleton:=true
-Bundle-Version: 1.3.200.qualifier
+Bundle-Version: 1.3.300.qualifier
```
mickaelistria, line 5: same as above

mickaelistria (changes requested):
It's an issue that such a non-payload change in releng requires new versions of the bundles although no substantial content has changed.
You may want to first add the `build.properties` to the jgit.ignore description of the parent pom to make those changes not affect the version.

HannesWell:
> It's an issue that such a non-payload change in releng requires new versions of the bundles although no substantial content has changed. You may want to first add the `build.properties` to the jgit.ignore description of the parent pom to make those changes not affect the version.

In this case I agree that the version bump should not be necessary, but I don't think that changes in the build.properties are never of interest. For example if a new resource (e.g. an icon) is added then there should be a version bump, even if that inclusion was just forgotten before.

bundles/org.eclipse.equinox.frameworkadmin.test/META-INF/MANIFEST.MF
laeubi: > because changing it doesn't affect the bundle and should require to bump the version.

It depends, you can configure included content and that does affect the bundle, also this change affects the bundle (even though in a way that should be equivalent to the previous state)

akurtakov: Absolutely, we can't ignore build.properties due to bin.includes defined in it.

merks:
Just wondering though, if the build.properties change actually changes the content of the final jar (as in your icon example), then it will be noticed that the jar contents are different wouldn't it?  I only wonder about changes in the build.properties that affect the source bundle; I'm not sure how their versions are checked and determined.   In any case, if the build.properties is not in either jar then it seems to me that it could be ignored because only changes made to it that impact the contents of the jar(s), matter.  Or maybe I overlook something?

akurtakov:
I don't think that compare-version-with-baselines is that smart to check jar files, what it does is check last git commit and decide whether to replace based on it solely.

mickaelistria:
@akurtakov I think you're talking about baseline replacement (by artifact from previous build if there are from the same commit), not about baseline comparison with latest release (which is more an help to ensure we bump versions). But baseline replacement is also reporting the case when jar files are different for the same qualifier, as warning so far I believe.

As for build.properties participating to build output, then same is true for pom.xml or basically any build file: you can configure a pom file in a way that modifies the build output, yet pom.xml files are ignored because they usually do not participate to the final change, and in case they do, then this is supposedly detected during build.
I don't think build.properties is different from a pom.xml; even less with the proposed change.

merks:
@mickaelistria Yes, this is my impression as well, which is why I was wondering why others have a different impression...  I seem to recall that we noticed that a newer version of icu4j with changed method signatures such that a different "overload" was called, resulting in changing the generated .class file even though nothing was changed (committed) in the bundle itself was noticed by this mechanism...

HannesWell:
> @akurtakov I think you're talking about baseline replacement (by artifact from previous build if there are from the same commit), not about baseline comparison with latest release (which is more an help to ensure we bump versions). But baseline replacement is also reporting the case when jar files are different for the same qualifier, as warning so far I believe.
> 

It could be considered to increase the severity to error if the jar content is different for the same qualifier.
To only problem I see here is that verification builds often produce non signed jars, while the jars in the baseline are signed. comparison logic would have to be enhanced to ignore missing signatures.

If that would be possible the build.properties could be ignored without risking to alter the bundle content silently, either trough changes of the build.properties or pom.xml (which are indeed similar in this regard).

laeubi:
> If that would be possible the build.properties could be ignored without risking to alter the bundle content silently, either trough changes of the build.properties or pom.xml (which are indeed similar in this regard).

Actually I would not exclude **anything** so neither pom.xml nor build.properties, beside files temporarily created. In fact even if I change documentation or license info this is done for a reason (that is something should change) ... baseline replace will simply drop my change otherwhise and we have strange things like "force qualifier update" files ...

mickaelistria:
> Actually I would not exclude anything so neither pom.xml nor build.properties

This was already tried earlier and it failed because the pom (when they're present) do have an update of their parent version on each stream even if this generates the exact same payload/version.
But maybe this can be 2 steps: 1 adding the `build.properties` to `jgit.ignores` and later, when there are no pom left and all non-payload changes in build.properties are done, we can remove it.

laeubi:
> This was already tried earlier and it failed because the pom (when they're present) do have an update of their parent version on each stream even if this generates the exact same payload/version.

I just wanted to note that only because we exclude the pom (for whatever reason) it does not mean it is good to do so, because the *can* influence the build output... and actually changing the parent *do* changes the build output (e.g. other libraries might be pulled in and so on), so if **nothing changes** there is really no need to actually **update** the parent pom version!

mickaelistria:
@laeubi All you're saying is good in theory, but in practice, it's really not working as easy as you seem to believe. There are changing bits in the parent poms, there are also multiple streams that are built, yet we want reproducible version across streams. So we need versioning of parent (for different streams). pom-less is a good way to avoid that as the parent version is dynamically computed according to the working tree, that's good. But in the meantime, we should not generate new bundle versions for such change.
So instead of blaming the constraints for existing, let's discuss a way to make progress under those constraints, especially since this would later allow to get rid of those constraints.

laeubi:
> All you're saying is good in theory, but in practice, it's really not working as easy as you seem to believe.

I never stated it is/was easy ;-)

> So instead of blaming the constraints for existing,

I don't blame anyone/anything just wanted to point out a few things and for me it don't seem a big deal to bump the version for this change instead of ignoring build.properties completely. But if that's the preferred way, why not...

laeubi:
By the way, actually there is already a "maven way" of keeping the childs parent version constant by using https://maven.apache.org/maven-ci-friendly.html then the textual representation of the module `pom.xml` never changes.

An example could be found here (even though I use the enhanced Tycho version there):
https://github.com/eclipse-equinox/p2/pull/109/files#diff-5df1f6b125abedb4dbf31e2f19567f88b17043aab68d2b0dada67405c0162b31

HannesWell:
> It could be considered to increase the severity to error if the jar content is different for the same qualifier. To only problem I see here is that verification builds often produce non signed jars, while the jars in the baseline are signed. comparison logic would have to be enhanced to ignore missing signatures.

If we would do that, errors like https://github.com/eclipse-m2e/m2e-core/issues/896 would have been noticed immediately.

HannesWell:
> > @akurtakov I think you're talking about baseline replacement (by artifact from previous build if there are from the same commit), not about baseline comparison with latest release (which is more an help to ensure we bump versions). But baseline replacement is also reporting the case when jar files are different for the same qualifier, as warning so far I believe.
> 
> It could be considered to increase the severity to error if the jar content is different for the same qualifier. To only problem I see here is that verification builds often produce non signed jars, while the jars in the baseline are signed. comparison logic would have to be enhanced to ignore missing signatures.
> 

To wake this discussion up a bit, I created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/564 to implement the suggested approach.
I did the same for M2E for the reasons mentioned above:
https://github.com/eclipse-m2e/m2e-core/pull/906

laeubi:
Can this be merged?

mickaelistria:
> Can this be merged?

I didn't follow closely, but i don't think the initial concerns about ignoring build.properties in git timestamp computation are resolved, are they?

HannesWell:
> > Can this be merged?
> 
> I didn't follow closely, but i don't think the initial concerns about ignoring build.properties in git timestamp computation are resolved, are they?

Not really. I created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/564 to address them, but this probably requires some more configuration work in other platform repos. But since nobody said anything about it I did not start that effort yet. But IMHO that would be a beneficial approach and we use in M2E now for a bit and at least I did not notice any problems yet. But let's discuss that in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/564.

That being said, the version bumps only affect test Plug-ins. Therefore users of p2 are likely never affected of this version bump without out a Plug-in content change. So actually we could merge this as it is now, even when https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/564 is not ready.

akurtakov:
As it's only tests being affected - Go for it.

PR #114, opened 2022-07-19
Move all products to dedicated folder

HannesWell:
Additionally
- Remove now obsolete pom-files. They were only necessary because Tycho-pomless otherwise does not use packaging-type 'eclipse-plugin'.
- Remove p2.inf files for products, because they add repositories that don't exist. And if a update-repo is desired later it can be configured in the product.
- Remove launch-configs associated to products. If one wants to launch the product, it can be easaly done from the Product-editor.


HannesWell:
The org.eclipse.equinox.p2.director.app.product has to be changed to be based on Plug-ins. Otherwise we always have issues with features in other repos that include Plug-ins from this repo.

laeubi:
@HannesWell building products can add considerably to build times, I thin we should use a profile to explicitly enable that.

HannesWell:
> @HannesWell building products can add considerably to build times, I thin we should use a profile to explicitly enable that.

That's right. On other hand building them (at least occasionally) makes sure they continue to work.

Maybe it would be reasonable to only build them in I-builds?

HannesWell:
For now I changed this PR to only move the products to dedicated projects so that the pom-files can be removed. Fixing them will be done in a later PR.
However onces https://github.com/eclipse-tycho/tycho/pull/1299 is available (i.e. we use Tycho-3.0 in .mvn/extensions.xml) removing the pom.xml should be possible even with product-files within a Plug-in project because the BundleMapping now takes precedence over the Repository-Mapping. Nevertheless I think it is usefull to move those products into dedicated folders because this allows to build them (if desired), as long as Tycho does not support building products included in plug-in projects.

PR #119, opened 2022-07-24
Add bin/ entry to root-.gitignore and remove now obsolete .gitignore

HannesWell:


HannesWell:
Maybe it would be better to exclude `.gitignore` files from build-timestamp computation in the platform-parent pom.xml?

akurtakov:
@HannesWell do you plan to do the change in releng parent pom?

HannesWell:
> @HannesWell do you plan to do the change in releng parent pom?

Yes plan to do that sunday/next week when back from vacation.

HannesWell:
> > @HannesWell do you plan to do the change in releng parent pom?
> 
> Yes plan to do that sunday/next week when back from vacation.

Done in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/447

PR #120, opened 2022-07-26
Avoid File.getCanonicalPath

jukzi:
Solution copied from Bug 573456 / o.e.osgi.storage.Storage#getFile:

File.getCanonicalPath is much slower then java.nio.file.Path.normalize
under windows. Also avoid normalization if no ".." is involved.

tested by
org.eclipse.equinox.p2.tests.core.FileUtilsTest.testUnzipEscapeZipRoot()

### DIFF
--- a/bundles/org.eclipse.equinox.p2.core/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.core;singleton:=true
-Bundle-Version: 2.9.100.qualifier
+Bundle-Version: 2.9.200.qualifier
 Bundle-ClassPath: .
 Bundle-Activator: org.eclipse.equinox.internal.p2.core.Activator
 Bundle-Vendor: %providerName
--- a/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/helpers/FileUtils.java
+++ b/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/helpers/FileUtils.java
@@ -130,12 +130,44 @@ public class FileUtils {
 		return unzippedFiles.toArray(new File[unzippedFiles.size()]);
 	}
 
+	private static final boolean IS_WINDOWS = File.separatorChar == '\\';
+
+	// reserved names according to
+	// https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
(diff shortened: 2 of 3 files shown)


jukzi:
We have a usecase where java.io.File.getCanonicalPath() eats almost have the time of FileUtils.unzipFile() on JDK11
image

see also https://bugs.eclipse.org/bugs/show_bug.cgi?id=573456

laeubi:
> File.getCanonicalPath is much slower then java.nio.file.Path.normalize

Don't know if it is relevant, but `getCanonicalPath` and `Path.normalize` **do different things**!

### normalize():

>  general it derives from this path, a path that does not contain  *redundant* name elements

### getCanonicalPath()
> This typically involves removing redundant names such as "." and ".." from the pathname, **resolving symbolic links (on UNIX platforms)**, and converting drive letters to a standard case (on Microsoft Windows platforms).

So lets say I have a symbolic link named `B` that point to `A`, then `A.getCanonicalPath()` == `B.getCanonicalPath()` but `A.normalize()` != `B.normalize()`

jukzi:
> Don't know if it is relevant, but `getCanonicalPath` and `Path.normalize` **do different things**!

i know! but the only usecase here is to extract file from zip/tar and check that it is not extracted into another directory (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=535824). And that is the exact same usecase as https://bugs.eclipse.org/bugs/show_bug.cgi?id=573456. So the solution is the same. There is no need to follow links or find the actual case of letters here.

jukzi:
unrelated build error "Non-existing JDK home configuration at /opt/tools/java/openjdk/jdk-18/latest"

iloveeclipse:
Do we have test for https://bugs.eclipse.org/bugs/show_bug.cgi?id=266844?

jukzi:
> Do we have test for https://bugs.eclipse.org/bugs/show_bug.cgi?id=266844?

no. such a test could be
```
	public void testBug266844tar() throws IOException {
		File tar = TestActivator.getContext().getDataFile(getName() + ".tar.gz");
		try (TarOutputStream tos = new TarOutputStream(new FileOutputStream(tar))) {
			tos.putNextEntry(new TarEntry("./"));
			TarEntry entry = new TarEntry("./content.txt");
			entry.setSize(9);
			tos.putNextEntry(entry);
			tos.write("test data".getBytes());
			tos.closeEntry();
		}
		File temp = getTempFolder();
		FileUtils.unzipFile(tar, temp);
		File extracted = new File(temp, "content.txt");
		assertEquals("test data", Files.readString(extracted.toPath()));
		assertTrue("File not deleted", extracted.delete());
	}
```
It proves that the workaround for 266844 is no longer needed. As https://bugs.eclipse.org/bugs/show_bug.cgi?id=266844#c1 points out it was an error in jvm <=1.5

iloveeclipse:
> It proves that the workaround for 266844 is no longer needed

Could you then add this test to the PR?

jukzi:
> Could you then add this test to the PR?

done.

vogella:
Thanks @jukzi

PR #126, opened 2022-08-05
Add package-info and annotations to p2.core

laeubi:
Now PDE supports the OSGi Bundle/Version annotations, we can start to
use them in P2 to enable advanced tooling options

FYI @iloveeclipse as you wondered how this kind of stuff could be used.


akurtakov:
Is there anything blocking this one?

laeubi:
Lets see if anything is missing after rebase...

merks:
It's hard to see what the intended change here is with all the removals of "public static final", "public" and the Javadoc reformatting..  Wouldn't it be better to separate out the meaningful change from the syntactic changes and formatting changes?

merks:
Yes, often cleanup actions do unexpected things.   These changes changed the settings and then the source is no longer consistent with those settings:

image

Issue #123, opened 2022-07-28
Allow disabling the "Using unsafe http transport" warning


iloveeclipse:
I'm migrating our internal tooling from 4.21 to 4.25 platform, and see a lot of "unsafe http" warnings on startup at the time Oomph/update managers are checking for updates.
So log has few warnings like:
`Using unsafe http transport to retrieve http://our_local_server_in_intranet/, see CVE-2021-41033. Consider using https instead.`

I'm OK with reporting that in general, but either there should be a white list or a possibility to mute the warning.
We have few local update sites, and some are maintained by teams that simply run some http server with no https/ssl support, others have https but certificates are self signed and expired, etc.

So ideally one should be able to add some known intranet sites, or just mute the warning.

laeubi:
I think the main goal of this is to annoy people so the switch to some safe communication channel, maven for example even fails for http for a while, so probably using https with a company-root certificate would be a better solution than having a switch, waiting for the next eclipse release and then enabling that switch on each and every product?

iloveeclipse:
Our intranet **is** safe. If not, loading plugins to "unsafe" update sites will be the last thing an attacker will do.
I'm already annoyed, but that will change nothing, simply because in an engineering environment where multiple teams have freedom to do anything that helps them to work aren't managed by a single administrator who can magically switch something and all services will run over https.

So just imagine there are people who worked tens of years absolutely safe without this warning and want to coninue to do same, without being annoyed...

PR #129, opened 2022-08-10
Allow users mute "Using unsafe http transport" p2 warnings (#123)

iloveeclipse:
Added system property -Dp2.skipRepositoryProtocolCheck=true that allows users to run
Eclipse without getting dozens of "Using unsafe http transport" p2
warnings on every p2 update task.

Fixes https://github.com/eclipse-equinox/p2/issues/123

### DIFF
--- a/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/CacheManager.java
+++ b/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/CacheManager.java
@@ -82,6 +82,9 @@ public class CacheManager {
 	private static final String JAR_EXTENSION = ".jar"; //$NON-NLS-1$
 	private static final String XML_EXTENSION = ".xml"; //$NON-NLS-1$
 
+	/** Allows to mute "Using unsafe http transport" warnings */
+	private static final boolean SKIP_REPOSITORY_PROTOCOL_CHECK = Boolean.getBoolean("p2.skipRepositoryProtocolCheck"); //$NON-NLS-1$
+
 	private final HashSet<String> knownPrefixes = new HashSet<>(5);
 
 	/**
@@ -270,6 +273,9 @@ public class CacheManager {
 	}
 
 	private void checkLocationIsSecure(URI repositoryLocation) {
+		if (SKIP_REPOSITORY_PROTOCOL_CHECK) {
+			return;
+		}
 		if ("http".equals(repositoryLocation.getScheme())) { //$NON-NLS-1$
 			LogHelper.log(new Status(IStatus.WARNING, Activator.ID, NLS.bind(Messages.unsafeHttp, repositoryLocation)));
 		}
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/CacheManager.java
```diff
@@ -82,6 +82,9 @@ public CacheManager(IAgentLocation agentLocation, Transport transport) {
 	private static final String JAR_EXTENSION = ".jar"; //$NON-NLS-1$
 	private static final String XML_EXTENSION = ".xml"; //$NON-NLS-1$
 
+	/** Allows to mute "Using unsafe http transport" warnings */
+	private static final boolean NO_HTTP_WARNING = Boolean.getBoolean("p2.noHttpWarning"); //$NON-NLS-1$
```
laeubi, line 86: maybe `p2.skipLocationCheck` would be more generic (just in case we later find to warn about ftp or whatever location type as well)...

iloveeclipse: Good idea, will do tomorrow

merks: It seems to me that `location` is so generic a word it doesn't even suggest the location of what.  The installation, the cache, the pool?  Even if it's the location of repository, what about that repository's location is it checking or not?  Something like  *.skipRepositoryProtocolCheck would be more description of what is being checked or not.

I also don't know the extent to which p2's properties are consistently named/prefixed, but others are like this:

	public static final boolean MIRRORS_ENABLED = !"false".equals(Activator.getContext().getProperty("eclipse.p2.mirrors")); //$NON-NLS-1$//$NON-NLS-2$

But a full regex search reveal a number of naming patterns:

image

iloveeclipse: Ed, I was looking for `Boolean.getBoolean` / `System.getProperty`matches and that showed me "p2.setSomething" few times, like "p2.pgp.verifyExpiration" or "p2.fragments", but also "eclipse.p2" and "org.eclipse...".

I don't think this is really required to fully qualify "p2" system properties here, doesn't seem to be very important.

But I also see your point about location. 

So I would update it to "p2.skipRepositoryProtocolCheck". That is both specific & generic enough I think.

laeubi: > maybe `p2.skipLocationCheck` would be more generic (just in case we later find to warn about ftp or whatever location type as well)...

Please note that this always requires a started bundle, do not include *any* sanity checks and thus can be a real pain if your code simply fails with "static initialize error null" exceptions. And in most cases simply falls back to system properties if there are no explicit bundle properties.

So I think if one wants to use bundle-properties, then these should actually be assigned inside the activators start method and not in static final fields!

iloveeclipse: > if one wants to use bundle-properties

No. I just want a system property.

merks: @iloveeclipse I'm happy with the name you chose.  

As we  noticed, there is not really a consistent naming pattern and I don't think there are many other things name p2 floating around.

PR #136, opened 2022-08-18
replacing org.apache.commons.io with org.apache.commons.commons-io

sravanlakkimsetti:
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/489

### DIFF
--- a/bundles/org.eclipse.equinox.p2.tests.reconciler.product/reconciler.product
+++ b/bundles/org.eclipse.equinox.p2.tests.reconciler.product/reconciler.product
@@ -34,7 +34,7 @@
       <plugin id="org.apache.batik.i18n"/>
       <plugin id="org.apache.batik.util"/>
       <plugin id="org.apache.commons.codec"/>
-      <plugin id="org.apache.commons.io"/>
+      <plugin id="org.apache.commons.commons-io"/>
       <plugin id="org.apache.commons.jxpath"/>
       <plugin id="org.apache.commons.logging"/>
       <plugin id="org.apache.felix.scr"/>
--- a/bundles/org.eclipse.equinox.p2.ui.admin.rcp/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.ui.admin.rcp/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-Name: %bundleName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-SymbolicName: org.eclipse.equinox.p2.ui.admin.rcp; singleton:=true
-Bundle-Version: 1.3.200.qualifier
+Bundle-Version: 1.3.300.qualifier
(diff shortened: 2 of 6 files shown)


akurtakov (approved):
We should get rid of the old lib for RC1.

merks:
@sravanlakkimsetti  Note that it seems to me there has previously been a lot of churn about this particular bundle...

https://github.com/eclipse-tycho/tycho/issues/1133

So I'm concerned that we are bouncing back and forth and well end up with that issue again afterwards...

akurtakov:
I don't understand the concern here. What this patch does is to ensure only one symbolic name is used.

merks:
@akurtakov  You're right.  It's all just confusing, so I'm confused.  The repo has two:

image

But I don't see in the *.target where the `org.apache.commons.io` comes from there...  From some dependency I guess...

But I can't find it published in any dependency repos:

image

I see it in the TP:

image

But where does it come from?

Will any change make it not be in the TP?  If it weren't in the TP we'd have noticed a problem with it being used in products and we'd be less likely to use it accidentally...

Anyway, we probably have bigger fish to fry...

akurtakov:
I see the 2 issues and this change is supposed to make it not in published repo and TP investigation can go in parallel.

merks:
@akurtakov Yes, sorry for the disturbance!

merks:
For completeness, the older version appears to come from here:

image

I.e., from the recursive dependencies of  `commons-fileupload`. 

Looking at its actual OSGi requirements in a *.aggran editor:

image

we can see that its unconstrained package requirements are satisfied by either version of commons io.

Looking at the older bundle's provided packages, we can see that nothing requires something from it that isn't also provided by the newer version:

image

I assume it's pulled in because of how it specifies dependencies:

https://repo1.maven.org/maven2/commons-fileupload/commons-fileupload/1.4/commons-fileupload-1.4.pom

Specifically these:
```
  <dependencies>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.12</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>javax.servlet</groupId>
      <artifactId>servlet-api</artifactId>
      <version>2.4</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>portlet-api</groupId>
      <artifactId>portlet-api</artifactId>
      <version>1.0</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>commons-io</groupId>
      <artifactId>commons-io</artifactId>
      <version>2.2</version>
    </dependency>
  </dependencies>
```

So I think one lesson here is that using `includeDependencyDepth="infinite` in the maven target location can and does pull in dependencies older than explicitly specified dependencies, and in particular older than ones that would also serve to satisfy all the OSGi requirements of that IU.  Perhaps we should be more careful somehow?  Or the implementation should be smarter or different?

mickaelistria (approved):

PR #143, opened 2022-09-07
Don't set system-property skipTests=false in Jenkins build

HannesWell:
Otherwise no module can enable skipTests.

@mickaelistria is there any reason (that is still relevant), that this property was set?

And remove 'maven.test.error.ignore=true' and 'maven.test.failure.ignore=true' because the build already has '--fail-at-end' set

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,8 +18,7 @@ pipeline {
 					sh """
 					mvn clean verify --batch-mode --fail-at-end -Dmaven.repo.local=$WORKSPACE/.m2/repository \
 						-Pbuild-individual-bundles -Pbree-libs -Papi-check \
-						-DskipTests=false -Dcompare-version-with-baselines.skip=false \
-						-Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true \
+						-Dcompare-version-with-baselines.skip=false \
 						-Dproject.build.sourceEncoding=UTF-8 \
 						-T1C
 					"""


mickaelistria:
I don't remember a good reason, so let's merge. Thanks.

laeubi:
> And remove 'maven.test.error.ignore=true' and 'maven.test.failure.ignore=true' because the build already has '--fail-at-end' set

As explained elsewhere, these are **not** equivalent and will probably break the semantics of the Jenkins pipeline.  '--fail-at-end'  will still fail the build and thus Jenkins will report a Build Failure (red dot), while 'maven.test.error.ignore=true' and 'maven.test.failure.ignore=true' will only report a Build Failure when there are compile errors and then Jenkins will mark the Build as unstable (yellow dot) if there are test-failures.

Second, even if  '--fail-at-end'  is set, maven will skip any dependent module if there are any test-failures, while with the other dependent modules are at least compiled (and there tests executed).

> Otherwise no module can enable skipTests.

Why is this needed? If we want to disable test one should better disable the plugin execution, or simply delete / disable tests?

mickaelistria:
Red dot/yellow dot... This doesn't make much difference. There are only 2 kinds of states/dots: good/green and bad/not green. The color where it's not green doesn't matter as the remediation is the same: look at test and build log and fix what needs to be fixed.

> if '--fail-at-end' is set, maven will skip any dependent module if there are any test-failures

Actually, I think it's a good thing. A test failure in 1 module will cascade to other failures in other modules. By not building downstream modules, it allows to more quickly identify the root test failures and focus on it, before considering further steps. And additionally it gives faster feedback in case more work is needed.

laeubi:
> Red dot/yellow dot... This doesn't make much difference. There are only 2 kinds of states/dots: good/green and bad/not green. The color where it's not green doesn't matter as the remediation is the same: look at test and build log and fix what needs to be fixed.

There actually is a difference because I can't know if the failure is because of general compilation problem (red) --> look at the logfile or test-failure -> look at the test reports.

And if it dosen't matter why `fae` is better than the previous choice? This is simply useless optimization as it doesn't help anything.

> Actually, I think it's a good thing. A test failure in 1 module will cascade to other failures in other modules. By not building downstream modules, it allows to more quickly identify the root test failures and focus on it, before considering further steps. And additionally it gives faster feedback in case more work is needed.

The one should not use `fae` at all ... but that was not discussed or argued here, but that `fae` and `failure.ignore` seem redundant what is simply not the case ...

mickaelistria:
> There actually is a difference because I can't know if the failure is because of general compilation problem (red) --> look at the logfile or test-failure -> look at the test reports.

On the Jenkins job page, the test failures are still reported directly. So you don't need to look at the log if there is a test failure, uou can directly click the test report.

> why fae is better than the previous choice?

You're right, maybe we should even remove the `-fae` to fail fast.

HannesWell:
> > And remove 'maven.test.error.ignore=true' and 'maven.test.failure.ignore=true' because the build already has '--fail-at-end' set
> 
> As explained elsewhere, these are **not** equivalent and will probably break the semantics of the Jenkins pipeline. '--fail-at-end' will still fail the build and thus Jenkins will report a Build Failure (red dot), while 'maven.test.error.ignore=true' and 'maven.test.failure.ignore=true' will only report a Build Failure when there are compile errors and then Jenkins will mark the Build as unstable (yellow dot) if there are test-failures.

You are right that it is not exactly the same, but with the current set up of Jenkins and GitHub it has the same effect, especially since Github only have green or red results. So in case it is not green one has to check Jenkins anyways and test failures will be presented very prominent.

> 
> Second, even if '--fail-at-end' is set, maven will skip any dependent module if there are any test-failures, while with the other dependent modules are at least compiled (and there tests executed).

That is correct but usually there are no modules that depend on a test-plugin. And even if we have the test code in a 'regular' plugin module, then it is at least questionable if descendants should be tested (build is probably fine because the parent build fine) because of the mentioned follow up errors.
But of course it can be desired to still build descendants. Then the developer just has to keep in mind that the test-failures could be follow-ups caused by errors in the parent. I'm not sure what the best way is here. But it can make sense to have both -fae and test.failures/errors=ignore.

> > why fae is better than the previous choice?
> 
> You're right, maybe we should even remove the `-fae` to fail fast.

I think we should keep `-fae`. For example if one changed code in many bundles and has not bumped versions it is desired to have as many reports as possible at once to not have to wait for a build failure for each version bump. Of course ideally that is already indicated in the IDE but as long as we don't have that the build should not fail at the very first failure.

laeubi:
> You are right that it is not exactly the same, but with the current set up of Jenkins and GitHub it has the same effect, especially since Github only have green or red results.

Also not really true, Github has `Checks` so you will get a Green Check for a successful build and a Red Check for Failing tests, now the JUnit Check must always run and reports "grey" result (= no tests found) in case of compilation failure.

> That is correct but usually there are no modules that depend on a test-plugin. And even if we have the test code in a 'regular' plugin module, then it is at least questionable if descendants should be tested (build is probably fine because the parent build fine) because of the mentioned follow up errors.

It was regularity complained that e.g. missing bumps are reported to late (e.g. you fix one and then you see there are probably more), so there is actually a need for collection as many problems as possible, and even for tests it is interesting to see if only some or all test fail due to a change.

> I think we should keep `-fae`. For example if one changed code in many bundles and has not bumped versions it is desired to have as many reports as possible at once to not have to wait for a build failure for each version bump.

Either it should fail fast always or fail never "fast" but collect as many problems as possible ...

HannesWell:
> > You are right that it is not exactly the same, but with the current set up of Jenkins and GitHub it has the same effect, especially since Github only have green or red results.
> 
> Also not really true, Github has `Checks` so you will get a Green Check for a successful build and a Red Check for Failing tests, now the JUnit Check must always run and reports "grey" result (= no tests found) in case of compilation failure.

But isn't the JUnit Check just running for the GH-build?

> > I think we should keep `-fae`. For example if one changed code in many bundles and has not bumped versions it is desired to have as many reports as possible at once to not have to wait for a build failure for each version bump.
> 
> Either it should fail fast always or fail never "fast" but collect as many problems as possible and then it should include unittest ... and thats what the previous configuration satisfied.

Ideally Jenkins should magically know exactly what I really want to test and then only build that so that I get all information I want but fast. Unfortunately that's not possible as far as I know...
For now I created https://github.com/eclipse-equinox/p2/pull/144 to revert that part, so that we get as much information in one run as possible.

laeubi:
> Ideally Jenkins should magically know exactly what I really want to test and then only build that so that I get all information I want but fast. Unfortunately that's not possible as far as I know...

Jenkins actually can do this with maven incremental builds but

1. maven incremental builds are not often reliable
2. tycho has some issues with incremental builds (e.g. features / products)

A more reliable approach would be to take the git timestamp into account:

- https://github.com/eclipse-tycho/tycho/issues/87

What one could do already would be (but of course its a bit more work to setup):

1. Jenkins has an ENV variable `GIT_PREVIOUS_SUCCESSFUL_COMMIT`
2. You can determine all files that where changed between `GIT_PREVIOUS_SUCCESSFUL_COMMIT` and `GIT_COMMIT`
3. The filter that list to get a list of corresponding project folders, possibly including test projects always.
4. Call maven with `mvn -pl <your project folder list> -am clean install`
5. If you want to make sure you do not break any consumer add `-amd`

This will build the smallest possible set of projects that are affected by your change.

HannesWell:
>     1. Jenkins has an ENV variable `GIT_PREVIOUS_SUCCESSFUL_COMMIT`

Is that also available for Jenkins instances running on kubernetes agents? I'm not familiar with kubernetes but I assumed that the workspace is effectively wiped out on each run with that setup?
Furthermore I wonder if what that variables value is on the first run for a new PR? Is it the branching commit or just null?

But in general this could be build using the 'branching' commit.

laeubi:
Job metadata is stored at the maser node not in the workspace. I haven't tested that for new PRs but for sure one could then use the parent commit of the branch.

PR #150, opened 2022-10-02
Use Tycho 3.0.0 build extension

laeubi:
@akurtakov any objections to upgrade to 3.0.0 here?


akurtakov:
Go for it.Bonus points for updating platform releng parent

laeubi:
If https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/608 success I think we can merge this.

Issue #153, opened 2022-10-14
Compile error using org.bouncycastle.bcpg_1.72.0.v20221013-1810


iloveeclipse:
In the new bcpg version, the `org.bouncycastle.openpgp.PGPPublicKeyRingCollection.PGPPublicKeyRingCollection(Collection<PGPPublicKeyRing>)` constructor does not define exceptions, see 
https://github.com/bcgit/bc-java/commit/ebceced5a919485e9c6335b35efdabed3c169820

So if one installs org.bouncycastle.bcpg_1.72.0.v20221013-1810 from Orbit in the platform, `DefaultPGPPublicKeyService` doesn't compile anymore in the workspace.

I will provide a compatible patch.

PR #154, opened 2022-10-14
Fix compile error using org.bouncycastle.bcpg_1.72.0.v20221013-1810

iloveeclipse:
Fixes https://github.com/eclipse-equinox/p2/issues/153


akurtakov:
Catching Exception is a risky pattern. This raises the question how did our builds succeed?

iloveeclipse:
> Catching Exception is a risky pattern. This raises the question how did our builds succeed?

Before they thrown 2 different exceptions that are now **not** thrown anymore. The patch is compatible to both versions. But this is chicken and egg problem. We can't remove exception catch before we update the bcpg, and we can't update bcpg before we fix compile issue.

So I would say we can remove the catch after we have a stable build with new version.

laeubi (approved):
Did the same "hack" for Tycho, if one want to be extra correct maybe RuntimeExceptions can be thrown as is without wrapping, but actually if we came here something is really broken and one don't mind anyways.

PR #155, opened 2022-10-15
Use new static UIJob factories

HannesWell:


HannesWell:
This is probably not worth the version bump...

akurtakov:
Every code improvement is worth the version bump! These are 4 less anonymous classes, smaller bytecode, less classloading and etc. but most important there is less boilerplate which makes the code easier to grasp for the next person that will look into it.

PR #156, opened 2022-10-17
Remove now unneeded try/catch block (#153)

iloveeclipse:
After bcpg update there is no need in this try/catch anymore.

See discussion on https://github.com/eclipse-equinox/p2/issues/153


akurtakov:
I think version 1.72 of bouncycastle should be required in manifest.mf too so we are sure we would not end up with older version that may throw it.

iloveeclipse:
> I think version 1.72 of bouncycastle should be required in manifest.mf too so we are sure we would not end up with older version that may throw it.

Good point, I will update that.

PR #160, opened 2022-10-29
Support sha1/512 as additional checksums in P2

laeubi:
SHA1 is the default signature for maven and recently support was added up to sha-512 so P2 should support this as well.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/plugin.xml
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/plugin.xml
@@ -50,9 +50,27 @@
 	<extension
 			point="org.eclipse.equinox.p2.artifact.repository.artifactChecksums">
 		<artifactChecksum
+        algorithm="SHA-1"
+        id="sha-1"
+        priority="-1000"
+        publish="false"
+        warnInsecure="true">
+		</artifactChecksum>
+	</extension>
+	<extension
+			point="org.eclipse.equinox.p2.artifact.repository.artifactChecksums">
+		<artifactChecksum
         algorithm="SHA-256"
         id="sha-256"
         priority="1000">
+		</artifactChecksum>
+	</extension>
+	<extension
+			point="org.eclipse.equinox.p2.artifact.repository.artifactChecksums">
+		<artifactChecksum
+        algorithm="SHA-512"
+        id="sha-512"
+        priority="2000">
 		</artifactChecksum>
 	</extension>
 	
--- a/bundles/org.eclipse.equinox.p2.tests/src/org/eclipse/equinox/p2/tests/artifact/processors/AllTests.java
+++ b/bundles/org.eclipse.equinox.p2.tests/src/org/eclipse/equinox/p2/tests/artifact/processors/AllTests.java
@@ -21,7 +21,8 @@ import org.junit.runners.Suite;
  */
 @RunWith(Suite.class)
 @Suite.SuiteClasses({ ZipVerifierProcessorTest.class, ChecksumVerifierTest.class,
-		ChecksumUtilitiesTest.class, PGPSignatureVerifierTest.class, ProduceChecksumTest.class })
+		ChecksumUtilitiesTest.class, PGPSignatureVerifierTest.class, ProduceChecksumTest.class,
+		ChecksumPriorityTest.class })
 public class AllTests {
 // test suite
 }


merks:
I don't see the point in supporting sha1 when its security is considered marginal these days:

https://en.wikipedia.org/wiki/SHA-1

In that article it says `it is recommended to remove SHA-1 from products as soon as possible` so adding now seems completely questionable. 

Given that modern p2 repositories already have sha-256 checksums (and md5 which is deprecated), why add something that's highly likely to soon be deprecated?

laeubi:
> I don't see the point in supporting sha1

See above SHA-1 is the default signature for maven, and as we include more and more maven items p2 should support this as well. Even if maybe maven will slowly move on to sha-256 there will always be older items that are not.

About security, SHA-1 (and also md5) are vulnerable to certain attacks but for validating the download integrity they are still fine (and I get many complains about updatesites using md5 at least) as you need a full [pre-image attack](https://en.wikipedia.org/wiki/Preimage_attack), at least here it was considered "not practical at all": https://crypto.stackexchange.com/questions/41860/pre-image-attack-on-md5-hash

Also keep in mind, that this only means it is feasible to create a second file that produces the same md5 sum, it does not mean the file is good for anything! So one can consider an attack that produces a malicious jar AND producing the same md5 of an existing p2 artifact quite impossible.

Still MD5 and SHA1 are a bit outdated, but if nothing else is present it is better than nothing...

laeubi:
By the way this does not mean you can **produce** and site with such signature, just that it possible to **verify** a site containing such signatures!

laeubi:
@merks are you fine with it or still concerns? Would it be fine if we mark sha1 as insecure like md5? My main goal is to pull this part out of Tycho where we currently implement this as a workaround because p2 otherwise complains that no suitable algorithm can be found.

merks:
Even when verifying md5 @mickaelistria already added logic to produce a warning that md5 is not considered secure.  So I see that day coming for sha-1 as well.  Also, no p2 site has sha1-checksums currently because nothing produced them before now and hence there is nothing to check, good, bad or indifferent in this regard.  I also fail to see  how Maven's default is relevant.   It's not as if we need to copy the sha-1 from Maven; we can compute it, or something much better, if and when it's needed.  But we don't even preserve the pgp signatures from Maven, which cannot be recomputed though we can and do compute a sha-256.  So would it not be better to discourage the use of sha-1 rather than introduce something new that encourages new-never-before uses of sha-1?  I just don't get the compelling reason to add something to encourage sha-1 (or read it when it's never been produced before) when sha-256 is already broadly adopted in p2.

merks:
I just don't get it.  Where are the sites containing sha-1 coming from?  There are no such sites because p2 never supported this...

merks:
How does one control which checksums are produced by Tycho.  I believe the following does that work, but it doesn't say much in detail:

https://tycho.eclipseprojects.io/doc/latest/tycho-p2/tycho-p2-repository-plugin/fix-artifacts-metadata-mojo.html

I do vaguely recall complaints about checksum computations being really expensive.  So is it the case that with this p2 change all these checksums will be computed by Tycho?  Do we have any idea about the performance impact of that?

laeubi:
The checksums are downloaded from maven-central already and are embedded in the P2 metadata now by Tycho otherwise you get complains about missing/unsupported signatures see
- https://github.com/eclipse-tycho/tycho/issues/1482

if you reexport them they are usually always gets a sha-256 added, but currently they are just not checked when used *inside* the build so you just assume they are fine and probably add a broken artifact (with new computed sha-256 sum).

Because of this we now add the maven provided signatures, but then p2 complains that Sha1 sums can not be checked ... so not much of an improvement here.

> already added logic to produce a warning that md5 is not considered secure

Yes that's just a flag on the extension, so if you are concerned I would be fine with enabling the warning for sha1 as well.

> Do we have any idea about the performance impact of that?

I don't think that performance is a concern, but we should not "fake" meta-data here as we actually have no sha-256 for certain maven artifacts. So as a first step it should go into the build and gets verified by p2 at that point already, later on when producing an update/site we already add sha256 as an additional signature anyways.

merks:
Sorry, I continue not to get the point.   You point at an issue where apparently Tycho fails to generate check sums in the p2 artifact metadata...  How that relates to sha-1 in particular is not clear. It might be clearly obvious to you but not to me.    The vast majority of the p2 artifacts are ultimately produced from sources built by Tycho which computes the check sum when assembling the p2 repository.  They're all currently producing sha-256 and have never produced sha-1.

This current state begs the following still unanswered questions:

1. Where are the p2 sites containing sha-1 coming from?
2. Is it the case that with this p2 change all these checksums will be computed by Tycho?

laeubi:
> 1. Where are the p2 sites containing sha-1 coming from?

The sites are mostly constructed in memory and the sha-1 (and probably others like md5) are coming from maven central when you consume an artifact from there, this metadata then can show up in a materialized site as well (ass these maven GAV or pgp keys does).

> 2. Is it the case that with this p2 change all these checksums will be computed by Tycho

Tycho does never computes checksums, it use them either from P2 (e.g. because you add a P2 update site to your build), or from maven (e.g. because you use a maven target location). In case of an update site-assembly Tycho calls P2 that in turn seem to add sha256 sum by default and md5 for backward compatibility.

merks:
> > 1. Where are the p2 sites containing sha-1 coming from?
> 
> The sites are mostly constructed in memory and the sha-1 (and probably others like md5, sha512, ...) are coming from maven central when you consume an artifact from there, this metadata then can show up in a materialized site as well (as these maven GAV or pgp keys does).
> 

During this download process from Maven the checksum should be used to verify the download.  I'm pretty sure Maven does that already.   We're currently tossing out the PGP key.  I don't see a compelling reason why we should record the sha-1 in the p2 metadata.  No existing release of p2 can check that but it can check sha-256 so we if we do need to check again (after already having checked it during down!) we should produce/calculate that instead because a) it can and will be used now by any p2 application and b) because it can and will be used in the future.

> > 2. Is it the case that with this p2 change all these checksums will be computed by Tycho
> 
> Tycho does never computes checksums, it use them either from P2 (e.g. because you add a P2 update site to your build), or from maven (e.g. because you use a maven target location). In case of an update site-assembly Tycho calls p2 that in turn seem to add sha256 sum by default and md5 for backward compatibility.

So you're saying Tycho uses p2 and then p2 creates the checksums.  I think org.eclipse.equinox.internal.p2.artifact.processors.checksum.ChecksumUtilities.calculateChecksums(File, Map<String, String>, Collection<String>) does that work and it does compute **all** check sums, except ignored ones, though both calls to this method pass in an empty list for the ignored ones.  Notice how this method is implemented to read the artifact **once per algorithm**.  Hence my expressed concern about performance and my concern that forcing a sha-1 calculation when it's weak and won't generally be useful.  Even forcing a sha-512 calculation will have performance impact and it's not entirely clear that this add significant additional value.

laeubi:
> During this download process from Maven the checksum should be used to verify the download. I'm pretty sure Maven does that already.

Maven by default warns about checksum mismatches.

> I don't see a compelling reason why we should record the sha-1 in the p2 metadata.

If no checksum is there (or some unknown is used) P2 complains that it cannot verify the integrity (whats correct of course), and thats what this issue (and the tycho one) is about, this check is **not** performed only when downloading stuff from a remote site but also from other locations as well (e.g. memory backed ones) and P2 even has hard coded defaults to suppress this for locations starting with "file://" ... this is simply bad, and I don't see a reason why P2 should not support verify more checksums even if it is not capable of producing such checksum.

> does that work and it does compute **all** check sums, except ignored ones, though both calls to this method pass in an empty list for the ignored ones. Notice how this method is implemented to read the artifact **once per algorithm**. Hence my expressed concern about performance and my concern that forcing a sha-1 calculation when it's weak and won't generally be useful.

But why do you have no concerns about computing an even less secure md5? Actually if performance is a concern, we better enhance this class to allow for a priority of checksums and only checking those with the highest priority first and simply leave the others alone!

merks:
> > During this download process from Maven the checksum should be used to verify the download. I'm pretty sure Maven does that already.
> 
> Maven by default warns about checksum mismatches.
> 
> > I don't see a compelling reason why we should record the sha-1 in the p2 metadata.
> 
> If no checksum is there (or some unknown is used) P2 complains that it cannot verify the integrity (whats correct of course), and thats what this issue (and the tycho one) is about, this check is **not** performed only when downloading stuff from a remote site but also from other locations as well (e.g. memory backed ones) and P2 even has hard coded defaults to suppress this for locations starting with "file://" ... this is simply bad, and I don't see a reason why P2 should not support verify more checksums even if it is not capable of producing such checksum.

If the p2 repo needs a checksum (and it generally **should** have one!), then calculate a sha-256 and the problem is solved. But I have a feeling this augment will go in circles though.  😢 

> 
> > does that work and it does compute **all** check sums, except ignored ones, though both calls to this method pass in an empty list for the ignored ones. Notice how this method is implemented to read the artifact **once per algorithm**. Hence my expressed concern about performance and my concern that forcing a sha-1 calculation when it's weak and won't generally be useful.
> 
> But why do you have no concerns about computing an even less secure md5? Actually if performance is a concern, we better enhance this class to allow for a priority of checksums and only checking those with the highest priority first and simply leave the others alone!

The md5 is there for backward compatibility.  I think sha-256 was added in 2018-03 committed by @akurtakov so anything older will not find a checksum.  And yes, if we are going to add two more checksums,  including one that is little better than md5, we'd better ensure that the resource is only read once while computing it.   Performance is always a concern so the performance of org.eclipse.equinox.internal.p2.artifact.processors.checksum.ChecksumUtilities.getChecksumVerifiers(IArtifactDescriptor, String, Set<String>) would be another concern because I surely hope we don't need to check 4 different check sums...

laeubi:
> If the p2 repo needs a checksum (and it generally **should** have one!), then calculate a sha-256 and the problem is solved. But I have a feeling this augment will go in circles though. cry

Yes because you assume one wants to fake meta-data that is not the real one. If I'm not the **producer** of the artifact then I should not produce a fake checksum I should verify and use the one from the supplier. If that is **checked**, and I **redistribute** the artifact I might want to add an additional one. Otherwise you don't see the warning but you effectively only suppress it... and that's what has happens for Tycho for years, because error from OSGi never where visible no one noticed that until now.

> Performance is always a concern so the performance of org.eclipse.equinox.internal.p2.artifact.processors.checksum.ChecksumUtilities.getChecksumVerifiers(IArtifactDescriptor, String, Set) would be another concern because I surely hope we don't need to check 4 different check sums...

I now enhanced the PR in the following way:

1. marked `sha1` as "not recommended", so if only sha1 (or only `sha1` + `md5`) is used users get a warning.
2. added a priority (and test for that) so the only always the "best" algorithm is chosen.

Anyways, checking many different checksums is actually good, because breaking one would not sufficient, so if there is an day-0 exploit to sha-256, one still need to fully break md5 (and probably sha-1). Also modern CPUs can compute literally millions of hash-sums in less than a second, so if we assume that the data is read only once from net/disk this is simply negligible.

merks:
I don't like your talk about **fake** checksums.  There's no such thing.  A checksum is either correct or incorrect.  If it's incorrect we should **not propagate the artifact** farther downstream through the build system and to the consumers; this is a supply chain integrity problem if not security concern.   On the other hand, if the checksum is correct, we can correctly compute any checksum from the artifact.   

But this fakeness is digression and it's clear to me that this discussion will continue in circles from this point on.  You have no concern that the number of checksums is doubled such that each artifact is read two more times during repository creation nor do you have a concern about the overhead of checksum validation at install time is being doubled.  I can't address that lack of concern though logical arguments.

I'll look to @akurtakov and @mickaelistria to offer commentary.

laeubi:
> I don't like your talk about **fake** checksums. There's no such thing. A checksum is either correct or incorrect.

Sorry if it wasn't clear, I don't want to say that the **checksum** is fake, but the **metadata** is fake if I claim there is/was a sha25 sum for an artifact that was only published with a sha1. For example wit won't be hard for maven central to simply compute a sha256 (or even sha512) for older deployed artifacts, but that's not what actually happens!

> You have no concern that the number of checksums is doubled

As said above, nothing will be **doubled**, even nothing will change for P2 if you produce an update-site like before, P2 will just be able to **verify** more than today, so if there is any issue with having four extensions over two, then  something is really wrecked at P2 and we should fix this, but it is not the checksum to blame here.

> I can't address that lack of concern though logical arguments.

Because your only argument is that it *might* have an impact, I have given several ones that indicate it most probably is not, so if you have any source of datata that shows:

1. Extension Points are that incredible slow that adding one (or two) more implements will dramatically increase runtime
2. Computing checksums on modern CPU is magnitude slower than reading the data itself from disk or net I/O

then I can clearly understand the concerns and might provide mitigations as I already did with this PR by only ever computing the best checksum instead of every one that is around.

merks:
This loop will verify every registered checksum

https://github.com/eclipse-equinox/p2/blob/a7ea9baebea6f373a6bc02370b8f9fae87244e63/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java#L47-L83

This loop will compute every checksum:

https://github.com/eclipse-equinox/p2/blob/a7ea9baebea6f373a6bc02370b8f9fae87244e63/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java#L97-L112

Hence Tycho using the latter will double the cost of producing these sums.    Note again that the entire artifact is read from disk in full for each checksum calculation.  And once Tycho starts producing things like that then p2 will check twice as many checksums.  So it seems entirely clear that we do double the work.  

But it seems to me we could check just the strongest checksum making it work faster than it current does even with your changes (though you've argued best to check multiple ones) and we could improve the checksum calculation to read the bytes only once,a gain making it faster than what it currently does (but I'm not sure you've looked at the code to which I directed your attention earlier.

laeubi:
> This loop will verify every registered checksum

That's what I mentioned in (1) if that's slow, we probably should better cache this, for me it looks like the is no real need to compute this over and over again, but this is independent of if we have one or two or ... so if you agree that caching is a good idea I can include this in this change.

> Note again that the entire artifact is read from disk in full for each checksum calculation. 

From the code it seems they should be chained (so bytes are read once and passed along from one to the other), also here thats more a general problem that might needs to be fixed/improved anyways.

> But it seems to me we could check just the strongest checksum making it work faster than it current does even with your changes

That's what I already have included in this change, so yes that is possible. And yes I looked at that code already but you referenced code "read always full from disk", is actually only called when publishing the repository, so probably this should/could be optimized as well in a similar fashion. At the moment I have focused on the part where the repository is read and verified what is performed way more often than producing it. I would even stop producing md5 (or any deprecated ones) but then we are at "what about old consumers", so probably it is ok to spend more time on the producer side if we can make the consumer side fast enough and gain backward-compatibility.

But of course there is a chance I get the code wrong here...

merks:
@laeubi Thanks for changes to the checksum calculation.   That eliminates one of my concerns completely.

I think we could still consider reducing the number of checksums that are produced by default, i.e., in the two places it's called, we could ignore sha-1 and (eventually/in a few years md5) to not produce it:

image

But it will be good to hear the opinion of others on that.

Also, while one can validly argue that checking 3 or 4 checksums is better than checking only 1, it does feel a little like overkill; it's certainly not free though without an actual measurement like what @jukzi does, there is no basis a strong performance argument.

laeubi:
Yeah I have already think about it, I feel the problem is that currently there is no distinction between **verify** and **produce** of checksums .. I'll try to come up with another change, addressing this all in one PR.

As far as I understand sha256 was added with 9e5ac9146a99bc132960034e5312f1b449a31fd6 more than 4 years ago (even though git claims it was authored 14 years ago...) so probably it would be fine to stop producing md5 already ...

laeubi:
Regarding MD5 I opened: https://github.com/eclipse-equinox/p2/discussions/162

mickaelistria:
It seems to me that the priority is concretely not so useful: all checks are run anyway, aren't they? If so, I'd suggest we just remove the priority at the moment.

laeubi:
@mickaelistria you might better want to discuss about priority here: https://github.com/eclipse-equinox/p2/pull/163 the idea is to only check one an not all signatures.

mickaelistria:
OK, please rebase then.

laeubi:
Lets wait for https://github.com/eclipse-equinox/p2/pull/164 and I will rebase/resolve this then.

laeubi:
@merks @mickaelistria this should now finally be ready (and small enough) to be merged.

merks:
@laeubi Thanks for constructively addressing all of my concerns!  All the cases now perform better and the framework is more functional and flexible going forward.

mickaelistria:
Thanks!

PR #161, opened 2022-10-30
Do not read data multiple times when calculating checksum

laeubi:
Currently the file is read for each checksum algorithm. While calculating checksums is fast, disk I/O is often not.

This changes the method that calculates the hashes in a way that the data is only read once but chained to each algorithm.

As discussed here: https://github.com/eclipse-equinox/p2/pull/160

I think this can actually be speedup even more by caching the digest itself, but that's a different story...

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2019 Mykola Nikishov.
+ * Copyright (c) 2015, 2022 Mykola Nikishov and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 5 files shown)


bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2015, 2019 Mykola Nikishov.
+ * Copyright (c) 2015, 2022 Mykola Nikishov.
```
merks, line 2: It's just a nit, but this really should be 'Mykola Nikishov and others.'

merks (approved):
This is a really nice improvement.  I expect that the biggest cost is reading from disk and not the calculation itself.

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
laeubi: Done that!

laeubi:
Alright then I'll wait for the build and merge this!

PR #163, opened 2022-10-31
Always choose the most secure algorithm to check integrity

laeubi:
Currently always all checksums are checked for integrity of a file, this change the behavior by introducing a ranking and only checks the best matching one.

This is another spin-off from #160

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumVerifier.java
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumVerifier.java
@@ -34,13 +34,21 @@ final public class ChecksumVerifier extends MessageDigestProcessingStep {
 	final private String providerName;
 	final private String algorithmId;
 	private final boolean insecureAlgorithm;
+	private final int priority;
+
+	@Deprecated
+	public ChecksumVerifier(String digestAlgorithm, String providerName, String algorithmId, boolean insecure) {
+		this(digestAlgorithm, providerName, algorithmId, insecure, 0);
+	}
 
 	// public to access from tests
-	public ChecksumVerifier(String digestAlgorithm, String providerName, String algorithmId, boolean insecure) {
+	public ChecksumVerifier(String digestAlgorithm, String providerName, String algorithmId, boolean insecure,
+			int priority) {
 		this.algorithmName = digestAlgorithm;
 		this.providerName = providerName;
 		this.algorithmId = algorithmId;
 		this.insecureAlgorithm = insecure;
+		this.priority = priority;
 		basicInitialize(null);
 	}
 
@@ -95,5 +103,9 @@ final public class ChecksumVerifier extends MessageDigestProcessingStep {
 
 	public boolean isInsecureAlgorithm() {
 		return insecureAlgorithm;
+	}
+
+	public int getPriority() {
+		return priority;
 	}
 }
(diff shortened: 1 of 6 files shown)


bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumVerifier.java
```diff
@@ -34,13 +34,16 @@ final public class ChecksumVerifier extends MessageDigestProcessingStep {
 	final private String providerName;
 	final private String algorithmId;
 	private final boolean insecureAlgorithm;
+	private int priority;
```
merks, line 37: Why not final?

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumVerifier.java
```diff
@@ -34,13 +34,16 @@ final public class ChecksumVerifier extends MessageDigestProcessingStep {
...
 
 	// public to access from tests
-	public ChecksumVerifier(String digestAlgorithm, String providerName, String algorithmId, boolean insecure) {
+	public ChecksumVerifier(String digestAlgorithm, String providerName, String algorithmId, boolean insecure,
+			int priority) {
```
merks, line 41: If there was any expectation that someone might be using this class directly, leaving the original constructor to call this new one with defaults would be good.  But I doubt there are such...

merks (approved):
I think the changes are good as is.  I would be good to check that @mickaelistria also agrees is good to just check the most secure available checksum rather than all...

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumVerifier.java
laeubi: Done

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumVerifier.java
laeubi: Done.

merks: Thanks!

HannesWell: Would it be useful to mark that constructor for removal as well?
The new one that is public for tests only could also be marked as deprecated (but not for removal), that that anybody that calls it, without reading the comment, gets it strikedout in the IDE.

laeubi: It is internal.. so actually we can delete, modify, .. any time.

mickaelistria:
Yes, checking only the preferred algorithm seems good enough. Most other repository systems (eg rpm, Maven...) do only check 1 integrity algorithm.

PR #164, opened 2022-11-02
Disable md5 publishing

laeubi:
Currently all registered checksums are computed and published, but in some cases it might be usefully to be able to verify a checksum but we don't want to publish it.

This also includes disabling the md5 checksum, we publish sha-256 > 4 years and warn about md5 > 1 year now so it seems valid to stop publishing it now at all.


merks:
Should we perhaps also stop publishing (not start publishing) sha-1)?  I.e., perhaps sha-1 can/should be used to populate only when getting it from Maven but not when using p2 to publish a new repository given that both sha-256 and sha-512 are now published and only one of the latter two will ever be used at runtime such that a sha-1will never be consumed so is just useless overhead...

laeubi:
Yes if we eventually add sha-1 it could be disabled like md5 fro publishing.

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
```diff
@@ -115,7 +115,8 @@ public static IStatus calculateChecksums(File pathOnDisk, Map<String, String> ch
 		for (IConfigurationElement checksumVerifierConfiguration : ChecksumUtilities
 				.getChecksumComparatorConfigurations()) {
 			String id = checksumVerifierConfiguration.getAttribute("id"); //$NON-NLS-1$
-			if (checksumsToSkip.contains(id))
+			if (checksumsToSkip.contains(id)
+					|| !Boolean.parseBoolean(checksumVerifierConfiguration.getAttribute("publish"))) //$NON-NLS-1$
```
merks, line 119: Does getAttribute return the default value when the attribute is missing from the plugin.xml?

merks (approved):
It looks fine, but I'm not sure if the default value is returned from the getAttribute call...

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
mickaelistria: I don't think it does. `getAttribute` has no knowledge of the `.exsd` and the default value in exsd is supposed to reflect the default value implemented by consumers. So this condition would most likely lead to about no checksum being produced unless they explicitly set `publish=true`.

mickaelistria (changes requested):

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
laeubi: As we have a test-case for this lets see if that already fails.

laeubi: > As we have a test-case for this lets see if that already fails.

It does: `ProduceChecksumTest.testChecksums:45 SHA256 was not computed!` lets adjust this ... seems there are some test for md5 that need to be removed/adjusted.

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
```diff
@@ -115,7 +115,7 @@ public static IStatus calculateChecksums(File pathOnDisk, Map<String, String> ch
 		for (IConfigurationElement checksumVerifierConfiguration : ChecksumUtilities
 				.getChecksumComparatorConfigurations()) {
 			String id = checksumVerifierConfiguration.getAttribute("id"); //$NON-NLS-1$
-			if (checksumsToSkip.contains(id))
+			if (checksumsToSkip.contains(id) || !shouldPublish(checksumVerifierConfiguration))
```
merks, line 118: Testing `"false".equals(checksumVerifierConfiguration.getAttribute("publish"))` would suffice and be simpler...  But I don't feel strongly about that...

merks (approved):

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/artifact/IArtifactDescriptor.java
```diff
@@ -81,6 +81,17 @@ public interface IArtifactDescriptor {
...
+	/**
+	 * An artifact descriptor property (value "aartifact.sha-256") indicating the
+	 * SHA256 checksum of the artifact bytes in its native format (after processing
+	 * steps have been applied).
+	 *
+	 * @since 2.4
+	 * @see #ARTIFACT_CHECKSUM
+	 */
+
+	String ARTIFACT_SHA256 = "artifact.sha-256"; //$NON-NLS-1$
```
akurtakov, line 93: This one shouldn't be created. The new "dynamic" list of checksums is using ARTIFACT_CHECKSUM."algorytm" and there is even extension point to dyncmically produce new ones IIRC.

laeubi: @akurtakov yeah this was jsut more a guess as some tests are still referring to MD5 ... I'll try the other property now but tend to maybe revert this changes to the point where it is possible to disable algorithm but keep md5 for now as its a bit more work to adjust the test than initially thought and I can hardly guess if it not will not still have an impact.

laeubi:
Lets first get the option to disable  a checksum in:

- https://github.com/eclipse-equinox/p2/pull/165

laeubi:
@mickaelistria please update your review, @akurtakov thats how far I was able to get with the tests, I'll keep this PR just in case someone likes to build on top of it.

mickaelistria (approved):

akurtakov:
Just so I get things correct, what prevents this patch from getting merged is failing tests?

laeubi:
> Just so I get things correct, what prevents this patch from getting merged is failing tests?

Yes tests need to be adjust (or deleted or ...) because they depend on md5 calculated (what is no longer the case)

akurtakov:
Would you please reset this PR on master? aka removing my merge commit.

laeubi:
> Would you please reset this PR on master? aka removing my merge commit.

Done!

PR #172, opened 2022-11-17
Version bumps for 4.26 stream

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/661


merks:
@akurtakov It would be good to get this in for RC1 if that's still possible with the build delay...

akurtakov:
Better in RC1 than RC2. At least one good thing comes from the delay .

PR #197, opened 2022-12-03
Fix NPE in KeySigningInfoFactory and slightly modernize it

HannesWell:
Occasionally I see error reports about an NPE in `KeySigningInfoFactory` because the `bundlePoolRepository` is null.
Unfortunately I don't have any at hand now because I did the change a while ago but somehow did not submit it.

@merks can you tell if this change is reasonable?

And also apply some slight modernization/simplifications of the code.


HannesWell:
Thanks.

Issue #100, opened 2022-07-04
ProductFile should extract the locating of the instance area to a method to allow overrides by extenders


laeubi:
org.eclipse.equinox.internal.p2.publisher.eclipse.ProductFile currently uses 

Location instanceLocation = ServiceHelper.getService(Activator.getContext(), Location.class, Location.INSTANCE_FILTER);

this obviously fails outside of OSGi as the BC is null.

One way would be to extract it into a protected method so extenders like Tycho can override and provide alternative locations or to check that the BC is actually not null...

PR #199, opened 2022-12-04
ProductFile should extract the locating of the instance area to a method to allow overrides by extenders

laeubi:
Fix https://github.com/eclipse-equinox/p2/issues/100


akurtakov:
Change looks good but you have to bump the version.

Issue #214, opened 2023-01-24
NPE in .ChecksumUtilities.getChecksumComparatorConfigurations


laeubi:
Seen in Tycho because using static field that might not be initialized. Also i this should "never" happen, P2 could be more graceful here.

```
 Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.core.runtime.IExtensionRegistry.getConfigurationElementsFor(String)" because the return value of "org.eclipse.core.runtime.RegistryFactory.getRegistry()" is null
    at org.eclipse.equinox.internal.p2.artifact.processors.checksum.ChecksumUtilities.getChecksumComparatorConfigurations (ChecksumUtilities.java:99)
    at org.eclipse.equinox.internal.p2.artifact.processors.checksum.ChecksumUtilities.calculateChecksums (ChecksumUtilities.java:116)
    at org.eclipse.equinox.spi.p2.publisher.PublisherHelper.calculateChecksums (PublisherHelper.java:128)
    at org.eclipse.equinox.spi.p2.publisher.PublisherHelper.createArtifactDescriptor (PublisherHelper.java:117)
    at org.eclipse.equinox.spi.p2.publisher.PublisherHelper.createArtifactDescriptor (PublisherHelper.java:104)
    at org.eclipse.equinox.p2.publisher.eclipse.BundlesAction.generateBundleIUs (BundlesAction.java:1029)
    at org.eclipse.equinox.p2.publisher.eclipse.BundlesAction.perform (BundlesAction.java:915)
    at org.eclipse.equinox.p2.publisher.Publisher$ArtifactProcess.run (Publisher.java:288)
    at org.eclipse.equinox.p2.repository.artifact.spi.AbstractArtifactRepository.executeBatch (AbstractArtifactRepository.java:222)
    at org.eclipse.equinox.p2.publisher.Publisher.publish (Publisher.java:312)
    at org.eclipse.tycho.p2.publisher.AbstractMetadataGenerator.publish (AbstractMetadataGenerator.java:121)
    at org.eclipse.tycho.p2.publisher.AbstractMetadataGenerator.generateMetadata (AbstractMetadataGenerator.java:58)
    at org.eclipse.tycho.p2resolver.P2GeneratorImpl.generateMetadata (P2GeneratorImpl.java:181)
    at org.eclipse.tycho.plugins.p2.P2MetadataMojo.attachP2Metadata (P2MetadataMojo.java:188)
[...]

Issue #182, opened 2022-11-28
Allow ArtifactRepositoryFactory to be provided by OSGi services


laeubi:
Currently there is an extension point to provide `ArtifactRepositoryFactory`.

Currently the extension point has to provide the factory class and a filter (what is of type string), the filter can simply be a service property and `ArtifactRepositoryManager` is already created by a service component, so there should be not much of an issue with this.

The primary use-case would be tycho, where these P2 extension point is currently one reason we need special setups.

laeubi:
Closing this as not being relevant anymore (for me) as it seems more profitable to get rid of outdated P2 technology in Tycho and moving on to more modern and flexible approaches:
- https://github.com/eclipse-tycho/tycho/issues/2019

PR #215, opened 2023-01-24
Check for null of org.eclipse.core.runtime.RegistryFactory.getRegistry()

laeubi:
The API of `RegistryFactory.getRegistry()` states:

> return existing extension registry or null

but on several places there is no check for null.

This adds null checks on several places that previously would result in a NPE.

See https://github.com/eclipse-equinox/p2/issues/214

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java
@@ -96,7 +96,11 @@ public class ChecksumUtilities {
 	}
 
 	public static IConfigurationElement[] getChecksumComparatorConfigurations() {
-		return RegistryFactory.getRegistry().getConfigurationElementsFor(ARTIFACT_CHECKSUMS_POINT);
+		IExtensionRegistry registry = RegistryFactory.getRegistry();
+		if (registry == null) {
+			return new IConfigurationElement[0];
+		}
+		return registry.getConfigurationElementsFor(ARTIFACT_CHECKSUMS_POINT);
 	}
 
 	/**
@@ -184,6 +188,7 @@ public class ChecksumUtilities {
 		return properties;
 	}
 
+	@SuppressWarnings("deprecation")
 	private static void putLegacyMd5Property(String propertyNamespace, Map<String, String> checksums, HashMap<String, String> result) {
 		String md5 = checksums.get(ChecksumHelper.MD5);
 		if (md5 != null) {
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/provisional/p2/artifact/repository/processing/ProcessingStepHandler.java
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/provisional/p2/artifact/repository/processing/ProcessingStepHandler.java
@@ -97,7 +97,7 @@ public class ProcessingStepHandler {
 	 * step has not yet executed. If the step has executed the returned status
 	 * indicates the success or failure of the step.
 	 * @param deep whether or not to aggregate the status of any linked steps
-	 * @return the requested status 
+	 * @return the requested status
 	 */
 	public static IStatus getStatus(OutputStream stream, boolean deep) {
 		if (!deep)
@@ -175,8 +175,7 @@ public class ProcessingStepHandler {
 	}
 
 	public ProcessingStep create(IProvisioningAgent agent, IProcessingStepDescriptor descriptor, IArtifactDescriptor context) {
-		IExtensionRegistry registry = RegistryFactory.getRegistry();
-		IExtension extension = registry.getExtension(PROCESSING_STEPS_EXTENSION_ID, descriptor.getProcessorId());
+		IExtension extension = getExtension(descriptor);
 		Exception error;
 		if (extension != null) {
 			IConfigurationElement[] config = extension.getConfigurationElements();
@@ -197,6 +196,15 @@ public class ProcessingStepHandler {
 		return result;
 	}
 
+	private IExtension getExtension(IProcessingStepDescriptor descriptor) {
+		IExtensionRegistry registry = RegistryFactory.getRegistry();
+		if (registry == null) {
+			return null;
+		}
+		IExtension extension = registry.getExtension(PROCESSING_STEPS_EXTENSION_ID, descriptor.getProcessorId());
+		return extension;
+	}
+
 	public OutputStream createAndLink(IProvisioningAgent agent, IProcessingStepDescriptor[] descriptors, IArtifactDescriptor context, OutputStream output, IProgressMonitor monitor) {
 		if (descriptors == null)
 			return output;
@@ -213,7 +221,7 @@ public class ProcessingStepHandler {
 		}
 		if (steps.length == 0)
 			return previous;
-		// now link the artifact stream to the first stream in the new chain 
+		// now link the artifact stream to the first stream in the new chain
 		ArtifactOutputStream lastLink = getArtifactStream(previous);
 		if (lastLink != null)
 			lastLink.setFirstLink(previous);
--- a/bundles/org.eclipse.equinox.p2.engine/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.engine/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.engine;singleton:=true
-Bundle-Version: 2.7.500.qualifier
+Bundle-Version: 2.7.600.qualifier
 Bundle-Activator: org.eclipse.equinox.internal.p2.engine.EngineActivator
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/ActionManager.java
+++ b/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/ActionManager.java
@@ -7,7 +7,7 @@
  *  https://www.eclipse.org/legal/epl-2.0/
  *
  *  SPDX-License-Identifier: EPL-2.0
- * 
+ *
  *  Contributors:
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
@@ -42,7 +42,10 @@ public class ActionManager implements IRegistryChangeListener {
 
 	public ActionManager() {
 		this.touchpointManager = new TouchpointManager();
-		RegistryFactory.getRegistry().addRegistryChangeListener(this, EngineActivator.ID);
+		IExtensionRegistry registry = RegistryFactory.getRegistry();
+		if (registry != null) {
+			registry.addRegistryChangeListener(this, EngineActivator.ID);
+		}
 	}
 
 	public Touchpoint getTouchpointPoint(ITouchpointType type) {
@@ -91,7 +94,12 @@ public class ActionManager implements IRegistryChangeListener {
 	private synchronized Map<String, IConfigurationElement> getActionMap() {
 		if (actionMap != null)
 			return actionMap;
-		IExtensionPoint point = RegistryFactory.getRegistry().getExtensionPoint(EngineActivator.ID, PT_ACTIONS);
+		IExtensionRegistry registry = RegistryFactory.getRegistry();
+		if (registry == null) {
+			// maybe later...
+			return Map.of();
+		}
+		IExtensionPoint point = registry.getExtensionPoint(EngineActivator.ID, PT_ACTIONS);
 		IExtension[] extensions = point.getExtensions();
 		actionMap = new HashMap<>(extensions.length);
 		for (IExtension extension : extensions) {
--- a/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/TouchpointManager.java
+++ b/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/TouchpointManager.java
@@ -7,7 +7,7 @@
  *  https://www.eclipse.org/legal/epl-2.0/
  *
  *  SPDX-License-Identifier: EPL-2.0
- * 
+ *
  *  Contributors:
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
@@ -90,13 +90,16 @@ public class TouchpointManager implements IRegistryChangeListener {
 		}
 	}
 
-	// TODO: Do we really want to store the touchpoints? The danger is 
+	// TODO: Do we really want to store the touchpoints? The danger is
 	//	     that if two installations are performed simultaneously, then...
 	// TODO: Figure out locking, concurrency requirements for touchpoints.
 	private Map<String, TouchpointEntry> touchpointEntries;
 
 	public TouchpointManager() {
-		RegistryFactory.getRegistry().addRegistryChangeListener(this, EngineActivator.ID);
+		IExtensionRegistry registry = RegistryFactory.getRegistry();
+		if (registry != null) {
+			registry.addRegistryChangeListener(this, EngineActivator.ID);
+		}
 	}
 
 	/*
@@ -136,7 +139,12 @@ public class TouchpointManager implements IRegistryChangeListener {
 		if (touchpointEntries != null)
 			return touchpointEntries;
 
-		IExtensionPoint point = RegistryFactory.getRegistry().getExtensionPoint(EngineActivator.ID, PT_TOUCHPOINTS);
+		IExtensionRegistry registry = RegistryFactory.getRegistry();
+		if (registry == null) {
+			// maybe later...
+			return Map.of();
+		}
+		IExtensionPoint point = registry.getExtensionPoint(EngineActivator.ID, PT_TOUCHPOINTS);
 		IExtension[] extensions = point.getExtensions();
 		touchpointEntries = new HashMap<>(extensions.length);
 		for (IExtension extension : extensions) {
--- a/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/CertificateChecker.java
+++ b/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/CertificateChecker.java
@@ -496,7 +496,11 @@ public class CertificateChecker {
 		Map<ByteBuffer, Set<Bundle>> keys = new LinkedHashMap<>();
 
 		// Load from the extension registry.
-		for (IConfigurationElement extension : RegistryFactory.getRegistry()
+		IExtensionRegistry registry = RegistryFactory.getRegistry();
+		if (registry == null) {
+			return Map.of();
+		}
+		for (IConfigurationElement extension : registry
 				.getConfigurationElementsFor(EngineActivator.ID, "pgp")) { //$NON-NLS-1$
 			if ("trustedKeys".equals(extension.getName())) { //$NON-NLS-1$
 				String pathInBundle = extension.getAttribute("path"); //$NON-NLS-1$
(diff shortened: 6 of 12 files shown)


merks:
Failing "gracefully" sounds good though at this point in the code it's determining the collection of available checksum providers so if it can't access the registry, it will gracefully find no providers and gracefully produce no checksums, quietly such that one might not notice the repositories it creates are "broken" because of lack of checksums.   Then one might not realized there is a problem that needs to be fixed in Tycho and one might produce repositories without checksums without realizing it.  And then one has no clue why there are not checksums.  

So I'm not convinced that gracefully ignoring the lack of all registered extensions is actually an improvement rather than simply masking a "severe" problem that really needs to be addressed to ensure proper behavior.  I.e., it seems to me this change makes the problem worse because you will have no clue that there is a problem.  Surely this problem, missing extension registry,  needs to be fixed in Tycho in order to behave correctly, does it?

merks:
I think somewhere in Tycho there should be an "early assertion" that org.eclipse.core.runtime.RegistryFactory.getRegistry() is not null because I don't expect p2 to function properly if nothing that it registers is actually available/found.  I don't feel that masking this problem in the p2 framework is a good idea.

laeubi:
> it seems to me this change makes the problem worse because you will have no clue that there is a problem

If checksums are missing Tycho will print a warning, also technically speaking there is no difference between a missing registry or one without any extension for this extension point.

So if we really want to enforce at least one checksum provider (or warn about or ...) it should be handled with a different code path so probably return an error status (I have seen this on some places in P2). But this seems more a task for the callers of the `ChecksumUtil`, e.g. `ChecksumUtilities.getChecksumVerifiers(IArtifactDescriptor, String, Set<String>)` already prints a warning if it can't verify, `ChecksumUtilities.calculateChecksums(File, Map<String, String>, Collection<String>)` returns an OK status and leave it to the caller to decide...

laeubi:
> I don't feel that masking this problem in the p2 framework is a good idea.

As said the API is clear about that it might return null. Dereferencing a possible `null` value unconditionally is just a bug, and P2 already handle the null case on some places.

merks:
You didn't just change it so that there will be no checksum provider (that you then have to detect elsewhere).  There will also be no processing steps, no actions, no touchpoints, no IUAnalyzers, and no repository providers.  Surely nothing can function properly in this case none of the expect registered contributions are not available.  

I strongly suggest that you do not change this but rather fix the underlying problem in Tycho.

laeubi:
> I strongly suggest that you do not change this but rather fix the underlying problem in Tycho.

There is no problem in Tycho, there is a bug in P2 that it runs into a documented `null` return value. Whether or not the remaining functionality is useful for Tycho is not a concern of P2.

merks:
Please provide a test case to reproduce the bug of a null registry.  (Is that a p2 bug or an Equinox bug?)  The test case should define the 'graceful" behavior of p2 in that scenario, demonstrating that p2 functions is some useful/graceful way as a result of your additional changes, e.g., that p2 is still able to load a repository despite the null registry.

iloveeclipse:
Looking on the stack from issue #214 I assume OSGI wasn't started, and no one set the default registry provider.
Expecting this p2 code to do anything meaningful is most likely wrong.

I think instead of null checks the code around `org.eclipse.equinox.p2.publisher.eclipse.BundlesAction.perform()` should check for running platform, because the bundle provides extensions and without OSGI there are no extensions.

laeubi:
> Please provide a test case to reproduce the bug of a null registry. (Is that a p2 bug or an Equinox bug?)

A null registry is not a bug see 
https://github.com/eclipse-equinox/equinox/blob/f1790545f4b69eafd0cb21172bb598a28140a230/bundles/org.eclipse.equinox.registry/src/org/eclipse/core/runtime/RegistryFactory.java#L64-L70

>  Returns the default extension registry specified by the registry provider.
> **May return <code>null</code>** if the provider has not been set or if the
>  registry has not been created.

laeubi:
> I think instead of null checks the code around `org.eclipse.equinox.p2.publisher.eclipse.BundlesAction.perform()` should check for running platform, because the bundle provides extensions and without OSGI there are no extensions.

Several code in P2 do not require a "running platform" and that's actually a good thing, but it is bad to assume that methods never return null if they are documented to do so.

iloveeclipse:
> So the bug is that one assumes non-null is returned from a method that declares it may return null...

Cristoph, the problem is that the code **relies** on OSGI registry to work, but runs outside of OSGI.
The null handling in this case is really nonsense, since it will not function without the registry anyway.
The code should check early enough it is used in wrong context, see my comment above.

laeubi:
> Cristoph, the problem is that the code relies on OSGI registry to work, but runs outside of OSGI.

That's not true, it relies on the `IExtensionRegistry` (what can run without OSGi), but such a registry might not be there (yet), still it can work (as a no-op in that case) without failing with a NPE....

iloveeclipse:
> but it is bad to assume that methods never return null if they are documented to do so.

In the OSGI context this should never be the case. If OSGI is up and running, there will be no null registry. If it is not running, the code will fail to function anyway.

laeubi:
> > but it is bad to assume that methods never return null if they are documented to do so.
> 
> In the OSGI context this should never be the case. If OSGI is up and running, there will be no null registry. If it is not running, the code will fail to function anyway.

Yes, but code should simply behave correctly to the API contracts the call regardless if it "never happens" ... it does not fail to function, it simply behaves as if there is a registry with no extensions, what is a valid state.

merks:
The point is that when there are no extensions (which p2 definitely does not expect, because these are it's own extensions) that p2 will fail to function for any number of other downstream reasons.  But you could provide a test case that shows that p2 actually functions in some reasonable (graceful) way in the case  of a null registry, assuming you believe the preceding assertion to be false.

laeubi:
Running into a NPE obviously is not what one desires and this change prevent potential null-pointer access, and if it can "never happen" I don't understand why you care? 

All tests pass and this shows it does not introduce any regressions to be more defensive, anyone who thinks P2 should yell out more about this possible incident or think the test.coverage should be improved is free to do so.

iloveeclipse:
> Yes, but code should simply behave correctly to the API contracts the call regardless if it "never happens" ... it does not fail to function, it simply behaves as if there is a registry with no extensions, what is a valid state.

I would agree otherwise regarding API contracts and NP checks, but why this code that relies on extension registry to work should "gracefully" work without registry is beyond my understanding. On contraire, the code MUST fail saying "what do you want from me without extensions"?

laeubi:
> On contraire, the code MUST fail saying "what do you want from me without extensions"?

It already *does* if it would not fail before with a NPE...

iloveeclipse:
> It already _does_ if it would not fail before with a NPE...

Where does it? My point is: instead of running into NPE (or gracefully avoid it) it should check wrong state before, not later.

merks:
I think a test case, like a picture, is worth 1000 words...

akurtakov:
I'll try to be the "translator" here :) . So these are "extensions" aka checksums, comparators, processing steps, listeners, touchpoints, etc. and core p2 resolving will happen with or without all the bells and whistles. This is usable for cases like tycho which do not run in a full OSGi context. I don't see how a testcase would show that.

akurtakov:
It has been yesterday when I looked at reducing jdt.ls dependencies and while no p2 involved it shows similar symptoms. There are many components/bundles in Eclipse that rely on everything being there although they could be far more widely used if some *defensive* patterns are applied. 
Note that this is not a new thing - SWT and JFace already work in similar way and doing so will increase usability of artifacts produced and I can only hope contributions by people that use them even outside of Eclipse IDE.

merks:
I don't believe these things to be bells and whistles, but rather the core functionality of p2!

Moreover the motivation for these changes, as shown in this stack trace of the associated issue, shows that Tycho is trying to compute checksum, which is one of the "bells and whistles" that will simply not work. 

https://github.com/eclipse-equinox/p2/issues/214

I believe that issue should not be masked gracefully...

But then the argument is that this is just fixing bugs in p2; nothing to do with Tycho.  In that case it seems more than reasonable to ask for a test case that demonstrates **and defines** the expected graceful p2 behavior.

laeubi:
> Where does it? 

I'll just give one example
https://github.com/eclipse-equinox/p2/blob/0dfc4b4b4e0bb917841b967da2ac68ba45e51ad8/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/provisional/p2/artifact/repository/processing/ProcessingStepHandler.java#L177-L198

- at line 179, this will faill with a NPE if the registry is null (what is for whatever reason and even it should never happen is allowed from the API)
- with this change it will not fail with a NPE (what actually is the last resort for the JVM to say we do something really bad, C/C++ will just core-dump the process in such a case)
- but it will advance to the line 192, and depending on the serverity, return a status that describes the (maybe not desired) state.

> I believe that issue should not be masked gracefully...

There is nothing "masked", computing checksums is just one part, and in the failing case it actually *wants* checksums but it does not *require* them because it calls the `BundleAction` to get the basic P2 meta-data of an artifact and checksums are part of that metadata but they are _optional_.

Also I'd like to give a "picture" here:
https://github.com/eclipse-equinox/p2/blob/0dfc4b4b4e0bb917841b967da2ac68ba45e51ad8/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/processors/checksum/ChecksumUtilities.java#L103

So if anyone thinks that the **caller** of `ChecksumUtilities` decides **WRONG** if there are no checksums produced, please fix it, even provide a testcase for that, whatever is desired, but it is obviously not the contract of that function to throw a NPE in any case nor will it help compared to the state where there are no checksum providers at all (e.g. because someone has deleted all of them from the codebase, or moved them into a seperate bundle not part of the product or whatever...)

merks:
You are making changes that I think are inappropriate and misleadingly wrong.   This isn't implementing "graceful failure", e.g., an exception or even a log message saying "The extension registry is unavailable so not much of anything is going to work", but rather you are implementing some (in my opinion) ill-defined behavior, i.e., behave as if all extensions of have been deleted deliberately.  Then you refuse to create a test case demonstrating that this empty registry behavior results in functionally usable behavior to justify all the changes, while at the same time insisting others should fix things and also should create test cases to back up their arguments or concerns:

> So if anyone thinks that the caller of ChecksumUtilities decides WRONG if there are no checksums produced, please fix it, even provide a testcase for that, whatever is desired, but it is obviously not the contract of that function to throw a NPE in any case nor will it help compared to the state where there are no checksum providers at all (e.g. because someone has deleted all of them from the codebase, or moved them into a seperate bundle not part of the product or whatever...)

I feel as if it won't matter what anyone says about anything at this point...

laeubi:
> You are making changes that I think are inappropriate and misleadingly wrong.

It is wrong to not check for null if null is allowed by the API contract, and I really don't get why this is considered "inappropriate" beside that you feel you need too "teach" other by punish them with scary NPEs instead of giving suitable error messages.

> This isn't implementing "graceful failure", e.g., an exception or even a log message saying "The extension registry is unavailable so not much of anything is going to work", but rather you are implementing some (in my opinion) ill-defined behavior,

This behavior is already implemented elsewhere in the code, so I don't really get what we are talking about here, Tycho is using the full feature set of P2 regarding resolving without a single line of this requiring OSGi at all, so I really don't get why this should be considered "non functional", so please share the code-paths of this change where a error handling or whatever is not already performed and I'll remove those, but for the whole sake of *thinking* something *might* be wrong does not bring things forward, especially not for this trivial case I solely provided because no one else at P2 seems care writing code that works *always* in contrast to for *my use cases* and I need to get consistent error or graceful degradation instead of runtime exceptions that are clearly an indication of bad coding patterns.

laeubi:
> I feel as if it won't matter what anyone says about anything at this point...

Probably because instead of looking at the changed code (just note it never claims to be "graceful" anywhere, it all talks about missing null checks!) it is debated if we should NOT fix inappropriate null dereferencing in P2 because it _feels wrong_. So yes, from my point of view that does matter regardless of who says this, because this issue is not how P2 should handle the case better (it already does it to some degree on many places) where is might find things are missing.

So if anyone thinks that the handling of null is faulty or incomplete, please just comment on the changed code and it will be considered, improving P2 in general for situation where missing extensions might cause harm, please open a separate discussion or issue.

merks:
Your issue states "P2 could be more graceful here" so you did use that term and hence we do expect this commit to make p2 be more graceful.  We did state that we would expect some graceful **failure** in the case where expected contributions are unavailable for unexpected reasons, so there is no point in repeating that if it's rejected as irrelevant.   Even the suggestion of logging an error/warning appears rejected.  We are concerned that p2 will be ill-behaved in this scenario and will be silently masking the underlying failure/problem, making such a problem hard to track down, but that too appears rejected.  So it seems nothing short of quietly behaving as if all contributions are deliberately deleted will be an appropriate solution and that asking for a test case to demonstrate this produces useful behavior is out of the question.  So it appears all has been said and done...

akurtakov:
@tjwatson As a project lead you have to intervene here.

laeubi:
> Your issue states "P2 could be more graceful here" so you did use that term and hence we do expect this commit to make p2 be more graceful.

And while I implemented the fix I noticed that my initial report was not complete and there is no need to be graceful as the API clearly states that `null`  is allowed and thus fixed some other locations as well, changed my commit message and the PR description to state the real intend to fix possibly null pointer access :-)

Then it was brought up that:

1. It actually should never happen
2. And can never happen
3. And even if it happen we should not do anything but blame the users of P2 doing things wrong
4. ...

what now reminds me about the [Six Stages Of Debugging](https://www.reddit.com/r/ProgrammerHumor/comments/97o4pp/six_stages_of_debugging/)


So if that confuses anyone sorry for that ... I think I have rephrased it multiple times here that the goal is to fix the NPE not P2 behavior in general (graceful or not).

> Even the suggestion of logging an error/warning appears rejected. 

As said, this already happens, and I did not rejected it if someone feels the current state is inappropriate or incomplete it can be enhanced but not as part of this change, as this is irrelevant as long as the NPE prevents further processing anyways.

Also if you like [you can setup plain unit-tests for P2](https://tycho.eclipseprojects.io/doc/master/TestingBundles.html#maven-surefire-plugin) without a problem, I just doubt that this helps in any way here to convince those claiming it can never happen as the same arguments will be brought up again to argue against the test as was done here that it can never be null...

merks:
In terms of the motivating issue, haven't we gone from "I want to generate checksums but I get an NPE because the registry is null." to "I want to generate checksums, but none are generated, and I don't know why because everything appeared to work fine"?   I guess that's exactly what you want, but to me that seems questionable...

Furthermore, that's just the one change here.  Other cases will also produce surprising, unexplained, and diagnosed behaviors, e.g., I can't load a repository because there are no registered factories...

Anyway, it seems Tycho must have it this way, though I do wonder and question (sorry) how you get things working at all in the meantime.  To me the stack trace you show in the issue directs you to fix something (ensure the registry is initialized) so that the code can properly produce checksums.  And yes, an NPE is definitely **not pretty** so producing a more descriptive failure would be potentially helpful.

In the end, it's most likely that Tycho and someone contributing to Tycho will get bitten by this change rather than that it has any real impact on the use of p2 in a regular Equinox runtime.  It would just be nice if you could explain or motivate why this is so important to change and how this helps Tycho function correctly nevertheless when there is no registry because I just keep imagining nothing working properly in this case, which is perhaps misguided...

iloveeclipse:
> In neither case anything is "wrong" or "not working", as I hope is proven by all the hundreds of builds performed every day using Tycho... 

Same (or even more) applies to Eclipse, used also without Tycho by thousands of builds every day. 

> What defiantly _IS_ wrong that enormous efforts are obviously required to fix simple programming errors telling peoples actually don't want that fix and should fix "something else".

You should consider difference between "one man project" and a platform.

We (still) have more than one developer interested on having stable platform, so having discussion on a review should be natural. In this context & regarding Tycho proven stability: you already forgot this weekend build failures? Could it be they are coming from missing reviews?

All what we (me and Ed) wants is that a change makes sense not only for Tycho but for the entire platform, so we look at the patch and express our concerns. Having a different opinion **is** natural.

merks:
Please keep in mind that we all, including you clearly, have the best of intentions for this project and we all want p2 to be good for all use cases, including Tycho.   We do not intentionally frustrate you purely for the sake of being annoying.  I can see that you are frustrated and I'm very sorry for that.  It would be good to assume good intent and to remain focused on technical questions.

laeubi:
> We (still) have more than one developer interested on having stable platform, so having discussion on a review should be natural. 

As mentioned above, reviews are welcome... but no one reviews the *code* nor can explain how not running into a NPE would decrease stability.

>  In this context & regarding Tycho proven stability: you already forgot this weekend build failures? Could it be they are coming from missing reviews?

Platform is free to use the stable 3.0.1 Tycho release, Snapshots are for discovering new functions (that what platform did) and making them more bullet-prove, none of the change has failed a single test or even the platform-verification build that runs with each change and the fix for the issue was in some code even present in Tycho 3.x, and even though the demands on Tycho are often high the support (including for fixed in P2 that would help Tycho stabilize things including this) are generally low.

But of course everyone is invited to review and improve things, and as @merks already noticed the integration-test suite is quite powerful so unless something is not covered (as with platform using ancient orbit updatesites what caused the recent failure) its quite safe to just start.

>  It would be good to assume good intent and to remain focused on technical questions.

So what are the technical concerns in preventing null object dereferencing here? So look at the code(!) make suggestions on how to improve and assume _good intent_ of the PR author to improve even small parts of the code.

merks:
I have suggested throwing an exception with a meaningful message. I've seen no direct response to that suggestion.  I also suggested logging an error or warning, also with no response so far.  

Here is a concrete example of my concern.  You are implementing a change such that AbstractRepositoryManager.findMatchingRepositoryExtensions(String, String) will return an empty array.  So I look at the call hierarchy and see that this will cause a failure on the selected line of code:

image

image

I.e., it will fail with repository not found.  I think an earlier NPE, while definitely very ugly, is much more helpful for tracking down and fixing the underlying problem than a much later "repository not found exception" for a repository that does actually exist.

My sense is that several of the changes will lead to such subsequent downstream failures where the failure downstream will be much more mysterious than the explicit  failure of find expected registered contributions.

Even in the motivating Tycho example, you argue that the artifacts may or may not get checksums but that this doesn't matter because either way is not an error.  But I can imagine that one write Tycho tests with one set of expectations and then suddenly the registry is available and the test results of different than before.  But Tycho is not my business.  

Making sure that the changes in p2 are sensible is more than reasonable and the example I describe above does not seem sensible to me.  I could analyze each change you are making to see what would stop behaving properly...

laeubi:
> AbstractRepositoryManager.findMatchingRepositoryExtensions(String, String) will return an empty array. So I look at the call hierarchy and see that this will cause a failure on the selected line of code:

And that is _EXACTLY_ what I would expect as this is _EXACTLY_ what I get when trying to use an unknown repository type, if no repository type is registered, then obviously _EVERY_  type of repository could not be loaded.

And as this "usually never happens" I also don't see how this would introduce any regression and/or problem, so "no registry" is not a meaningful message as it essentially the concern of the system provider if/how it provides a registry and then must scope with the results.

Please also see specifically this enhancement request:
- https://github.com/eclipse-equinox/p2/issues/182

> Making sure that the changes in p2 are sensible is more than reasonable and the example I describe above does not seem sensible to me. 

Because your only goal is to prove me "wrong" and constructing hypothetical use-case no one ever asked for and no one ever claimed this will or should solve. instead you are arguing that Tycho do not want this because it will fail Tycho expectations what is simply not the case...

> My sense is that several of the changes will lead to such subsequent downstream failures where the failure downstream will be much more mysterious than the explicit failure of find expected registered contributions.

There are simply no such "mysterious downstream failures" and there are no "suddenly" changing tests, but don't mind, this discussion clearly shows that P2 is lacking any chance to evolve in the future so the choice is to get rid of it in tycho-core:
- https://github.com/eclipse-tycho/tycho/issues/2019

And I already applied another hack to Tycho:
- https://github.com/eclipse-tycho/tycho/pull/2020

but just keep in mind that mostly that will mean Tycho one day will simply drop P2 support at all as we did with the tycho-p2-product already if no one else likes to maintain outdated P2 stuff anymore that simply is a nightmare for any adopters.

so closing this as ir does not seems sensible (with best intend)...

merks:
I made two constitutive suggestions, but again with zero response. I showed a concrete example of where I think the new behavior will be worse than the old behavior, in general, and even for Tycho, explaining in detail why I think that this makes tracking down resulting problems even more difficult.

In the end, it appears as if suggestions, differences of opinion, and critique are all viewed as unacceptable and somehow purely offensive.  But please do not to ascribe offensive motivations to me.  Most normal people would have stopped communicating with you much earlier in the day out of a sense of futility.  I keep trying explain how I see things technically yet I feel like I've been effectively punched in the face simply for speaking.  Surely it must be possible to keep a technical focus without taking or giving personal offense...

tjwatson:
@laeubi I'm coming very late to this 12 hour old conversation.  I have to say, this all escalated very fast and I am finding it hard to follow with all the elevated opinions.

I think one of the big things missing in all of this are actual tests of p2 in a non OSGi environment.  With my limited exposure to all of p2, that seems to be at the heart of what is going wrong here.  If p2 is to support environments that are not using OSGi as the runtime then surely we need tests for that.

merks:
I too thought tests would help and suggested that early in the conversation:

https://github.com/eclipse-equinox/p2/pull/215#issuecomment-1401497058

I know @laeubi does tonnes of really great work and that he is super smart. I wish he would not take offense where none is intended.  I know he means well and I can only ask again that he please assume the same of me.

If I recall correctly, there is a way to use the Equinox registry stand-alone.  Perhaps that would be helpful in this type of scenario.

Issue #230, opened 2023-03-21
CVE-2021-41033: Provide improved support for dealing with http protocol in p2


merks:
Improve the handling of http protocol by the p2 framework.

The following links provide background detail:

- https://nvd.nist.gov/vuln/detail/CVE-2021-41033
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=575688

Some work has been done to log warnings when the use of http protocol is detected in some limited contexts:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=576428

Such warnings were considered undesirable in some environments, e.g., inside a corporate firewall, so support was provided to allow logged warnings to be disabled:

https://github.com/eclipse-equinox/p2/issues/123

To provide both maximum flexibility and maximum security, I propose to expand the current approach to fully address CVE-2021-41033 and beyond.

I propose that we focus on guarding/monitoring p2's transport layer as a whole such that **all** internet access is taken into consideration, including the following mainstream use cases:
- p2 repository content metadata and artifact metadata, composite children, and repository reference URLs, e.g.,
  - https://download.eclipse.org/releases/2023-03/202303151000/p2.index
- mirror server URLs, i.e., the `p2.mirrorsURL` property in the artifact metadata, e.g.,
  - https://www.eclipse.org/downloads/download.php?format=xml&file=/releases/2023-03/202303151000
- mirror URLs, e.g.,
  - https://ftp.fau.de/eclipse/releases/2023-03/20230315100
- artifacts URLs, e.g.,
  - https://ftp.fau.de/eclipse/releases/2023-03/202303151000/plugins/bcpg_1.72.2.jar

I propose three modes of http handling:
1. **allow**
  - This is basically what we have now, i.e., log a warning but allow that warning to be suppressed.  The warning will be logged for all http access, not simply once per repository and only for repository URLs.
3. **redirect**
  - Any http URI will be redirected/written to use https such that no http request ever reaches out into the network.
5. **block**
  - Any http URI will be blocked as if the network were down/inaccessible.

One important question is "what should be the default behavior in the absence of an explicit choice of one of those three"? I propose the framework defaults to **block** http access.  Product/application builders (or even users) can then choose whether to **allow** such access or **redirect** such access; they will be alerted to that choice by the exception that blocks the access.  (Note that within a corporate firewall, it might well be the case that redirection to https causes failures due to missing root certificates so **allow**, while not ideal, may be needed in some limited environments where security is ensured via external mechanisms.)

[...]

PR #231, opened 2023-03-22
Provide improved support for dealing with http protocol in p2

merks:
https://github.com/eclipse-equinox/p2/issues/230

### DIFF
--- a/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/messages.properties
+++ b/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/messages.properties
@@ -80,4 +80,5 @@ connection_to_0_failed_on_1_retry_attempt_2=Connection to {0} failed on {1}. Ret
 UnableToRead_0_TooManyAttempts=Unable to read repository at: {0}. Too many failed login attempts.
 UnableToRead_0_UserCanceled=Unable to read repository at: {0}. Login canceled by user.
 RepositoryTransport_failedReadRepo=Error while reading from repository: {0}.
-unsafeHttp=Using unsafe http transport to retrieve {0}, see CVE-2021-41033. Consider using https instead.
+RepositoryTransport_unsafeHttp=Using unsafe http transport to retrieve {0}, see CVE-2021-41033. Consider using https instead.
+RepositoryTransport_unsafeHttpBlocked=Using unsafe http transport to retrieve {0} is blocked, see CVE-2021-41033. Only https is supported. Use the system property -Dp2.httpAction=redirect to automatically redirect to https or -Dp2.httpAction=allow to permit unsafe access.
--- a/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
+++ b/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
@@ -15,13 +15,14 @@ package org.eclipse.equinox.internal.p2.transport.ecf;
 
 import java.io.*;
 import java.net.*;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import org.eclipse.core.runtime.*;
 import org.eclipse.ecf.core.identity.IDCreateException;
 import org.eclipse.ecf.core.security.ConnectContextFactory;
 import org.eclipse.ecf.core.security.IConnectContext;
 import org.eclipse.ecf.filetransfer.*;
+import org.eclipse.equinox.internal.p2.core.helpers.LogHelper;
 import org.eclipse.equinox.internal.p2.repository.*;
 import org.eclipse.equinox.internal.p2.repository.Credentials.LoginCanceledException;
 import org.eclipse.equinox.internal.p2.repository.Messages;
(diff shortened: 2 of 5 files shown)


merks:
@iloveeclipse 

FYI, this will make the default behavior to block all http access.  You can use one of these to change that behavior:
- -Dp2.httpAction=redirect
- -Dp2.httpAction=allow

If you use the latter, your p2.skipRepositoryProtocolCheck will continue to work as before.

Any concerns?

iloveeclipse:
> Any concerns?

Not from me, and thanks for the note. Ideally this information would be added to the commit message.

merks:
I will augment the commit message to include some of the details described in the associated issue.

https://github.com/eclipse-equinox/p2/issues/230

merks:
@laeubi 

I prefer to stick to a very simple approach that focused on http, i.e., the specific topic of the CVE.

@mickaelistria 

Any concerns?  I think this is the next logical step past simply logging warnings, which has been in place long enough for folks to have taken action if possible.

akurtakov:
I think I'm done with this one, just waiting for verification build to confirm. Don't wait if you're ready to push this one I'll rebase if needed.

merks:
I'm going out for a while and will give folks time to comment until I get back this afternoon, so no problem.  Thanks for all the hard work on the Java 17 changes!

mickaelistria:
No concern from my side.

akurtakov:
This patch generates extra compile warning in I-build https://download.eclipse.org/eclipse/downloads/drops4/I20230322-1800/testResults.php#PluginsErrors . Direct link is https://download.eclipse.org/eclipse/downloads/drops4/I20230322-1800/compilelogs/plugins/org.eclipse.equinox.p2.transport.ecf_1.4.0.v20230322-1430/@dot.html#OTHER_WARNINGS

laeubi:
> I prefer to stick to a very simple approach that focused on http, i.e., the specific topic of the CVE.

I just wanted to note that even though the CVE only mentions http, I think this is too limited as a mirror list can contain **any** protocol specifically ftp urls even though http is the most common today, so the same exits for `ftp` vs `ftps`

Also I think that `p2.httpAction` is quite a confusing name, what does "action" refer to? Please note that "action" in http refers to the [http-verb](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods)  and is likley not what one wants to express here.

So if one really likes to limit this to http I would more like something like "enforceHttps" with the values:

- never
- always
- rewrite / redirect

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/messages.properties
```diff
@@ -80,4 +80,5 @@ connection_to_0_failed_on_1_retry_attempt_2=Connection to {0} failed on {1}. Ret
 UnableToRead_0_TooManyAttempts=Unable to read repository at: {0}. Too many failed login attempts.
 UnableToRead_0_UserCanceled=Unable to read repository at: {0}. Login canceled by user.
 RepositoryTransport_failedReadRepo=Error while reading from repository: {0}.
-unsafeHttp=Using unsafe http transport to retrieve {0}, see CVE-2021-41033. Consider using https instead.
\ No newline at end of file
+RepositoryTransport_unsafeHttp=Using unsafe http transport to retrieve {0}, see CVE-2021-41033. Consider using https instead.
+RepositoryTransport_unsafeHttpBlocked=Using unsafe http transport to retrieve {0} is blocked, see CVE-2021-41033. Only https is supported. Use the system property -Dp2.httpAction=redirect to automatically redirect to https or -Dp2.httpAction=allow to permit unsafe access.
```
laeubi, line 84: I would strip the "Only https is supported" as it is obviously supported but blocked as indicated by the first sentence.

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
```diff
@@ -43,8 +44,33 @@ public class RepositoryTransport extends Transport {
 
 	public static final String TIMEOUT_RETRY = "org.eclipse.equinox.p2.transport.ecf.retry"; //$NON-NLS-1$
 	private static Map<URI, Integer> socketExceptionRetry = null;
+	private final Set<URI> loggedURIs = ConcurrentHashMap.newKeySet();
 	private IProvisioningAgent agent = null;
 
+	private enum HttpAction {
```
laeubi, line 50: As mentions I thin `HttpAction` is a bit confusing, Maybe `HttpsRule` or something like that?

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
```diff
@@ -268,6 +302,33 @@ public long getLastModified(URI toDownload, IProgressMonitor monitor)
 		throw new AuthenticationFailedException();
 	}
 
+	private URI getSecureLocation(URI location) throws ProvisionException {
```
laeubi, line 305: I see one problem with doing the check here is that it will only apply to this specific implementation of a transport, so if there are alternative repository transports used the check is not performed anymore!
Would it be possible to move this from the transport to the application layer so "unsafe" URIs not even reach the transport?

laeubi:
Too bad I didn't notice that this already was merged :-(

Nerveless I think this should be adjusted, the name is way to confusing and handling this on the transport implementations makes it open to the check to be bypassed, either by intend or by accident.

PR #233, opened 2023-03-22
Fix the implementation of socket timeout retry handling.

merks:
Ensure that the map used to save the per-URI retry count is thread safe and ensure that the retry count resets back to zero after a period of time.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
+++ b/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
@@ -15,8 +15,10 @@ package org.eclipse.equinox.internal.p2.transport.ecf;
 
 import java.io.*;
 import java.net.*;
-import java.util.*;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import org.eclipse.core.runtime.*;
 import org.eclipse.ecf.core.identity.IDCreateException;
 import org.eclipse.ecf.core.security.ConnectContextFactory;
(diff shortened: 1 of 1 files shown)


merks:
@laeubi 

I noticed that the implementation maintaining a per-URI retry count via map was not thread safe (nor was the map itself thread safe).   It also seems odd that the retry count can never go back to zero, so I implemented a simple mechanism so that after 10 minutes the retry count will go back to 1.   At least its thread safe now, but still feels kind of ugly, though less horrible than before.  I don't have a better idea for how to fix what was arguably a bit of a hack...

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
```diff
@@ -43,7 +44,30 @@
 public class RepositoryTransport extends Transport {
 
 	public static final String TIMEOUT_RETRY = "org.eclipse.equinox.p2.transport.ecf.retry"; //$NON-NLS-1$
-	private static Map<URI, Integer> socketExceptionRetry = null;
+
+	private static class Retry {
+		// 10 minutes.
+		long LIFETIME = 10 * 60 * 1000;
```
laeubi, line 50:
```suggestion
		long LIFETIME = TimeUnit.MINUTES.toMillis(10);
```

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
```diff
@@ -43,7 +44,30 @@
...
+		long expiration;
+		int count;
+
+		public Retry() {
+			expiration = System.currentTimeMillis() + LIFETIME;
+		}
+
+		int increment() {
```
laeubi, line 58:
```suggestion
		synchronized int increment() {
```
Otherwise this is not concurrency safe!

laeubi (changes requested):
Maybe it should reset the count once a successful attempt was made as well?

laeubi:
The last modified and download methods seems like a good place to reset if they succeed without the need to do it everywhere.

merks:
This still seems to be blocked.  I'd like to merge as is because while not perfect it's better than what's there now, which is definitely not thread safe and will definitely timeout forever after the first time.  If you want to further improve it, you can do so.  But I have limited time for that purpose.

laeubi (approved):
Concurrency issues seems to be addressed now.

merks:
BTW, I think are correct that one could do stuff in download and last modified if one guarded all the return points.   The stream method isn't so clear.  But you and I probably both have bigger fish to fry.  😄

Issue #238, opened 2023-03-30
Remove deprecated pack200 support


akurtakov:
It has been deprecated 2 years ago in https://bugs.eclipse.org/bugs/show_bug.cgi?id=572043, deprecation period ended and now that Java 17 is required there is not even theoretical chance for this to be useful anymore.

akurtakov:
Removal documentation updated via https://github.com/eclipse-platform/eclipse.platform.common/pull/130

iloveeclipse:
@akurtakov : there are PDE test fails in https://download.eclipse.org/eclipse/downloads/drops4/I20230330-1800/testResults.php related to this removal.

Do you plan / can work on them? 

See https://download.eclipse.org/eclipse/downloads/drops4/I20230330-1800/testresults/html/org.eclipse.pde.build.tests_ep428I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

akurtakov:
Let me look into them.

PR #239, opened 2023-03-30
Remove pack200 API

akurtakov:
It's tracked in https://github.com/eclipse-equinox/p2/issues/238 . This is first iteration removing the already non-functional API.


mickaelistria (approved):
All looks good.

PR #245, opened 2023-04-08
Improve the handling of revoked PGP keys

merks:
In general, handle a revoked key as if all signatures that use it are potentially tampered and require explicit review and additional extra confirmation.

Check revocation during download in PGPSignatureVerifier, but only log error rather than fail the download so that p2 does not try to repeatedly download the same artifact because it assumes that it's corrupted in some way. Revocation is handled later in the CertificateChecker for presentation to the user.

DefaultPGPPublicKeyService.getVerifiedRevocationDate handles exceptions when checking revocation as if the key were revoked:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=581453

The CertificateChecker removes from the set of trusted keys all PGP keys that are revoked.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=581452

I.e., a revoked key can never be trusted, even if present in the trusted preferences, but rather will always be presented to the user via the TrustCertificateDialog.  The TrustCertificateDialog is made aware of revoked PGP keys and presents detailed information to the users. Moreover, accepting to trust revoked keys involves an extra confirmation so that the user is explicitly aware of the risk.


merks:
@laeubi  

The build has been stuck like this for an hour:

> + mvn clean verify --batch-mode --fail-at-end -Dmaven.repo.local=/home/jenkins/agent/workspace/p2_PR-245/.m2/repository -Pbuild-individual-bundles -Pbree-libs -Papi-check -Dcompare-version-with-baselines.skip=false -Dmaven.test.error.ignore=true -Dmaven.test.failure.ignore=true -Dproject.build.sourceEncoding=UTF-8 -T1C

Is this a Maven 3.9.0 problem?

laeubi:
Not that I'm aware of, but I think no one has yet tested to build P2 with maven 3.9.x

merks:
It finally started.  Go figure.  If it isn't one thing, it's another!  😱

laeubi:
I see some fatal infra issues latley, so maybe the job was affected, I just killed the old run and started a new one.

laeubi:
By the way, maybe you want to add some screenshots on how this looks like with a revoked certificate.

merks:
Thanks for helping restart the build!

It's a good idea to include more documentation in this PR.

Here's the prompt when all the keys are revoked:

image

Here's what happens if you select them and trust them anyway:

image

This happens despite these keys being trusted previously:

image

merks:
It would also be good if the preference page also showed revocation information:

image

laeubi:
Looking at the screens I think that "a revoked PGP" (second sceeenshot) might not be understood by all users, so maybe the popup dialog should somehow be more explicit (e.g. use "Signer Key (Type PGP)" or something like that.

In the first screen it would be good to have an orange background to make the user more aware that something dangerous might happen.

Independently from that (maybe in a new PR) it would be very useful to somehow have a clickable link that leads to some page that explains how to verify similar to that one:

- https://www.secureideas.com/blog/how-to-verify-pgp-signatures
- https://www.apache.org/info/verification.html

even though I think it would be best to link to an eclipse website, maybe even mentioning the official project key signatures.

Another goody would be to have a button where I can click on to query a keyserver for more information.

merks:
I'll consider these things, but many of these things are per-existing issues.  E.g., I suspect the majority don't even know what a PGP key is; I know two years ago I would have had no clue.  There is a help link, but unfortunately, but that only links to the workbench's help and there is no workbench in the installer.  Given the user will explicitly be told that what they are accepting might tampered if they eagerly try to get ride of the dialog, I think colors might be overkill; there's already a warning symbol too.  Not only that, when we make something orange, someone will complain they can't read in it dark theme.  😨 

I really would like a link to a key server though and could just use https://keyserver.ubuntu.com/ ...

I will revisit such things in a separate PR when the other outstanding CVEs are addressed....

Issue #235, opened 2023-03-25
CVE-2021-41037: p2 metadata (even without artifacts) can be an attack vector


merks:
This issue's description and proposed remediation will be edited as this issue evolves.  It is currently a draft work in progress.

Background:

- https://bugs.eclipse.org/bugs/show_bug.cgi?id=577029
- https://nvd.nist.gov/vuln/detail/CVE-2021-41037

The CVE has this description:

> In Eclipse p2, installable units are able to alter the Eclipse Platform installation and the local machine via touchpoints during installation. Those touchpoints can, for example, alter the command-line used to start the application, injecting things like agent or other settings that usually require particular attention in term of security. Although p2 has built-in strategies to ensure artifacts are signed and then to help establish trust, there is no such strategy for the metadata part that does configure such touchpoints. As a result, it's possible to install a unit that will run malicious code during installation without user receiving any warning about this installation step being risky when coming from untrusted source.

The bugzilla has description:

> p2 metadata, even without artifacts, can affect the application in unsafe way: touchpoints can run some local commands, change the command line that will be used to start the IDE (ie changing the JVM, adding some agents...). It seems relatively easy to inject malicious changes in p2 metadata.
> Unlike artifacts, installing a unit that consists only of metadata doesn't show any warning to users. So they may install malicious code without any warning. There is no concept of "trusting" metadata like it exists for artifacts.
>
> On the long run, the solution would be to sign metadata as well, like other package managers -eg rpm- do.

One thing seems clear up front.  The oceans will need to significantly heat up before all p2 metadata will be properly and consistently signed and before the frameworks will properly check that.  Often there can be a content.xml, content.jar, and/or content.xml.xz, all of which will need to be signed, though they typically contain (and should contain) the same semantic content.  As such, all the ways of building these would need to produce such signatures and all such signatures would need to follow the artifact (metadata files) that they sign, e.g., the various of mirror applications and scripts would need to be aware of these new things, often needing to resign the resulting content.  And finally, the consuming applications would all need to have installed a framework that checks these signatures.  The whole ecosystem would need to support producing such things, not just Eclipse projects, i.e., those project with enough resource to deal with the associated workload. It seems to me like a giant hurdle, and then the actual concrete benefit versus the cost is questionable at best.

[...]

PR #249, opened 2023-04-15
Support tracking and reviewing content origin during installation

merks:
Tracking tracking is provided by ProvisioningContext's getInstallableUnitSources and getArtifactSources which do a deep analysis of all repositories, including children of composites and repository reference to track installable units and artifact keys back to their originating repository.

This is used during the Engine's Collect phase as implemented by the new AuthorityChecker, which also manages underlying preferences for keeping track of trusted authorities.

IInstallableUnitUIServices provide support for prompting the user and is implemented by ValidationDialogServiceUI to display TrustAuthorityDialog which is very similar in design to TrustCertificateDialog.

The TrustPreferencePage is enhanced to provide user-level access to the trusted authorities preferences.

https://github.com/eclipse-equinox/p2/issues/235


bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/metadata/spi/IInstallableUnitUIServices.java
```diff
@@ -0,0 +1,94 @@
...
+		 */
+		public boolean isSave() {
+			return save;
+		}
+
+		/**
+		 * Return whether to always trust all authorities, both during this operation
+		 * and for all future operations.
+		 *
+		 * @returns whether to always trust all authorities, both during this operation
```
iloveeclipse, line 71: See https://download.eclipse.org/eclipse/downloads/drops4/I20230416-0550/testresults/html/org.eclipse.releng.tests_ep428I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

```
../../../rt.equinox.p2/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/metadata/spi/IInstallableUnitUIServices.java:71: error: unknown tag: returns
		 * @returns whether to always trust all authorities, both during this operation
		   ^
1 error
```

PR #258, opened 2023-04-24
Notify user about updates for write-protected components

### LINKED ISSUE #254: Automatic update scheduler does not notify user about updates that target write-protected  components
The Eclipse update check is not showing any notification of available updates when the user does not have sufficient access privileges. This issue occurs even when automatic updates are enabled in the preferences. 

### Test steps

- Install an [old Eclipse Package](https://www.eclipse.org/downloads/packages/release/2022-12/r)
- Disable write access to the installation directory 
  - or switch to user account with limited privileges, that lacks such access
- Start Eclipse
- Configure Preferences/Install/Update/Available Software Site to add [the latest Eclipse update site](https://download.eclipse.org/releases/2023-03)
- Invoke Help/Check for Updates
- Wait for pending jobs to complete
- Verify that some update are present, but none are available due to "Insufficient access privileges to apply this update."
- Cancel the update dialog.
- Configure Preferences/Install/Update/Automatic Updates/Update schedule/Look for updates each time Eclipse IDE is started
- Apply and restart
### Expected Result
The user should receive some kind of notification indicating that the system administrator may be able to install updates.
### Actual Result
The update check is completely silent, no indication of available update is given.
### Entrypoint
The entrypoint is org.eclipse.equinox.internal.p2.ui.sdk.scheduler.AutomaticUpdater.validateIusToUpdate() - it ignores all non-installable updates.
### Suggested Fix
Warn user about all updates. Enable new behavior by Preferences. 
This will allow security-sensitive products to  inform the user about updates, without annoying users in  non-secure contexts.

basilevs:
Fixes #254.
Function is enabled with JVM argument `-Dorg.eclipse.equinox.p2.ui.sdk.scheduler.warnLocked=true`

### Previous behavior
Automatic update check fails silently on write-protected installations.
Explicit check for updates shows message "Insufficient access privileges to apply this update".
Screenshot 2023-04-22 at 11 41 48

### New behavior
If automatic update check fails due to insufficient privileges, user is presented with a new popup:
Screenshot 2023-04-24 at 22 01 27

On click, an update dialog shown, just like for explicit "Check For Updates", allowing user to see which updates can't be installed.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/.settings/org.eclipse.core.resources.prefs
+++ b/bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
--- a/bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/src/org/eclipse/equinox/internal/p2/ui/sdk/scheduler/AutomaticUpdateMessages.java
+++ b/bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/src/org/eclipse/equinox/internal/p2/ui/sdk/scheduler/AutomaticUpdateMessages.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2016 IBM Corporation and others.
+ * Copyright (c) 2008, 2023 IBM Corporation and others.
  *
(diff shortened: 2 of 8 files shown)


merks:
This looks well done.

basilevs:
The size and location of the popup are different from the original. Investigating.

vgcpge:
I've tested on an integration build and the new functionality works as intended.

merks:
Thanks again for your high quality contribution!!

PR #260, opened 2023-05-05
Improve UI Consistency

BeckerWdf:
Don't use icons on buttons on dialogs. I know no other dialog of Eclipse Platform that uses icons on buttons on dialogs. Also the Eclipse UI Guideline
(https://eclipse-platform.github.io/ui-best-practices/) does not say anything about icons on buttons.

They where added with
https://bugs.eclipse.org/bugs/show_bug.cgi?id=577193

So to be consistent with all the other dialogs: Let's remove these icons.


BeckerWdf:
@merks: Can you pls. review?

BeckerWdf:
The TrustPreference Page:
Screenshot 2023-05-05 at 14 23 36

Other pages with similar buttons:
Screenshot 2023-05-05 at 14 23 19

it's even not consistent with pages from the same repos:
image

BeckerWdf:
> Note such unicode icons also are used in dialogs that pop up...

What do you mean with "dialogs that pop up"?

BeckerWdf:
> @mickaelistria started adding such things and is likely to have an opinion on the matter.

Correct.

BeckerWdf:
> https://github.com/eclipse-equinox/p2/blob/84389c702b8d27417cc11446c187acd76653b4db/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/messages.properties#L293
> 
> Some of the unicode icons are used in descriptive messages or within table cell labels that should stay as they are...

I have seen that. And I think this is somewhat ok. So I did not change it. But on buttons it looks odd from my point of view.

BeckerWdf:
> Here's an example:
> 
> https://github.com/eclipse-equinox/p2/blob/84389c702b8d27417cc11446c187acd76653b4db/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/messages.properties#L316
> 
> I've tried to keep newer dialogs consistent with that, e.g., this one:

Appreciate that you seeking consistency.

I can add additional removals regarding this once we agreed on the direction.

vogella:
+1 for the removal of such icons.

BeckerWdf:
@merks:
I just pushed another commit.
You know the code best. Did I still miss some of these icons on buttons?

merks:
Yes, it looks like you found all the correct button labels. 

It's really great to have someone looking at consistency issues like this!

laeubi:
By the way I think they mostly look "odd" because the label is centered, if one wants such icons the label should be right aligned.

merks:
Given most folks all to agree it looks odd and inconsistent, and given I'm about to change these same files, I will merge this now.

BeckerWdf:
> By the way I think they mostly look "odd" because the label is centered, if one wants such icons the label should be right aligned.

Yes. But it's correct that they are centered. See all the other buttons that are right next to a table.

BeckerWdf:
> Given most folks all to agree it looks odd and inconsistent, and given I'm about to change these same files, I will merge this now.

Thanks.

PR #271, opened 2023-06-06
use try-with-resource

jukzi:


mickaelistria:
Reference  continuous-integration/jenkins/pr-head  succeeds, let's merge.
thanks!

PR #280, opened 2023-06-14
Avoid reuse of SimpleDateFormat

jukzi:
SimpleDateFormat is not thread-safe.

  SimpleDateFormat.format(new Date());
returns same string as immutable and thread-safe
  DateTimeFormatter.format(new Date().toInstant())


HannesWell:
> DateTimeFormatter.format(new Date().toInstant())

I know it is after the fact, but instead of `new Date(long).toInstant()` can we just use `Instant.ofEpochMilli(long)`?

jukzi:
> I know it is after the fact, but instead of `new Date(long).toInstant()` can we just use `Instant.ofEpochMilli(long)`

yes

HannesWell:
> > I know it is after the fact, but instead of `new Date(long).toInstant()` can we just use `Instant.ofEpochMilli(long)`
> 
> yes

Do you want to provide a PR to adjust that? If not, I can change it.

PR #281, opened 2023-06-14
Use new static IPath factories and constants

HannesWell:
Use the new IPath factory methods introduced in https://github.com/eclipse-equinox/equinox/pull/228 in P2.

With this PR the `org.eclipse.core.runtime.Path` does not occur anymore in P2's entire code base.


HannesWell:
Jenkins builds succeeds and the GH workflow failures seems to be unrelated, since they occur the same way on the master.

PR #282, opened 2023-06-15
Remove unused composite parser parameter

laeubi:
The CompositeParser `BundleContext` parameter is actually unused, removing this also allows to cleanup further things that previously required static access to the Activator class.

FYI @merks do you see any issues with that?


merks (approved):
I don't find any other uses of it, in Oomph nor in the CBI p2 analyzer or aggregator.  One could of course deprecate the old constructor, but I doubt anyone is using it...

laeubi:
As it is an internal class I think deprecation do not really help much and if there is no platform consumer I think its better to clean this right now.

merks:
I agree, there is no strong argument in favor of maintaining a deprecated constructor.

PR #287, opened 2023-06-16
Use declarative service annotations for p2.core

laeubi:


bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/DefaultAgentProvider.java
```diff
@@ -18,10 +18,12 @@
 import java.util.Hashtable;
 import org.eclipse.equinox.p2.core.*;
 import org.osgi.framework.*;
+import org.osgi.service.component.annotations.Component;
 
 /**
  * Default implementation of {@link IProvisioningAgentProvider}.
  */
+@Component(service = IProvisioningAgentProvider.class, name = "org.eclipse.equinox.p2.di.agentProvider")
```
HannesWell, line 26: Until now I assumed that a DC component name is not really relevant and for informational purpose only. Is that assumption is false? If not I suggest to not specify the name explicitly here and below.

HannesWell (approved):
Looks good to me, just one remark regarding the names.

bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/DefaultAgentProvider.java
laeubi: I tried to stay 100% same as before jsut in case, the name is relevant for two cases:

1. The name is used if one can configure a component
2. It is used in the Component Service

so I don't think it matters much here but also don't harm anything.

HannesWell: Yes it definitely does not harm. I was just wondering if we could save it, if it is not used at all, but if there is a chance that it could be useful its fine to keep it 👍🏽

PR #289, opened 2023-06-17
Remove static Activator reference access

laeubi:
Currently two classes only access the "instance" of the Activator class in a static way what often is a source of failures when using P2 in other context as a potential NPE can occur.

This replaces the static access with some OSGi low level access that will fullfill a similar purpose but do not require to tie these to the activator class.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/Activator.java
+++ b/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/Activator.java
@@ -21,18 +21,15 @@ import java.util.Hashtable;
 import org.eclipse.core.runtime.*;
 import org.eclipse.equinox.internal.p2.core.helpers.LogHelper;
 import org.eclipse.equinox.p2.core.*;
-import org.eclipse.osgi.framework.log.FrameworkLog;
 import org.osgi.framework.*;
-import org.osgi.util.tracker.ServiceTracker;
 
 public class Activator implements BundleActivator {
 	public static IAgentLocation agentDataLocation = null;
 
-	public static BundleContext context;
+	private static BundleContext context;
 	private static final String DEFAULT_AGENT_LOCATION = "../p2"; //$NON-NLS-1$
 	public static final String ID = "org.eclipse.equinox.p2.core"; //$NON-NLS-1$
 
-	private static Activator instance;
 	// Data mode constants for user, configuration and data locations.
 	private static final String NO_DEFAULT = "@noDefault"; //$NON-NLS-1$
 	private static final String NONE = "@none"; //$NON-NLS-1$
@@ -51,8 +48,6 @@ public class Activator implements BundleActivator {
 
 	private IProvisioningAgent agent;
 	private ServiceRegistration<IAgentLocation> agentLocationRegistration = null;
-
-	ServiceTracker<FrameworkLog, FrameworkLog> logTracker;
 
 	/**
 	 * NOTE: This method is copied from LocationHelper in org.eclipse.osgi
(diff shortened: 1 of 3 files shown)


merks:
The Oomph build is already failing like this as of today:

> 
> [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:2.7.5:compile (default-compile) on project org.eclipse.oomph.p2.core: Compilation failure: Compilation failure: 
> [ERROR] /opt/public/hipp/homes/genie.oomph/workspace/integration/git/plugins/org.eclipse.oomph.p2.core/src/org/eclipse/oomph/p2/internal/core/ProvisioningAgentProvider.java:[38] 
> [ERROR] 	super.activate(org.eclipse.equinox.internal.p2.core.Activator.getContext());
> [ERROR] 	                                                              ^^^^^^^^^^
> [ERROR] The method getContext() is undefined for the type Activator
> [ERROR] /opt/public/hipp/homes/genie.oomph/workspace/integration/git/plugins/org.eclipse.oomph.p2.core/src/org/eclipse/oomph/p2/internal/core/ProvisioningAgentProvider.java:[67] 
> [ERROR] 	URIUtil.fromString(org.eclipse.equinox.internal.p2.core.Activator.getContext().getProperty(OSGIUtils.PROP_CONFIG_AREA) + "../p2"), ""); //$NON-NLS-1$ //$NON-NLS-2$
> [ERROR] 	                                                                  ^^^^^^^^^^
> [ERROR] The method getContext() is undefined for the type Activator
> [ERROR] 2 problems (2 errors)
> [ERROR] -> [Help 1]
> [ERROR] 
> [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
> [ERROR] Re-run Maven using the -X switch to enable full debug logging.
> [ERROR] 
> [ERROR] For more information about the errors and possible solutions, please read the following articles:
> [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
> [ERROR] 
> [ERROR] After correcting the problems, you can resume the build with the command
> [ERROR]   mvn <args> -rf :org.eclipse.oomph.p2.core

So I thought, oh well, I'll just use the public static field then, but now you are making it private, in which case I'll have to use reflection to access it, in which case it can get broken and I won't notice right away. 

It's not really necessary to make Activator.context private.

laeubi:
@merks can you explain why you need access to the p2.core activator at all?
You can take a look at the changed `Tracing` how to get access to the bundle context in a similar way, but be aware that using the `BundleContext` of another bundle is actually bad style so oomph probably should use its own one.

The ultimate goal would be to get rid of any code in Activator anyways, so making this private seems a good start.

merks:
I need it to implement this class:

https://github.com/eclipse-oomph/oomph/blob/master/plugins/org.eclipse.oomph.p2.core/src/org/eclipse/oomph/p2/internal/core/ProvisioningAgentProvider.java

It needs to behave like p2's own DefaultAgentProvider.

The ultimate goal is to ensure this keeps working as well as it has for the past 7 years.

laeubi:
See https://github.com/eclipse-oomph/oomph/pull/1

laeubi:
> I need it to implement this class

By the way It seems you are mostly performing some optimizations / fixups that might be useful to be part of P2 directly, so you should consider to contribute the fixes here as well.

Beside that, I think setting the `agentDataLocation` field has not really the desired effect, as at this point p2 has already read the value of the field and you maybe better want to register a `IAgentLocation` with a higher service rank instead?

merks:
Cool.  Thank you!  😁

laeubi:
> Cool. Thank you! grin

So do you think with the change we can merge this?

merks (approved):
It looks fine. I already tested your Oomph improvement so that’s great and compatible with older releases too. Thanks so much!

PR #293, opened 2023-06-22
Do not include Bouncycastle bundles in p2.core feature

akurtakov:
This allows updating deps without touching features.


akurtakov:
Prereq for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1171

Issue #308, opened 2023-08-22
Could not instantiate step:org.eclipse.equinox.p2.processing.Pack200Unpacker


laeubi:
```
[ERROR] All attempts to read artifact osgi.bundle,org.apache.batik.i18n,1.14.0.v20210324-0332 failed:
[ERROR]    An error occurred while transferring artifact canonical: osgi.bundle,org.apache.batik.i18n,1.14.0.v20210324-0332 from repository https://download.eclipse.org/releases/2021-09/202109151000:
[ERROR]       Retry another mirror:
[ERROR]          download from https://mirror.ibcp.fr/pub/eclipse/releases/2021-09/202109151000/plugins/org.apache.batik.i18n_1.14.0.v20210324-0332.jar failed
[ERROR]    An error occurred while transferring artifact packed: osgi.bundle,org.apache.batik.i18n,1.14.0.v20210324-0332 from repository https://download.eclipse.org/releases/2021-09/202109151000:
[ERROR]       Result of processing steps.:
[ERROR]          Could not instantiate step:org.eclipse.equinox.p2.processing.Pack200Unpacker
```

Now e removed pack200 support it seems there are some problems if it is tried to transfer artifacts from some mirrors, we should somehow add special handling that such requests are just a noop / redirected to the plain artifact. I'm not even sure if the packed artifact is actually used or if it is just a problem of reading that metadata.

merks:
The error messages suggest that **all** attempts have failed, including 'artifact canonical`.  We've disabled this support years ago.  Can it really be the case that only just now we a problem?  I suspect there was a network/host problem when this occurred and that the problem can't be reproduced...

laeubi:
Yes it seems to be a bad choice of a mirror here, **but** what makes me a bit curious is the `Could not instantiate step:org.eclipse.equinox.p2.processing.Pack200Unpacker` so as there is no more Pack200 support I would expect we actually never try to use it anyways...

merks:
I agree it should no longer try to use something that will definitely not work and is known not to work.   

I seem to recall I had problems with mirroring as well where it should also skip trying to mirror the packed things rather than fail.

akurtakov:
This was one hell of a removal in so many places and it looks like I missed some.

akurtakov:
I can't find Pack200Unpacker anywhere in the aggregator so I'm quite puzzled who/where defines this step. Is this error seen in the IDE or on the console/Tycho?

akurtakov:
[...]

PR #310, opened 2023-08-23
Ignore packed artifacts when parsing

akurtakov:
Prevents issues when packed artifacts are returned by iterator but they can no longer be handled.
Fixes https://github.com/eclipse-equinox/p2/issues/308


merks (approved):
This looks good.  Thanks!

akurtakov:
I prefer to land this for M1 but I'm open to the idea of pushing it now if people consider it very important fix. @tjwatson wdyt?

akurtakov:
Jenkins succeeded and I didn't retrigger GH actions.

PR #312, opened 2023-09-02
Add new method IQueryResult.stream()

HannesWell:
This improves the interoperability of `IQueryResult` and the Java's modern Stream-API.

Additionally Clean-up `IQueryResult` and its implementations in a separate first commit.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.metadata/.settings/.api_filters
+++ b/bundles/org.eclipse.equinox.p2.metadata/.settings/.api_filters
@@ -8,4 +8,12 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="src/org/eclipse/equinox/p2/query/IQueryResult.java" type="org.eclipse.equinox.p2.query.IQueryResult">
+        <filter id="404000815">
+            <message_arguments>
+                <message_argument value="org.eclipse.equinox.p2.query.IQueryResult"/>
+                <message_argument value="stream()"/>
+            </message_arguments>
+        </filter>
+    </resource>
 </component>
--- a/bundles/org.eclipse.equinox.p2.metadata/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.metadata/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.metadata;singleton:=true
-Bundle-Version: 2.7.100.qualifier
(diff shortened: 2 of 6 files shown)


bundles/org.eclipse.equinox.p2.metadata/.settings/.api_filters
```diff
@@ -8,4 +8,12 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="src/org/eclipse/equinox/p2/query/IQueryResult.java" type="org.eclipse.equinox.p2.query.IQueryResult">
```
HannesWell, line 11: IQueryResult is not marked with `@noimplement/@noextend` but I wonder if clients are really supposed to create own implementations?

merks (approved):
Looks fine.

bundles/org.eclipse.equinox.p2.metadata/.settings/.api_filters
laeubi: Tycho has some implementations of `IQueryResult`, I don't see why one should not allow this?

merks: People do what they're not supposed to do, so if they're allowed, one can assume that someone will to it.  😱 

What about a default method that does toSet().stream() so that by default there is always an implementation?

HannesWell: OK with custom queryable this makes sense.
Tycho even already has a stream() method in its own implementation, but it only needs to be made public when depending on the next Eclipse Release, so I think this is fine.
https://github.com/eclipse-tycho/tycho/blob/master/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/ListQueryable.java#L80-L83

laeubi: `IQueryable` already implements `Iterable` so using the iterator for default implementations seems a better choice to prevent creation of an intermediate set.

HannesWell: > What about a default method that does toSet().stream() so that by default there is always an implementation?

That's possible. But as Christoph suggested it could use the `Spliterator` of the `IQueryResult`, like it is currently done in `QueryResult`.

`Iterable` itself unfortunately does not provide a stream() method (if it would this PR would be pointless).

HannesWell: > > What about a default method that does toSet().stream() so that by default there is always an implementation?
> 
> That's possible. But as Christoph suggested it could use the `Spliterator` of the `IQueryResult`, like it is currently done in `QueryResult`.

Done.

PR #314, opened 2023-09-03
Add IRepository.contains(T) method

HannesWell:
This methods provides a more convenient and in most cases faster way to test if a repository contains an element, compared to querying the repository for it.
`IArtifactRepository` already has a contains() method so this effectively only adds it for `IMetadataRepository` and generalizes it.

The implementation is based on the query() or everthing() implementation of the corresponding class, if possible.

The first of the two commits contains the clean-ups in the the `IRepository` and `IArtifactRepository` interfaces.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.repository/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.repository/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.repository;singleton:=true
-Bundle-Version: 2.7.100.qualifier
+Bundle-Version: 2.8.0.qualifier
 Bundle-Activator: org.eclipse.equinox.internal.p2.repository.Activator
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
@@ -33,7 +33,7 @@ Export-Package: org.eclipse.equinox.internal.p2.persistence;
    org.eclipse.equinox.p2.ui,
    org.eclipse.equinox.p2.updatesite",
  org.eclipse.equinox.internal.provisional.p2.repository,
- org.eclipse.equinox.p2.repository;version="2.0.0",
+ org.eclipse.equinox.p2.repository;version="2.1.0",
  org.eclipse.equinox.p2.repository.artifact;version="2.3.0",
  org.eclipse.equinox.p2.repository.artifact.spi;version="2.0.0",
  org.eclipse.equinox.p2.repository.metadata;version="2.0.0",
@@ -65,7 +65,7 @@ Import-Package: javax.crypto,
  org.eclipse.equinox.p2.core.spi;version="[2.0.0,3.0.0)",
  org.eclipse.equinox.p2.metadata;version="[2.4.0,3.0.0)",
  org.eclipse.equinox.p2.metadata.expression;version="[2.0.0,3.0.0)",
- org.eclipse.equinox.p2.query;version="[2.0.0,3.0.0)",
+ org.eclipse.equinox.p2.query;version="[2.1.0,3.0.0)",
  org.eclipse.equinox.security.storage,
  org.eclipse.osgi.service.debug,
  org.eclipse.osgi.util;version="1.1.0",
--- a/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/IRepository.java
+++ b/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/IRepository.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2007, 2011 IBM Corporation and others.
+ *  Copyright (c) 2007, 2023 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -7,7 +7,7 @@
  *  https://www.eclipse.org/legal/epl-2.0/
  *
  *  SPDX-License-Identifier: EPL-2.0
- * 
+ *
  *  Contributors:
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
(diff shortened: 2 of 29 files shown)


bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/IRepository.java
```diff
@@ -222,4 +222,13 @@ public interface IRepository<T> extends IAdaptable, IQueryable<T> {
...
 	String setProperty(String key, String value, IProgressMonitor monitor);
+
+	/**
+	 * Returns true if this repository contains the given element.
+	 *
+	 * @param element the element to query
+	 * @return true if the given element is already in this repository
+	 * @since 2.8
+	 */
+	boolean contains(T element);
```
laeubi, line 233: could this be default implemented?

merks: An "everything" query can be like this
```
new ExpressionMatchQuery<>(IInstallableUnit.class, ExpressionUtil.TRUE_EXPRESSION)
```
so I think it would be tricky to create a correct query even for that case though I suppose with some instanceof tests one could do that...

laeubi: what about

`query(QueryUtil.createMatchQuery(ExpressionUtil.TRUE_EXPRESSION)).toSet().contains(element)`

merks: See org.eclipse.equinox.p2.query.QueryUtil.ALL_UNITS and note the type of that query:
```
IQuery<IInstallableUnit> ALL_UNITS = QueryUtil.createMatchQuery(ExpressionUtil.TRUE_EXPRESSION);
```

HannesWell: The problem is that here it is operated on the generic type T and not `IInstallableUnit` and I didn't found a way to create a query for any type (generic type probably won't work since that info is lost at runtime).

But it could be default implemented in the IMetadataRepository.

laeubi: As far as I can see the class is not actually used if expression is true, but maybe its even simpler:

```
return !query.query(new IQuery<T>() {

			@Override
			public IQueryResult<T> perform(Iterator<T> iterator) {

				while(iterator.hasNext()) {
					T t = iterator.next();
					if(Objects.equals(t, item)) {
						return new CollectionResult<>(List.of(t));
					}
				}
				return new CollectionResult<>(List.of());
			}

			@Override
			public IExpression getExpression() {

				return ExpressionUtil.TRUE_EXPRESSION;
			}
		}, null).isEmpty();
```

laeubi: Thinking a bit more I now think it would even be better to have  	`boolean contains(T element); ` in the `IQueryable` interface with the last provided default implementation...

HannesWell: I also tought about moving contains() to queryable but discarded it due to the high implementation work.

But with a good default impl this would be much less work and it can be overriden where suitable.
I'll try it this evening.

HannesWell: Added it now and also provided more direct implementations in most complementing classes.

What do you think about keeping the contains(boolean) override in the `IRepository` in order to provide a bit better javadoc?
Unfortunately we cannot remove the contains method from the `IArtifactRepository` interface since this would be a binary incompatible change.

bundles/org.eclipse.equinox.p2.repository/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.repository;singleton:=true
-Bundle-Version: 2.7.100.qualifier
+Bundle-Version: 2.8.0.qualifier
```
laeubi, line 5: API tools now assumes no new API here so 2.7.200 seems sufficient.

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/IRepository.java
laeubi: Looks good to me if it helps to have better javadoc why not, but I think you need to default implement (and delegate to IQueryable impl) this as well as otherwhise you remove the default impl inherited from the parent otherwhise.

merks: It also should have an `@Override` if it's already declared in the base interface.  And it would seem odd to add a method to API without an `@since`; I don't think I've ever seen an `@since` that includes a micro version.  I mention this because of the other comment about there being no additional API; perhaps/probably a default method would additional API.

bundles/org.eclipse.equinox.p2.repository/META-INF/MANIFEST.MF
merks: Maybe this assumption will change if/when a default method is implemented...

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/IRepository.java
HannesWell: > Looks good to me if it helps to have better javadoc why not, but I think you need to default implement (and delegate to IQueryable impl) this as well as otherwhise you remove the default impl inherited from the parent otherwhise.

Good point. Done that.



> It also should have an `@Override` if it's already declared in the base interface.

Yes that, that's right as well. Done, too.

> And it would seem odd to add a method to API without an `@since`; I don't think I've ever seen an `@since` that includes a micro version. I mention this because of the other comment about there being no additional API; perhaps/probably a default method would additional API.

Agree on that, lets see if the API tools are now fine with it.

HannesWell:
The build is now all-green.

For me this is ready.
@merks, @laeubi any final words?

iloveeclipse:
See https://github.com/eclipse-equinox/p2/issues/320.

PR #316, opened 2023-09-07
Replace obsolete product unit by its contained feature-units

HannesWell:
Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1235

Additionally update to new Orbit Repo and set targetJRE to match the requirement of P2 for Java-17.

I didn't test it full to build p2 against it, but it at least resolves in my workspace.


akurtakov:
Thanks!

Issue #320, opened 2023-09-15
Comparator error with CachingArtifactRepository in build I20230915-0520


iloveeclipse:
```
Comparator differences from current build
	/home/jenkins/agent/workspace/Builds/I-build-4.30/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/siteDir/eclipse/downloads/drops4/I20230915-0520
compared to reference repo at 
	https://download.eclipse.org/eclipse/updates/4.30-I-builds

1.  rt.equinox.p2/bundles/org.eclipse.equinox.p2.directorywatcher/.polyglot.META-INF
   no-classifier: different
      org/eclipse/equinox/internal/provisional/p2/directorywatcher/CachingArtifactRepository.class: different
    The main artifact has been replaced with the baseline version.
    The following attached artifacts have been replaced with the baseline version: [sources]
```

Diff (see https://download.eclipse.org/eclipse/downloads/drops4/I20230915-0520/buildlogs.php): 
```
  // access flags 0x1041
  public synthetic bridge contains(Ljava/lang/Object;)Z
    ALOAD 0
    ALOAD 1
    CHECKCAST org/eclipse/equinox/p2/metadata/IArtifactKey
    INVOKEVIRTUAL org/eclipse/equinox/internal/provisional/p2/directorywatcher/CachingArtifactRepository.contains (Lorg/eclipse/equinox/p2/metadata/IArtifactKey;)Z
    IRETURN
    MAXSTACK = 2
    MAXLOCALS = 2
```

Side effect from c43af59a8ea7bae05f2d5a59cdef820938849c4f / https://github.com/eclipse-equinox/p2/pull/314

org.eclipse.equinox.p2.directorywatcher need to be touched.

I will push a PR

PR #321, opened 2023-09-15
Comparator error with CachingArtifactRepository in build I20230915-0520

iloveeclipse:
Side effect from c43af59 / https://github.com/eclipse-equinox/p2/pull/314

Fixes https://github.com/eclipse-equinox/p2/issues/320


akurtakov:
I'll fire new I-build so we get some build pushed.

HannesWell:
Thanks Andrey for the fix.

I wonder why this didn't show up in the verification build?

iloveeclipse:
> I wonder why this didn't show up in the verification build?

I believe such errors are reported only on official build, must be one of maven / jenkins build arguments

HannesWell:
Ok, in m2e I've seen such errors
Probably we have it enabled there and I assumed it is also done in Platform verification builds.

PR #328, opened 2023-09-20
Swithch default jre version in JREAction to 17

akurtakov:
Eclipse doesn't support running on older version for quite some time.


bundles/org.eclipse.equinox.p2.publisher/src/org/eclipse/equinox/p2/publisher/actions/JREAction.java
```diff
@@ -16,32 +16,56 @@
...
-import org.osgi.framework.*;
+import org.osgi.framework.BundleException;
+import org.osgi.framework.Constants;
+import org.osgi.framework.FrameworkUtil;
 
 public class JREAction extends AbstractPublisherAction {
 	private static final String DEFAULT_JRE_NAME = "a.jre"; //$NON-NLS-1$
-	private static final Version DEFAULT_JRE_VERSION = Version.parseVersion("11.0"); //$NON-NLS-1$
+	private static final Version DEFAULT_JRE_VERSION = Version.parseVersion("17.0"); //$NON-NLS-1$
 	private static final String DEFAULT_PROFILE = "JavaSE-11"; //$NON-NLS-1$
```
iloveeclipse, line 69: Should "JavaSE-11" be changed too?

akurtakov: Good catch.

PR #334, opened 2023-09-25
Make CompoundQueryable(Collection) constructor public

HannesWell:


bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/p2/query/CompoundQueryable.java
```diff
@@ -52,21 +53,23 @@ public Iterator<T> getCandidates(IEvaluationContext ctx, IExpression variable, I
...
 	 * @param queryables The collection of queryables to be combined
+	 * @since 2.8
 	 */
-	CompoundQueryable(Collection<? extends IQueryable<T>> queryables) {
+	@SuppressWarnings("unchecked")
+	public CompoundQueryable(Collection<? extends IQueryable<T>> queryables) {
 		// don't suppress the warning as it will cause warnings in the official build
 		// see bug 423628. Write this without unchecked conversion.
-		this(queryables.toArray(new IQueryable[queryables.size()]));
+		this.queryables = (Collection<IQueryable<T>>) queryables;
```
laeubi, line 72: why not make the filed of the same type? Also the previous code has made a copy of the collection, so List.copyOf( ... ) would also solve your issue here without a cast.

HannesWell: Good suggestions. Applied them both.
A copy is not necessary in all cases, but often a `List.of()` is passed any way, which is only passed through `List.copyOf()`.

HannesWell:
I added a second commit to add a similar `QueryableArray(Collection)` constructor and use it.
Tycho also has a copy of this class named QueryableCollection that is modified only to have such constructor AFAICT.

Issue #335, opened 2023-09-26
Error log message: beginTask should only be called once per instance. At least call done() before further invocations on install


ajohnson1:
This was seen on a development version of Eclipse Memory Analyzer built against an Eclipse 4.30-I-build.

eclipse.buildId=unknown
java.fullversion=17.0.6+10
JRE 17 Windows 11 amd64-64-Bit Compressed References 20230117_367 (JIT enabled, AOT enabled)
OpenJ9   - e68fb241f
OMR      - f491bbf6f
JCL      - 927b34f84c8 based on jdk-17.0.6+10
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_GB
Command-line arguments:  -os win32 -ws win32 -arch x86_64

org.eclipse.jface
Warning
Mon Sep 25 18:25:19 BST 2023
beginTask should only be called once per instance. At least call done() before further invocations

java.lang.IllegalStateException: beginTask should only be called once per instance. At least call done() before further invocations
	at org.eclipse.jface.operation.AccumulatingProgressMonitor.beginTask(AccumulatingProgressMonitor.java:148)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:483)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:449)
	at org.eclipse.equinox.internal.p2.director.SimplePlanner.getProvisioningPlan(SimplePlanner.java:516)
	at org.eclipse.equinox.internal.p2.ui.ProvUI.toCompabilityWithCurrentJREProvisioningPlan(ProvUI.java:329)
	at org.eclipse.equinox.internal.p2.ui.dialogs.ProvisioningOperationWizard.lambda$10(ProvisioningOperationWizard.java:346)
	at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(ModalContext.java:122)
Caused by: java.lang.IllegalStateException: beginTask(, 1000) was called here previously
	at org.eclipse.jface.operation.AccumulatingProgressMonitor.beginTask(AccumulatingProgressMonitor.java:155)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:483)
	at org.eclipse.core.runtime.SubMonitor.convert(SubMonitor.java:449)
	at org.eclipse.equinox.internal.p2.operations.PlannerResolutionJob.runModal(PlannerResolutionJob.java:75)
[...]

PR #336, opened 2023-09-26
ProfileChangeOperation: fix progressmonitor usage #335

jukzi:
passing a monitor to two callees requires it to be splitted.

https://github.com/eclipse-equinox/p2/issues/335

Also fixed forgotten monitor.done() in performProvisioningPlan and improved tests.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
@@ -46,11 +46,11 @@ public class RawMirrorRequest extends MirrorRequest {
 
 	@Override
 	public void perform(IArtifactRepository sourceRepository, IProgressMonitor monitor) {
-		monitor.subTask(NLS.bind(Messages.downloading, getArtifactKey().getId()));
+		SubMonitor subMon = SubMonitor.convert(monitor, NLS.bind(Messages.downloading, getArtifactKey().getId()), 1);
 		setSourceRepository(sourceRepository);
 		// Do we already have the descriptor in the target?
 		if (target.contains(targetDescriptor)) {
-			setResult(new Status(IStatus.INFO, Activator.ID, NLS.bind(Messages.mirror_alreadyExists, targetDescriptor, target)));
+			setResult(Status.info(NLS.bind(Messages.mirror_alreadyExists, targetDescriptor, target)));
 			return;
 		}
 		// Does the source actually have the descriptor?
@@ -58,7 +58,7 @@ public class RawMirrorRequest extends MirrorRequest {
 			setResult(new Status(IStatus.ERROR, Activator.ID, NLS.bind(Messages.artifact_not_found, getArtifactKey())));
 			return;
 		}
-		IStatus status = transfer(targetDescriptor, sourceDescriptor, monitor);
+		IStatus status = transfer(targetDescriptor, sourceDescriptor, subMon.newChild(1));
 
 		// if ok, cancelled or transfer has already been done with the canonical form return with status set
 		if (status.getSeverity() == IStatus.CANCEL) {
@@ -88,6 +88,7 @@ public class RawMirrorRequest extends MirrorRequest {
 	// Perform the mirror operation without any processing steps
 	@Override
 	protected IStatus getArtifact(IArtifactDescriptor artifactDescriptor, OutputStream destination, IProgressMonitor monitor) {
+		SubMonitor subMon = SubMonitor.convert(monitor, 2);
 		if (SimpleArtifactRepository.CHECKSUMS_ENABLED) {
 			Collection<ChecksumVerifier> steps = ChecksumUtilities.getChecksumVerifiers(artifactDescriptor,
 					IArtifactDescriptor.DOWNLOAD_CHECKSUM, Collections.emptySet());
@@ -98,9 +99,9 @@ public class RawMirrorRequest extends MirrorRequest {
 			ProcessingStep[] stepArray = steps.toArray(new ProcessingStep[steps.size()]);
 			// TODO should probably be using createAndLink here
 			ProcessingStepHandler handler = new ProcessingStepHandler();
-			destination = handler.link(stepArray, destination, monitor);
+			destination = handler.link(stepArray, destination, subMon.split(1));
 		}
-
-		return getSourceRepository().getRawArtifact(artifactDescriptor, destination, monitor);
+		subMon.setWorkRemaining(1);
+		return getSourceRepository().getRawArtifact(artifactDescriptor, destination, subMon.split(1));
 	}
 }
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/SimpleArtifactRepository.java
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/SimpleArtifactRepository.java
@@ -617,7 +617,7 @@ public class SimpleArtifactRepository extends AbstractArtifactRepository impleme
 	}
 
 	protected IStatus downloadArtifact(IArtifactDescriptor descriptor, OutputStream destination, IProgressMonitor monitor) {
-		monitor = IProgressMonitor.nullSafe(monitor);
+		SubMonitor subMon = SubMonitor.convert(monitor, 2);
 		if (isFolderBased(descriptor)) {
 			File artifactFolder = getArtifactFile(descriptor);
 			if (artifactFolder == null) {
@@ -655,8 +655,8 @@ public class SimpleArtifactRepository extends AbstractArtifactRepository impleme
 		URI baseLocation = getLocation(descriptor);
 		if (baseLocation == null)
 			return new Status(IStatus.ERROR, Activator.ID, NLS.bind(Messages.no_location, descriptor));
-		URI mirrorLocation = getMirror(baseLocation, monitor);
-		IStatus status = downloadArtifact(mirrorLocation, destination, monitor);
+		URI mirrorLocation = getMirror(baseLocation, subMon.split(1));
+		IStatus status = downloadArtifact(mirrorLocation, destination, subMon.split(1));
 		IStatus result = reportStatus(descriptor, destination, status);
 		// if the original download went reasonably but the reportStatus found some issues
 		// (e..g, in the processing steps/validators) then mark the mirror as bad and return
--- a/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/Engine.java
+++ b/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/Engine.java
@@ -71,6 +71,7 @@ public class Engine implements IEngine {
 		Profile profile = profileRegistry.validate(iprofile);
 
 		profileRegistry.lockProfile(profile);
+		SubMonitor subMon = SubMonitor.convert(monitor, 3);
 		try {
 			eventBus.publishEvent(new BeginOperationEvent(profile, phaseSet, operands, this));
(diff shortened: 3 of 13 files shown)


bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
```diff
@@ -46,39 +46,43 @@ public RawMirrorRequest(IArtifactDescriptor sourceDescriptor, IArtifactDescripto
...
-			setResult(new Status(IStatus.INFO, Activator.ID, NLS.bind(Messages.mirror_alreadyExists, targetDescriptor, target)));
-			return;
-		}
-		// Does the source actually have the descriptor?
-		if (!source.contains(getArtifactDescriptor())) {
-			setResult(new Status(IStatus.ERROR, Activator.ID, NLS.bind(Messages.artifact_not_found, getArtifactKey())));
-			return;
-		}
-		IStatus status = transfer(targetDescriptor, sourceDescriptor, monitor);
+		SubMonitor subMon = SubMonitor.convert(monitor,NLS.bind(Messages.downloading, getArtifactKey().getId()),1);
```
HannesWell, line 49: Converting the monitor into a `SubMonitor` with work one does not look necessary here and it can just be passed directly.
Furthermore I think it is not necessary to add the try-catch-block to ensure done is called in any ways.
Just calling done below is sufficient and if the method throws the method returns any way and a caller that catches an exception either calls done on the passed monitor or (should) call split which also cleans up children.

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
```diff
@@ -88,19 +92,25 @@ public IArtifactDescriptor getArtifactDescriptor() {
 	// Perform the mirror operation without any processing steps
 	@Override
 	protected IStatus getArtifact(IArtifactDescriptor artifactDescriptor, OutputStream destination, IProgressMonitor monitor) {
-		if (SimpleArtifactRepository.CHECKSUMS_ENABLED) {
-			Collection<ChecksumVerifier> steps = ChecksumUtilities.getChecksumVerifiers(artifactDescriptor,
-					IArtifactDescriptor.DOWNLOAD_CHECKSUM, Collections.emptySet());
-			if (steps.isEmpty()) {
-				LogHelper.log(new Status(IStatus.WARNING, Activator.ID,
-						NLS.bind(Messages.noDigestAlgorithmToVerifyDownload, artifactDescriptor.getArtifactKey())));
+		monitor = IProgressMonitor.nullSafe(monitor);
```
HannesWell, line 95: `SubMonitor.convert` can handle null monitors and can't we just call done on the sub-monitor?
Again I think the try-catch is not necessary.

HannesWell (commented):
Thanks Jörg for this PR.
In general in most places `SubMonitor.convert()` can be used directly and calling `IProgressMonitor.nullSafe()` seems not necessary.

Furthermore as already written in the detailed comments, a try-catch-block is often not necessary because in case of an exception the method terminates immediatly and progress-reporting usually stops anyways.

Besides the already covered locations it looks like `MirrorRequest.transfer()` also seem not to handle its monitor correctly, just like `SimpleArtifactRepository.getArtifact()` and `getRawArtifact()`.

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
laeubi: > Furthermore I think it is not necessary to add the try-catch-block to ensure done is called in any ways

Correct, IProgressmonitor states:

CALLER/CALLEE RESPONSIBILITIES:

Methods that receive an [IProgressMonitor](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82IProgressMonitor) ("callees") must either obey the following conventions or include JavaDoc explaining how it differs from these rules. The called method:

Will call [beginTask](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82%E2%98%82beginTask) on the argument 0 or 1 times, at its option.
Will not promise to invoke [done()](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82%E2%98%82done%E2%98%82) on the monitor.
Will not call [setCanceled](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82%E2%98%82setCanceled) on the monitor.
May rely on the monitor not being null.
May rely on the monitor ignoring the string passed to [beginTask](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82%E2%98%82beginTask).

laeubi: > Furthermore I think it is not necessary to add the try-catch-block to ensure done is called in any ways

Correct, IProgressmonitor states:

CALLER/CALLEE RESPONSIBILITIES:

Methods that receive an [IProgressMonitor](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82IProgressMonitor) ("callees") must either obey the following conventions or include JavaDoc explaining how it differs from these rules. The called method:

Will call [beginTask](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82%E2%98%82beginTask) on the argument 0 or 1 times, at its option.
Will not promise to invoke [done()](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82%E2%98%82done%E2%98%82) on the monitor.
Will not call [setCanceled](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82%E2%98%82setCanceled) on the monitor.
May rely on the monitor not being null.
May rely on the monitor ignoring the string passed to [beginTask](eclipse-javadoc:%E2%98%82=org.eclipse.equinox.common/src%3Corg.eclipse.core.runtime%7BIProgressMonitor.java%E2%98%83IProgressMonitor%E2%98%82%E2%98%82beginTask).

laeubi: > furthermore I think it is not necessary to add the try-catch-block to ensure done is called in any ways.

Yep, see javadoc of IProgressMonitor ...

jukzi: By passing a submonitor with progress one and unconditional calling done() the progress is unconditional increased which ensures to use up all the reserved progress. That is technically not necessary but ensures a user readable progress from 0 to 100% . This commit fixes the error and proofs that with junit test. This pattern is easy to understand and not error prone. I am not interested in further "optimizations" that "may" work under special conditions.

laeubi: I just wanted to note that when one splits a monitor previous aquired "splits" will be completed automatically, so if the toplevel is calling `done()` and the other code is always using `split()` progress si still reported correctly in every case.

According to `IProgressMonitor` javadoc the called method: 

1. Will call beginTask on the argument 0 or 1 times, at its option.
2. Will not promise to invoke done() on the monitor.
3. Will not call setCanceled on the monitor.
4. May rely on the monitor not being null.
5. May rely on the monitor ignoring the string passed to beginTask.

so calling done is of course _allowed_, but not _required_ by a method receiving a monitor.

jukzi:
Mind to submit as-is if nobody works on further improvements?

HannesWell:
> Mind to submit as-is if nobody works on further improvements?

I can take care of this later/in the next days.

HannesWell:
@jukzi as discussed I now corrected/simplified the usage of the SubMonitor.
Please have a look if your main points are still addressed.

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
jukzi: all my concerns are expressed by the junit result.

HannesWell: But asserting in AbstractProvisioningUITest that for example `ProvisioningSession.performProvisioningPlan()` consumes all work of the monitor is not right.

From the Java-doc of IProgressMonitor:
https://github.com/eclipse-equinox/equinox/blob/8beb9ce0b1dea1176b06ad7300f2663ff6db4eff/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java#L57

```
CALLER/CALLEE RESPONSIBILITIES: 

Methods that receive an {@link IProgressMonitor} ("callees") must either obey the following conventions or include  these rules. The called method:
[...]
- Will not promise to invoke {@link #done()} on the monitor.

[...]
The caller:
[...]
Will not rely on the callee to invoke {@link #done()} on the monitor. It must either select an implementation of {@link IProgressMonitor} that does not require {@link #done()} to be called, for example {@link SubMonitor}, or  it must invoke {@link #done()} itself after the method returns.
```

And the only real caller of `ProvisioningSession.performProvisioningPlan()` in ProfileModificationJob.runModal() even explicitly calls done, although even that would actually not be necessary since the caller of that method respectively the creator of a progress-monitor is responsible to call done.

So asserting that in the tests is too strict.

jukzi:
> @jukzi as discussed I now corrected/simplified the usage of the SubMonitor.

@HannesWell the tests fail now - which shows the problem

HannesWell:
> > @jukzi as discussed I now corrected/simplified the usage of the SubMonitor.
> 
> @HannesWell the tests fail now - which shows the problem

As said above it is not a problem. A callee does need not call done on the received monitor as the doc states.
It is perfectly fine if a method has e.g. ten pieces of work to do if it returns and the monitor only has 9/10 pieces of work consumed.
It is the responsibility of the creator of of a Monitor to call done on it.
Therefore I think the new assumption in the tests is not right.

Did you encounter any creator of a Monitor that did not call done? If yes that should be fixed instead of requiring the calles to call done. Of course they should consume some amounts of work, but the last bit can be left out. Therefore the SubMonitor.split/newChild() pattern was introduced, which actually makes it relatively simple to use monitors right.

laeubi:
> Of course they should consume some amounts of work

A receiver of the monitor might or might not call any method on the monitor... The good thing about submonitor is that one can reuse unused progress if desired.

HannesWell:
I know changed the two methods that the test complains about (although I still have the opinion that they are too strict) to call done at the end of the method, but not in a finally block. This way all the callers that make false assumptions should be happy too.

This way we can have at least the more important changes to avoid multiple monitor usages.

PR #338, opened 2023-09-28
Add IMetadataRepository.removeReferences()

HannesWell:
In order to allow removing references to other repositories from an `IMetadataRepository`, just like one can also remove IUs, this adds an `IMetadataRepository.removeReferences(Collection)` method and implements it in all classes that also implement `IMetadataRepository.addReferences()`.

I have not yet added them, but a little test-case, at least for the most imporant implementation `LocalMetadataRepository` would probably be good.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.metadata.repository/src/org/eclipse/equinox/internal/p2/metadata/repository/CompositeMetadataRepository.java
+++ b/bundles/org.eclipse.equinox.p2.metadata.repository/src/org/eclipse/equinox/internal/p2/metadata/repository/CompositeMetadataRepository.java
@@ -246,11 +246,7 @@ public class CompositeMetadataRepository extends AbstractMetadataRepository impl
 			//			return new File(spec + extension);
 			return spec;
 		}
-		if (path.endsWith("/")) //$NON-NLS-1$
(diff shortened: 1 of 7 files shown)


HannesWell:
> I have not yet added them, but a little test-case, at least for the most imporant implementation `LocalMetadataRepository` would probably be good.

After thinking about that again, a test case would be quite trivial and just test that a collection would not contain elements anymore after they have been removed. Such trivial tests are usually not necessary.

Therefore I consider this ready. Any other opinions?

merks:
I personally don't think trivial test cases are needed.

laeubi:
> a test case would be quite trivial

If its trivial it should be trivial to add it, no? :-D

HannesWell:
> > a test case would be quite trivial
> 
> If its trivial it should be trivial to add it, no? :-D

It is. :) But testing trivial things just to increase the test-coverage also does not bring much value.
Anyway I now added tests for `addReferences()` and `removeReferences()` of `LocalMetadataRepository`.

Furthermore I made the `LocalMetadataRepository.repositories` a `LinkedHashSet` to ensure a stable order of repository references, which simplifies manual inspection of the generated `content.xml` files (technically there is no benefit).

Issue #207, opened 2022-12-25
Popup for "Error Contacting Site" should offer to delete the update site


erik-brangs:
When searching for new items to install and when an update site cannot be contacted, there is a popup "Error contacting site" that says "No software site found at $URL. Do you wish to edit that location?".

When this occurs during an installation that contacts all update sites, this popup is not very helpful. It's difficult to remember the affected site, particularly if this occurs for multiple sites. Additionally, this may affect older update sites that aren't used any longer. It would be useful to give the user the option to simply delete outdated sites so that they don't cause problems in the future. Another useful possiblity would be to allow the user to disable the location. That would make the outdated site easier to find in the software sites list later.

mickaelistria:
That's a good proposal. Would you be interested in trying to submit a PR implementing such enhancement?

erik-brangs:
Yes, in theory. However, I'm not sure when I will get around to do it. So it's fine for me if someone else wants to do it instead.

ajohnson1:
A related idea is whether a 'HTTP/1.1 301 Moved Permanently' response should update the location - automatically or with a 'Yes' 'No' and perhaps one of 'Never' (meaning for the update site source location) or 'Never' (meaning for the update site source location and redirection location pair).
https://www.rfc-editor.org/rfc/rfc2616#section-10.3.2
>The requested resource has been assigned a new permanent URI and any
   future references to this resource SHOULD use one of the returned
   URIs.  Clients with link editing capabilities ought to automatically
   re-link references to the Request-URI to one or more of the new
   references returned by the server, where possible. This response is
   cacheable unless indicated otherwise.
 
https://www.rfc-editor.org/rfc/rfc7231#section-6.4.2

> The 301 (Moved Permanently) status code indicates that the target
   resource has been assigned a new permanent URI and any future
   references to this resource ought to use one of the enclosed URIs.
   Clients with link-editing capabilities ought to automatically re-link
   references to the effective request URI to one or more of the new
   references sent by the server, where possible.
   
   appear to suggest that it should happen automatically.
   
Also, a 410 Gone should probably delete the location entry (and leave the name) or disable the update site (or remove the whole update site?). Is it useful for a user to see that that there used to be a software site but it isn't now usable.
https://www.rfc-editor.org/rfc/rfc7231#section-6.5.9

>The 410 response is primarily intended to assist the task of web
   maintenance by notifying the recipient that the resource is
   intentionally unavailable and that the server owners desire that
[...]

PR #343, opened 2023-10-02
Issue 207 - Add additional buttons für deleting and removing update sites to the dialog "Error Contacting Site"

erik-brangs:
Here's an attempt to fix the issues discussed in the first few posts of #207 , i.e. additional buttons for disabling and removing update sites. This PR does not implement any link rewriting as discussed later in #207.

I've made an attempt to write some tests but I don't know if that's the correct way to go about it. It's also not clear to me if events are needed for the operations. ColocatedRepositoryTracker also doesn't seem to use them. I've added TODOs for that in the PR.


laeubi:
@akurtakov @merks the build seems green so far can we integrate this?

merks:
I see 4 failed builds. I don’t think they failed because of your changes though.

laeubi:
Arggg... I missed the missing version bumps... @erik-brangs can you please increase the versions now?

laeubi:
@erik-brangs thanks Jenkins succeeds now and I approved GH actions, can you probably squash your commits into one?

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/LocationNotFoundDialog.java
```diff
@@ -0,0 +1,102 @@
+package org.eclipse.equinox.internal.p2.ui;
```
laeubi, line 1: Please add a copyright header

bundles/org.eclipse.equinox.p2.tests.ui/src/org/eclipse/equinox/p2/tests/ui/operations/LocationNotFoundDialogTest.java
```diff
@@ -0,0 +1,80 @@
+// TODO is this the correct package? I used the same one as ColocatedRepositoryTracker
```
laeubi, line 1: Please add copyright header

laeubi:
@merks  for me this looks sane now, a test is added that also succeeds so maybe we should give this a try?

merks:
Trying it out locally, the first thing I tried was to add a bogus site and do check for updates, which fails like this:

image

I realize that this PR does not try to address that case.  

Just entering/choosing a bogus site fails like this:

image

which is also fine I suppose.

If I enter a good site, but then ask for looking in All Available Sites, it fails like this, though I can close the dialog and ignore that failure:

image

When I install something  like this, I see exceptions in the log, but nothing in the UI about that:

image

and the install succeeds:

image

I actually have not been able to figure out when the dialog kicks in that I might test this manually.  Perhaps someone can enlighten me how I can test this manually...

erik-brangs:
You should see the dialog if you do the following:
- Add an update site for ``http://download.eclipse.org/technology/m2e/releases/`` via ``Available Software Sites``
- Use ``Install New Software`` with ``All Available Sites``

I suppose there can be caching involved. I'm not sure what to delete to get a clean state.

laeubi:
@erik-brangs can you verify the mentioned places by @merks so the dialog is showing there as well?

merks:
Note that I don't want to block anything here.  I would just be nice to see it in action...

laeubi:
@erik-brangs do you plan to further work on this? Stream is open so you need to resolve the manifest conflict and rebase but this would be a good time for such changes.

erik-brangs:
I will continue working on this pull request this month.

merks:
It seems to me one reason it doesn't show is because the failures might be accumulated instead of reported:

image

erik-brangs:
Thank you for the hints. The dialog indeed only seems to be shown when accumulation of failures is disabled. Unfortunately, the new `LocationNotFoundDialog` is not reusable for the accumulation case because it assumes a `ColocatedRepositoryTracker`. The `LoadMetadataRepositoryJob` only provides a `RepositoryTracker`.

I don't see a way to handle the accumulated case without doing larger changes. Calling `reportLoadFailure` is not an option for the accumulated case because `reportLoadFailure` checks the `reposNotFound` in `RepositoryTracker` which have already been filled during the accumulation. That is, `reportLoadFailure` ignores repositories reported via `addNotFound(URI)` because the JavaDoc of `addNotFound(URI)` says that callers are assumed to have reported failures already. Handling the accumulated case would therefore mean changing the reporting from using `IStatus` to something else. I assume this would need new UI.

I have adjusted the original changes for the non-accumulated cases a bit. Is the PR suitable for merging in this state?

PS: I don't know why the build failed.

laeubi:
@erik-brangs I don't completely understand why the accumulation is not usable here, is it because you miss the exception? Can you share an screenshot of the dialog to illustrate the problem maybe?

erik-brangs:
The accumulation case implies - per the JavaDoc of the methods that its implementation uses - that the failures are already reported. You can therefore not use the reporting functionality from `ColocatedRepositoryTracker` or any other `RepositoryTracker`.

When the accumulation is done, the failure has not actually been reported to the user yet - but from the point of `RepositoryTracker` it has been reported because the failing locations were added to `reposNotFound`. Any call to `reportLoadFailure` on a `RepositoryTracker` will do nothing for repos that are contained in `reposNotFound`.

So you have to do the reporting for the accumulated case in `LoadMetadataRepositoryJob`. This is currently done via the `IStatus` as shown by Ed's screenshot from the accumulation.

The `LocationNotFoundDialog` uses methods like `getMetadataRepositoryManager` which are only available on `ColocatedRepositoryTracker`. The dialog could be rewritten to use `ProvUI.getMetadataRepositoryManager(..)` and `ProvUI.getArtifactRepositoryManager(..)`.

However, the dialog only handles the failure for one site: 
LocationNotFoundDialog

During the accumlation case, multiple failures could occur. How should those be handled? I don't think it would be a good idea to display the dialog for each site. So the accumulation case needs new UI.

laeubi:
Thanks for the explanation and screenshot, that makes it now a bit clearer.

> During the accumlation case, multiple failures could occur. How should those be handled? I don't think it would be a good idea to display the dialog for each site. So the accumulation case needs new UI.

I think the best would be to check at the place where the error dialog is shown if the accumulated failures are *only* caused by updatesites not found (is this possible?).

If that is the case, a dialog similar to yours can be shown that ask if they should be disabled or removed altogether would be enough, if it is just one single failed one then showing the existing dialog.

If it makes things easier we also can always assume they are colocated I'm not aware of any case where a user can enter individual items.

erik-brangs:
I've tried to adjust the PR according to the feedback. I'm not sure why the ECA check fails. According to the Eclipse Foundation website, my ECA is current.

laeubi:
@erik-brangs if you click on the failed check you should get additional information that (hopefully) help you to see whats wrong.

erik-brangs:
I already did that and the error message made no sense to me. It said there was no ECA for my account which isn't true.

The ECA verification tool on https://api.eclipse.org/git/eca/status/gh/eclipse-equinox/p2/343 says there's no Eclipse Foundation account for my email which can't be true because it just logged in with my account at the Eclipse Foundation website.

laeubi:
Did you try the troubleshooting?



> A common error experienced by users is that the email associated with commits do not align with the email set in the Eclipse Foundation account. If you have an Eclipse Foundation account and a signed ECA-equivalent document, there are ways to validate the users and email addresses associated with a commit.
> 
> Below are 3 sample commands that will display the committer and author emails for a given commit or range. The first command shows the latest commit, while the second command will show results for a given commit, replacing `<SHA>` with the commit SHA. The last example will render a range of sequential commits. As is, it will show the last 5 commits starting from HEAD for the current branch. Both sides of the range can be replaced with commit SHAs to represent a different range.
> ```
> git show -s --format='Commit: %h, Committer: %ce, Author: %ae
> git show -s --format='Commit: %h, Committer: %ce, Author: %ae' <SHA>
> git show -s --format='Commit: %h, Committer: %ce, Author: %ae' HEAD...HEAD~5 
> ```

Also make sure that your github user is set in the account details.

erik-brangs:
It seems the problem with the ECA check is fixed. Revalidation of the ECA succeeded.

mickaelistria:
I find 4 buttons in a row is very heavy. Have you considered just adding a link to the "Install/Updates > Available Sites" preference page?

erik-brangs:
In my installations, the "Available Sites" preference page normally has more sites than can fit on one page. In that case, I'd need use the filtering on the preference page. That means I have to remember which site failed. As I wrote in the original issue, one reason for the change was to relieve the user from remembering the failing site.

laeubi:
These dialogs should not show up very often and the intention is for a user to quickly fix the issue "in place" without having to browse lists and perform additional steps so I think it is fine.

If one wants the Dialog can be improved later on, as currently there is effectively no way at all as to dismiss the dialog completely.

@merks do you think your concerns are addressed and this can be merged? Anything left now?

merks (approved):
If the host is bogus, it doesn't do much:

image

But if one specifies something bogus on an actual host, I finally see this dialog:

image

So that seems okay.

I have **not reviewed the code in detail**, I assume @laeubi  will do that.

We should probably also rebase to ensure that we're not missing version increments.

laeubi:
> But if one specifies something bogus on an actual host, I finally see this dialog:

I would say: something that do not exits, but yes we can assume that at least at one point in time the URL was valid and the udatesite just vanished (or most likely was http in the past).

bundles/org.eclipse.equinox.p2.operations/src/org/eclipse/equinox/p2/operations/RepositoryTracker.java
laeubi, line 1: It seems some reformating not related to the change have slipped int

laeubi (approved):
Code looks good.

bundles/org.eclipse.equinox.p2.operations/src/org/eclipse/equinox/p2/operations/RepositoryTracker.java
erik-brangs: That's just trimming of unnecessary whitespace, probably via save actions. I can remove it later if it's that important.

PR #347, opened 2023-10-08
Modernize internal p2 metadata Member classes

HannesWell:
This also improves the `hashCode()` function of Requirement.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/ResolvedInstallableUnit.java
+++ b/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/ResolvedInstallableUnit.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2007, 2018 IBM Corporation and others.
+ *  Copyright (c) 2007, 2023 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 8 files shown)


bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/ResolvedInstallableUnit.java
```diff
@@ -240,37 +206,24 @@ public boolean satisfies(IRequirement candidate) {
...
-			return getTouchpointData();
-		if (InstallableUnit.MEMBER_TOUCHPOINT_TYPE == memberName)
-			return getTouchpointType();
-		if (InstallableUnit.MEMBER_UPDATE_DESCRIPTOR == memberName)
-			return getUpdateDescriptor();
-		if (InstallableUnit.MEMBER_SINGLETON == memberName)
-			return Boolean.valueOf(isSingleton());
-		throw new IllegalArgumentException("No such member: " + memberName); //$NON-NLS-1$
+		return switch (memberName) {
+		case MEMBER_FRAGMENTS -> fragments;
```
HannesWell, line 210: Can anybody tell if it is an issue if this returns a List instead of an array?

HannesWell: @merks or @mickaelistria do you have an idea?
I followed the callers a bit and got the impression that arrays are not considered as expressions at all, so could this be a hidden fix for a problem?
At least the `ResolvedIUInPCR`, `OperationGenerationTest` and `EngineTest` do not cover this code.

merks: Sorry, no I have no clue about this.

mickaelistria: I don't know much about how this class is used, so I don't have solid arguments here.
But my first impression is that it could be better to stick to arrays to:
* prevent regressions in potentially unknown consumers that are casting to array
* allow some cast the returned Object and get some stronger typing checked (ie `o instanceof String[]` works while `o instanceof List<String>` doesn't).

HannesWell: Agree.
Changed the code to `fragments.toArray(IInstallableUnitFragment[]::new)`. Since this is probably not called often that's not a concern.

PR #348, opened 2023-10-08
Add IInstallableUnitQueryable interface to allow easier adaption

laeubi:
Currently there is only the generic IQueryable<T> but this don't work well in cases where one can't pass type information for the generic e.g. when adapting an object there is no way to pass the parameter T.

This adds a new interface that lock in the type T for a IQueryable that can supply IInstallableUnits.

This is a prerequisite of https://github.com/eclipse-pde/eclipse.pde/issues/207 where I'd like to use the Adapter pattern to support other locations to supply additional Installable units.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/p2/metadata/IInstallableUnitQueryable.java
+++ b/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/p2/metadata/IInstallableUnitQueryable.java
@@ -1,0 +1,28 @@
+/*******************************************************************************
+ *  Copyright (c) 2023 Christoph Läubrich and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


laeubi:
@merks @mickaelistria @HannesWell any objections merging this (as it is actually new API we should have one additional comiter approval). I think there should neither compatibility nor maintain concerns of having this simple interface and this currently prevents the PDE verification build from happening.

laeubi:
An alternative would be to make this PDE "private" and not general available to P2...

merks:
It seems okay, though I'm not entirely sure why something needed by PDE has to be added to p2...

laeubi:
> It seems okay, though I'm not entirely sure why something needed by PDE has to be added to p2...

Main purpose is to `IMetadataRepository` can implement this interface as well (what would be to heavy for general purpose) and its easier to reuse in other context, but as said it could also be done directly in PDE it just seems better placed at P2 (as its interface only).

merks:
I'm fine with whatever you and @HannesWell decide is best...

laeubi:
Then lets go so it is available in tomorrows i-build

Issue #345, opened 2023-10-06
Support names for Updates repositories in product definition


HannesWell:
It should be possible to specify a name for an Repository referenced in the Updates configuration section of a Product.

In P2 this requires a consideration of a name attribute in org.eclipse.equinox.internal.p2.publisher.eclipse.ProductFile,processRepositoryInformation() and when the read repositories are processed.
In PDE the `UpdatesSection` and `IRepositoryInfo` need to be enhanced to also support adding a name attribute.

@jcompagner since you raised the issue, are you interested in working on this?

### Discussed in https://github.com/eclipse-equinox/p2/discussions/342

<div type='discussions-op-text'>

<sup>Originally posted by **jcompagner** October  2, 2023</sup>
Hi,

not sure where the product file belongs to that makes the product (is this platform or rcp? or p2?)

So i asked the question here, in the product file (editor) you can define your repositories where this product builds from (or updates from)

Problem is the name is not asked for, and then when you install it, in the preferences it is displayed very weird, because you have a lot without a name. The name seems kind of a bit mandatory (also in the p2.inf where i now place it)

Is there a reason for that? i tries to just adjust the product xml file by adding "name=xxx" or "displayname=xxx" but that isn't picked up.

</div>

HannesWell:
@jcompagner are you interested in working on this? If not I'll take care of this.

jcompagner:
yes i can have a look
But a bit busy, and next week i will be at Eclipse Con in Ludwigsburg ...

ajohnson1:
I did not get the update section of the product definition to work with Eclipse Memory Analyzer.
https://git.eclipse.org/c/mat/org.eclipse.mat.git/tree/org.eclipse.mat.product/mat.product

Instead I had to use a p2.inf file:
https://git.eclipse.org/c/mat/org.eclipse.mat.git/tree/org.eclipse.mat.product/mat-2023-12.p2.inf

This is with a Tycho/Maven build.

jcompagner:
> Great. I'll be there too, so maybe we can have a chat about this topic. :)

we can have a meet up then in the coming few days..  to see each other face to face is always nice.

HannesWell:
@jcompagner can you tell already If/when you can look into the PDE-Editor support for this feature?

jcompagner:
i will have a look hopefully today,

But first i need to look at another PDE issue that suddenly pops up today
our main target file has "com.sun.jna" showing up under the content
[...]

PR #353, opened 2023-10-15
Add support for Product Update-site names

HannesWell:
P2 part of https://github.com/eclipse-equinox/p2/issues/345


jcompagner:
i see this commit already makes sure if a name is found it is set?

So is the current problem purely a ui thing? that the UI should ask for the name?

PR #356, opened 2023-10-20
Do content sensing to make sure the file is really not a jar

laeubi:
Currently we rely on the file-ending to determine the type of file but this can fail in cases a CacheManger returns a cached file without that extension (for whatever reason).

This now adds another check for content sensing if the simple check fails to ensure a zipped item is not parsed as plain text.


merks (approved):
Looks fine.

laeubi:
Great so lets merge this after Eclipse Platform M2 promotion freeze is over...

merks:
There is no freeze for m2. Just merge when ready.

Issue #361, opened 2023-10-25
URISyntaxException


jukzi:
running help/Perform Setup Tasks... in an oomphed platform workspace results in

```
An error was detected while performing the engine operation and the changes are being rolled back. See the log for details.
ERROR: org.eclipse.equinox.p2.engine code=4 An error occurred while unconfiguring the items to uninstall
  at org.eclipse.oomph.util.OomphPlugin.coreException(OomphPlugin.java:296)
  at org.eclipse.oomph.p2.internal.core.ProfileTransactionImpl$3.commit(ProfileTransactionImpl.java:578)
  at org.eclipse.oomph.p2.internal.core.ProfileTransactionImpl.commit(ProfileTransactionImpl.java:359)
  at org.eclipse.oomph.setup.p2.impl.P2TaskImpl.perform(P2TaskImpl.java:904)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.doPerformNeededSetupTasks(SetupTaskPerformer.java:3864)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer$WorkspaceUtil$1.run(SetupTaskPerformer.java:5200)
  at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2453)
  at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2478)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer$WorkspaceUtil.performNeededSetupTasks(SetupTaskPerformer.java:5193)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.performNeededSetupTasks(SetupTaskPerformer.java:3798)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.performTriggeredSetupTasks(SetupTaskPerformer.java:3773)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.perform(SetupTaskPerformer.java:3651)
  at org.eclipse.oomph.setup.ui.wizards.ProgressPage$9.run(ProgressPage.java:592)
  at org.eclipse.oomph.setup.ui.wizards.ProgressPage$11$1.run(ProgressPage.java:721)
  at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
[...]

PR #362, opened 2023-10-26
eclipse.ini: skip comment lines  #361

jukzi:
Commment lines have '#' as its first character

https://github.com/eclipse-equinox/p2/issues/361

### DIFF
--- a/bundles/org.eclipse.equinox.frameworkadmin.equinox/src/org/eclipse/equinox/internal/frameworkadmin/equinox/utils/FileUtils.java
+++ b/bundles/org.eclipse.equinox.frameworkadmin.equinox/src/org/eclipse/equinox/internal/frameworkadmin/equinox/utils/FileUtils.java
@@ -17,7 +17,8 @@ package org.eclipse.equinox.internal.frameworkadmin.equinox.utils;
 import java.io.*;
 import java.net.*;
 import java.util.*;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.URIUtil;
 import org.eclipse.equinox.internal.frameworkadmin.equinox.EquinoxConstants;
 import org.eclipse.equinox.internal.frameworkadmin.equinox.ParserUtils;
 import org.eclipse.equinox.internal.provisional.frameworkadmin.LauncherData;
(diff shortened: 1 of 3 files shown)


jukzi:
@merks mind to review?

merks:
It looks like you're treating `!` as a comment character but the wiki link you showed mentions `;` and not `!`.

I was going to ask, could you please double check what the actual *.ini reader is doing at runtime?  But why ask someone else to do something you can do yourself.  😱  I believe this is the place where the native launcher ignores lines that start with '#'.

https://github.com/eclipse-equinox/equinox/blob/8beb9ce0b1dea1176b06ad7300f2663ff6db4eff/features/org.eclipse.equinox.executable.feature/library/eclipseConfig.c#L146-L150

This logic does not ignore any other from of content.  It also does not ignore lines where there is whitespace before the comment.  So I think it's best that you support only and exactly what will really work at runtime as a commented out line, i.e., a line starting with `#`.

laeubi:
@jukzi du you like to complete this before freeze on Monday?

jukzi:
thanks for the reminder, i changed commentlines to start with '#' only

merks (approved):
This looks exactly right.

bundles/org.eclipse.equinox.frameworkadmin.equinox/src/org/eclipse/equinox/internal/frameworkadmin/equinox/utils/FileUtils.java
```diff
@@ -204,16 +204,20 @@ public static URI fromFileURL(String url) throws URISyntaxException {
...
 	public static List<String> loadFile(File file) throws IOException {
 		try (BufferedReader br = new BufferedReader(new FileReader(file));) {
 			String line;
 			List<String> list = new ArrayList<>();
 			while ((line = br.readLine()) != null) {
-				//skip whitespace
-				if (line.trim().length() > 0)
+				// skip whitespace and comments
+				String stripped = line.strip();
+				if (stripped.length() > 0 && !line.startsWith("#")) {//$NON-NLS-1$
```
HannesWell, line 218: Could be `!stripped.isEmpty()`

HannesWell (approved):
LGTM.
Just one suggestion below.

bundles/org.eclipse.equinox.frameworkadmin.equinox/src/org/eclipse/equinox/internal/frameworkadmin/equinox/utils/FileUtils.java
jukzi: done, please submit

PR #368, opened 2023-11-01
Remove processing of unused plugin element attributes in feature.xml and distinction between Plug-ins and Fragments in ProductFile

HannesWell:
This PR consists of two parts:

1. Remove unused distinction between Plug-ins and Fragments in `ProductFile`
There is no usage of a product's fragments in P2 or PDE.
Consequently PDE support for adding a `fragment` attribute for fragments listed in a product file can be removed too.

2. Remove processing of unused plugin element attributes in `feature.xml`
Which happens in the context of https://github.com/eclipse-pde/eclipse.pde/issues/730
From the removed attributes only the unpack attribute has been considered in in the `FeaturesAction` and the `RemoteFeaturesAction` in order to create corresponding `BundleShapeAdvice`.
But at least for Tycho those advies did not have any effect.

I assume, but I haven't verified it, that `BundleShapeAdvices` might have an effect in PDE-build tasks. But since a bundle itself should specify its bundle-shape in its MANIFEST.MF I'm fine with removing that support from PDE with this.
Nevertheless it requires some rework of the PDE-builds task since they use `FeatureEntry.isUnpack()/unpackSet()/...`, therefore this is currently a draft.


laeubi:
@HannesWell is this still relevant? It requires a manual rebase now!

laeubi:
@HannesWell would be good to either finish or close this...

laeubi:
Closing as no response and getting even more conflicts now, feel free to reopen / redo whenever it becomes important again.

Issue #365, opened 2023-10-30
Improve the DirectorApplication to provide better security control


merks:
The DirectorApplication is designed to install all content using DirectorApplication.AvoidTrustPromptService.

As described in 

https://gitlab.eclipse.org/security/vulnerability-reports/-/issues/159

users may wish to have more control over the content being installed.  For example, ideally it would be possible to reject unsigned content, and even to reject unexpected PGP keys, unrooted certificates, and/or content from untrusted authorities.

PR #370, opened 2023-11-02
Improve the DirectorApplication to provide better security control

merks:
Support -verboseTrust for logging trust details.
Support -trustSignedContentOnly for rejecting unsigned content. Support -trustedAuthorities to list authorities to trust for content. Support -trustedPGPKeys to list PGP keys to trust for signing. Support -trustedCertificates to list SHA-256 fingerprints of unanchored certificates to trust for signing.

All the options are "off" by default to ensure that there are no behavioral changes.

Each list option supports an empty list to outright reject all such content.

https://github.com/eclipse-equinox/p2/issues/365


merks:
@laeubi 

Do you have any concerns about this?

I personally doubt many people will want to use this, but given that this is likely to be (will be) reported as a CVE, it seems better to take preventative measures.  I haven't changed the default behavior so no one needs to take action...

laeubi:
Looks good.

Issue #374, opened 2023-11-04
It's not possible to verify a PGP signature of a folder


merks:
When p2 is asked to transfer an artifact from one repository to another and both repositories are in the file system, it's possible that some of such artifacts to be transferred are folders and then p2 will just copy the folder (recursively) from one place in the file system to another place in the file system.  It will also try to do signature verification, which works fine for jar-signed content because p2 creates a jar and then verifies the jar-signature of that synthesized jar.   But for PGP,  this just doesn't work because the order of the contents in the jar will produce different bytes that do not match the original bytes of the original jar.   As such, p2 must omit the PGP verification step in such a case.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/discussions/1502

PR #375, opened 2023-11-04
Don't attempt to PGP verify folder-based content

merks:
https://github.com/eclipse-equinox/p2/issues/374


laeubi (approved):
Seems easy and valid to assume, I wonder if the same is true fro checksum verifiers?
Will such content be reported as unsigned afterwards?

merks:
If other verifiers were a problem (i.e., if they failed) we'd have noticed long ago. 😱   Moreover, they are not an issue because those properties are not in the target platform's artifacts.xml.

laeubi:
> they are not an issue because those properties are not in the target platform's artifacts.xml

Should md5 sha256 and alike not be part of artifacts? ...

merks:
In general, artifact properties of the source artifact descriptor are **not transferred** to the target except when doing raw transfers, i.e., when actually mirroring a repository.  For operations such as populating a target platform or installing/updating into an installation, the processing instructions are applied so the target might end up being a folder.  In the past, one would download the pack200 artifact, if available, and unpack200 it to create the jar at the destination.  So in general the checksums are not applicable after the transfer.   Remember well that I had to specialize p2 when PGP signing was introduced to ensure that the PGP signature and key are, unlike the other properties, transferred from source to target in all cases, otherwise the signing information is lost...

Issue #373, opened 2023-11-03
Supply a general entry point for accessing P2 pools


laeubi:
There is some demands to share already downloaded data when accessing updatesites and different tools have different locations/caches to store items (see https://github.com/eclipse-pde/eclipse.pde/issues/875, https://github.com/eclipse-tycho/tycho/issues/2990).

We should define a general entry point for such tools to discover, read and add such pools with a global registry.

It could start as simple as `P2Pool.getPools()` returning a Stream of `IArtifactRepository`s that list all known to the registry.
Later on we might want  `P2Pool.addPool(Path)` to add one permanently and `P2Pool.addPool(IArtifactRepository)` to add one transitively for the current run of the virtual machine.

mmosconi1:
It seems that `~/.p2/pools.info` is regarded as fixed/global reference for bundle pools. But this is only true if `~/.p2` is set as Agent location!

When I change to a different Agent (and Bundle Pool) via Oomph "Bundle Pool Management", ~/.p2/pools.info

- is empty, if I removed the original Agent
- contains only the **old** bundle pool, if I didn't remove the old Agent 

Oomph itself seems to get its information from agents.info / defaults.info below `~/.eclipse/org.eclipse.oomph.p2`.
Maybe this or some other place under  `~/.eclipse` is better suited to host the "global" pools index?

Background: in our company (using Windows mainly), installation files and/or caches should not be stored under the user home directory. For that reason, we define a different Agent/Pool location.

mmosconi1:
> But that’s like 1 in 1000 users, or fewer.

I could imagine that many companies don't want/allow such things goings to user directories (e.g. think about "roaming profiles" or shared desktops or virus scanner exclusions).
Just be aware that (technically) `~/.p2/pools.info` is not a safe anchor. Maybe we need sth. similar as for Maven local repositories, where I can override the default location (`~/.m2/repository`) via `~/.m2/settings.xml` or `-Dmaven.repo.local`.

When thinking about support for Agents: if someone goes for different Agents to separate scopes for some reason*, I think PDE or Tycho should not use a "flattened" list of bundle pools across Agents, but only those of the "active" Agent. For PDE this seems clear (the Agent used in this workspace or Target Platform), for Tycho I have no idea how to define such "active" Agent. In the Installer, I have to select exactly one bundle pool already.

(*) e.g. regarding security level ("certified" from trusted sources vs. "wild")

laeubi:
[...]

PR #377, opened 2023-11-05
Provide RepositoryHelper.getSharedBundlePools()

merks:
This returns shared bundle pools as used by the Eclipse Installer.


bundles/org.eclipse.equinox.p2.repository/META-INF/MANIFEST.MF
```diff
@@ -31,7 +31,8 @@ Export-Package: org.eclipse.equinox.internal.p2.persistence;
    org.eclipse.equinox.p2.publisher,
    org.eclipse.equinox.p2.repository.tools,
    org.eclipse.equinox.p2.ui,
-   org.eclipse.equinox.p2.updatesite",
+   org.eclipse.equinox.p2.updatesite,
+   org.eclipse.pde.core",
```
laeubi, line 35: This looks strange

HannesWell (commented):
Looks good to me.

I wonder are there already classes/packages in P2 dedicated to bundle-pooling? Or is a bundle pool for p2 simply a local repository?
The p2 wiki is not clear about this: https://wiki.eclipse.org/Equinox/p2/Getting_Started#Bundle_pooling

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
```diff
@@ -138,4 +144,36 @@ public static <T> IRepository<T> createMemoryComposite(IProvisioningAgent agent,
 		return null;
 	}
 
+	/**
+	 * Returns an unmodifiable list of global shared bundle pools, e.g., as used by
+	 * the Eclipse Installer.
+	 *
+	 * @return an unmodifiable list of global shared bundle pools.
+	 *
+	 * @throws IOException if there are problems loading the pool information.
```
laeubi, line 153: On the long term I think we might simply return an empty list in such case?

HannesWell (approved):

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
```diff
@@ -138,4 +144,36 @@ public static <T> IRepository<T> createMemoryComposite(IProvisioningAgent agent,
...
+	 */
+	@SuppressWarnings("nls")
+	public static List<Path> getSharedBundlePools() throws IOException {
+			Path bundlePools = Path.of(System.getProperty("user.home"), ".p2/pools.info");
```
laeubi, line 157: Indentation looks wrong here

merks: Feel free to improve things.  This often happens because only touched lines are formatted, and I removed the if guard.  Sorry about that.

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
merks: Yes, I wondered too about guarding all exceptions within the method....

PR #378, opened 2023-11-05
Remove unused distinction between Plug-ins and Fragments in products

HannesWell:
Remove unused distinction between Plug-ins and Fragments in `.product` files.
There is no usage of a product's fragments in P2 or PDE.
I also checked Tycho (5 but I assume it's the same for Tycho 4) and it never calls `IProductDescriptor.getFragments()` and calls `IProductDescriptor.getBundles(boolean includeFragments)` with `includeFragments=true`.

So it looks like this distinction is unused and can be removed too.

Consequently PDE support for adding a fragment attribute for fragments listed in a product file can be removed too and I will create a PR dedicated to that.

This was originally part of https://github.com/eclipse-equinox/p2/pull/368, but since the second part there needs some more adoptions in PDE I will postpone that but would like to land this for the December release.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/internal/p2/publisher/eclipse/IProductDescriptor.java
+++ b/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/internal/p2/publisher/eclipse/IProductDescriptor.java
@@ -50,21 +50,15 @@ public interface IProductDescriptor {
 	/**
 	 * Returns the bundles listed in this product. Note: These bundles are only part of 
 	 * the product if {@link #useFeatures()} returns <code>false</code>.
-	 * @param includeFragments whether or not to include the fragments in the return value
 	 * @return the list of bundles in this product.
 	 */
-	public List<IVersionedId> getBundles(boolean includeFragments);
+	public List<IVersionedId> getBundles();
 
 	/**
-	 * Returns <code>true</code> when <code>getBundles(includeFragments)</code> returns a non-empty list.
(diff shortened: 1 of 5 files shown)


laeubi:
> Consequently PDE support for adding a fragment attribute for fragments listed in a product file can be removed too and I will create a PR dedicated to that.

Never been aware of that one can switch it, I think its there for the same historic reasons it was added to features...

laeubi (approved):
Its all internal API and adjustments seems minimal, maybe @merks has some concerns but from Tycho side we will simply adopt.

HannesWell:
> > Consequently PDE support for adding a fragment attribute for fragments listed in a product file can be removed too and I will create a PR dedicated to that.
> 
> Never been aware of that one can switch it, I think its there for the same historic reasons it was added to features...

I assume the same about the history.
You cannot explicitly set it. The PDE Product editor (currently) just adds a `fragment` attribute if a plugin is a fragment:
```
   <plugins>
      <plugin id="org.eclipse.core.net.win32" fragment="true"/>
      <plugin id="org.eclipse.pde.core"/>
   </plugins>
```

Here is the PDE pendant: https://github.com/eclipse-pde/eclipse.pde/pull/882

> Its all internal API and adjustments seems minimal, maybe @merks has some concerns but from Tycho side we will simply adopt.

For Tycho I already started a commit locally to adapt to all the internal changes done in P2 in this release.
I'll create a PR with that once the release is further down the ramp.

merks (approved):
Fortunately I don't call the changed methods so it's not a problem.

HannesWell:
Created N&N entry for this: https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/84

PR #379, opened 2023-11-05
Improve RepositoryHelper.getSharedBundlePools() to avoid exceptions

merks:
Just log an exception and return an empty list in the unlikely event of failure.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
+++ b/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
@@ -15,7 +15,6 @@
 package org.eclipse.equinox.internal.p2.repository.helpers;
 
 import java.io.File;
-import java.io.IOException;
 import java.net.*;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


merks:
Hopefully this is the last round!  😱

merks:
FYI, I've already committed these changes to Oomph and they are in a nightly build:

https://github.com/eclipse-oomph/oomph/issues/41

So soon the encoding.info file will exist everywhere...

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
```diff
@@ -149,18 +148,19 @@ public static <T> IRepository<T> createMemoryComposite(IProvisioningAgent agent,
...
-					return lines.map(Path::of).filter(Files::isDirectory).toList();
-				}
+	public static List<Path> getSharedBundlePools() {
+		Path bundlePools = Path.of(System.getProperty("user.home"), ".p2/pools.info");
+		if (Files.isRegularFile(bundlePools)) {
+			try (Stream<String> lines = Files.lines(bundlePools, getSharedBundlePoolEncoding())) {
+				return lines.map(Path::of).filter(Files::isDirectory).toList();
+			} catch (Exception ex) {
+				LogHelper.log(
+						new Status(IStatus.WARNING, Activator.ID, "The bundle pool load failed: " + bundlePools, ex));
```
HannesWell, line 160:
```suggestion
						Status.warning("The bundle pool load failed: " + bundlePools, ex));
```
would be a bit simpler. :)

HannesWell (approved):
LGTM, just one little suggestion. :)

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
laeubi: `ILog.error("The bundle pool load failed: " + bundlePools, ex);` would even be simpler ;-)

laeubi (approved):
Makes sense as the user most likely can't do anything than log/ignore the exception if it will ever arise...

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
HannesWell: > `ILog.warn("The bundle pool load failed: " + bundlePools, ex);` would even be simpler ;-)

I think it needs to be `ILog.get().warn(...`, but I agree that there is a lot of potential in P2 to improve logging (and other things).^^

merks: What was the comment?  Perfect is the enemy of good.  😈

PR #382, opened 2023-11-07
Remove blank Javadoc

jukzi:
This commit cleans up Javadoc that does not add information. It resolves ecj warnings:
 `Javadoc: Description expected after ...`
It helps to prevent future empty javadoc by disabling missingJavaDoc warnings. This resolves
 `Javadoc: Missing ...`

The modification is a result of regular expression search&replace:

in files `*.java`
 `^[\s]*\*[\s]*(@return|@param[\s]*[^\s]+|@throws[\s]*[^\s]+)\R([\s]*\*[\s]*@|[\s]*\*/\R)`
 ->`$2`
 `^([\s]*\*[\s]*\R)([\s]*\*/\R)`
 ->`$2`
 `^[\S\t ]*/\*\*\R[\s]*\*/\R`
 ->``

in files `org.eclipse.jdt.core.prefs`
 `org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc(Comments|Tags)\=[^\s]*`
 ->`org\.eclipse\.jdt\.core\.compiler\.problem\.missingJavadoc$1\=ignore`

### DIFF
--- a/bundles/org.eclipse.equinox.frameworkadmin.equinox/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.frameworkadmin.equinox/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.frameworkadmin.equinox;singleton:=true
-Bundle-Version: 1.3.0.qualifier
+Bundle-Version: 1.3.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Import-Package: org.eclipse.equinox.frameworkadmin;version="[2.0.0,3.0.0)",
--- a/bundles/org.eclipse.equinox.frameworkadmin.equinox/src/org/eclipse/equinox/internal/frameworkadmin/equinox/EquinoxBundlesState.java
+++ b/bundles/org.eclipse.equinox.frameworkadmin.equinox/src/org/eclipse/equinox/internal/frameworkadmin/equinox/EquinoxBundlesState.java
@@ -51,8 +51,6 @@ public class EquinoxBundlesState implements BundlesState {
 	 * eclipse.exe will launch a fw where plugins/org.eclipse.osgi_*.*.*.*.jar is an
 	 * implementation of fw.
 	 *
-	 * @param launcherData
-	 * @param configData
 	 * @return File of fwJar to be used.
 	 */
 	static File getFwJar(LauncherData launcherData, ConfigData configData) {
@@ -238,11 +236,6 @@ public class EquinoxBundlesState implements BundlesState {
 	/**
 	 * If useFwPersistentData flag equals false, this constructor will not take a
 	 * framework persistent data into account. Otherwise, it will.
-	 *
-	 * @param context
-	 * @param fwAdmin
-	 * @param manipulator
-	 * @param useFwPersistentData
 	 */
 	EquinoxBundlesState(BundleContext context, EquinoxFwAdminImpl fwAdmin, Manipulator manipulator, PlatformAdmin admin,
 			boolean useFwPersistentData) {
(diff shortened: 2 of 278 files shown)


jukzi:
The build failed with wrong API error messages:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.4-SNAPSHOT:verify (verify) on project org.eclipse.equinox.p2.ui: There are API errors:
[ERROR] src/org/eclipse/equinox/p2/ui/AcceptLicensesWizardPage.java:49 The type org.eclipse.equinox.p2.ui.AcceptLicensesWizardPage has been removed from org.eclipse.equinox.p2.ui_2.8.200
[ERROR] src/org/eclipse/equinox/p2/ui/ICopyable.java:25 The type org.eclipse.equinox.p2.ui.ICopyable has been removed from org.eclipse.equinox.p2.ui_2.8.200
...
```

laeubi:
@jukzi have you checked the the generated javadoc is still correct? I have the feeling that even though empty these end up in the generated doc while now e.g. `@throws` removed it might no longer show the exceptions thrown there.

Also silencing a valid warning seems odd to me, because now I would expect a warning that the param/return/whatever is missing ...

jukzi:
> @jukzi have you checked the the generated javadoc is still correct?

I hope this proves enough:
image

When a parameter is not listed in the javadoc it is added automatically. If instead the parameter is listed in the javadoc but the description is missing it's an error. There is no point in adding an parameter to javadoc without adding a description. Optimally one would describe the parameter but there is nobody that does that for ages

laeubi:
> When a parameter is not listed in the javadoc it is added automatically.

Great, then this looks fine, but we need a green build here whether or not is it considered "valid", you can use the offending bundle and strip it down like I did with the pde.build bundle:

https://github.com/eclipse-tycho/tycho/commit/c7733ee36a5bab7fc1714bab11aec4e367378ec4

if it is reproducible we can look how to fix that but this realistically will not get into a Tycho release before the 2023-12 release ...

jukzi:
> eclipse-tycho/tycho@c7733ee

@laeubi i do not understand a single line of that commit. Can you help please? The jenkins build is constantly failing with the same false positive API errors
https://ci.eclipse.org/equinox/job/p2/job/PR-382/3/console
https://ci.eclipse.org/equinox/job/p2/job/PR-382/4/console

laeubi:
You copy the offending project to https://github.com/eclipse-tycho/tycho/tree/master/tycho-its/projects/api-tools/embedded-jars

then add it to the `pom.xml` as a module

Next you download the baseline jar from the previous release here https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/02aa3dcd09c79bc68c19ba898770f1550e92b85c/eclipse-platform-parent/pom.xml#L143

and add it to the "baseline repo": https://github.com/eclipse-tycho/tycho/tree/master/tycho-its/repositories/api-tools (remove all requirements from the p2 unit as it is not needed/wanted)

then you run 

- `mvn clean install -DskipTests -T1C`
- cd into `tycho-its`
- `mvn clean install -Pits -Dtest=org.eclipse.tycho.test.apitools.ApiToolsTest`

and should see the test fails with the same error in Jenkins

laeubi:
By the way I think this is mostly caused by p2 projects are missing 

`output.. = bin/`

in the build properties, no clue why that is left out (maybe its the default in PDE?!?)

jukzi:
That is beyond my skills. I have 0 knowledge of maven, tycho, poms, baselines etc , all i know is the errors are wrong. - i did not change the API

jukzi:
> By the way I think this is mostly caused by p2 projects are missing
> 
> `output.. = bin/`
> 
> in the build properties, no clue why that is left out (maybe its the default in PDE?!?)

if you know whats wrong could you please fix it?

laeubi:
> > By the way I think this is mostly caused by p2 projects are missing
> > `output.. = bin/`
> > in the build properties, no clue why that is left out (maybe its the default in PDE?!?)
> 
> if you know whats wrong could you please fix it?

Just open build properties of project with error and add it there, then push and see if it helps ...

jukzi:
> Just open build properties of project with error and add it there, then push and see if it helps ...

any chance to fix tycho (or whatever) to use org.eclipse.jdt.internal.core.JavaProject.defaultOutputLocation()? or even hardcode "bin" as for example in org.eclipse.jdt.internal.junit.launcher.JUnitRuntimeClasspathEntry.developmentModeEntry()? The constant seems to be used so often that it can not be changed anyway

jukzi:
build passed

PR #384, opened 2023-11-10
Improve RepositoryHelper.getSharedBundlePools to load all agent pools

merks:
Oomph has moved the encoding.info to be in the same folder as the agents.info which lists all the agent locations (rather than in the folder of the default agent's location). The agents.info file can be loaded to determine the location of the pools.info of each agent. This refined approach covers the full generality of Oomph's shared agent manager allows all pools of all agents to be located with no assumptions about defaults.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
+++ b/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
@@ -19,8 +19,7 @@ import java.net.*;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
 import java.util.stream.LongStream;
 import java.util.stream.Stream;
 import org.eclipse.core.runtime.*;
@@ -151,20 +150,37 @@ public class RepositoryHelper {
 	 */
 	@SuppressWarnings("nls")
 	public static List<Path> getSharedBundlePools() {
-		Path bundlePools = Path.of(System.getProperty("user.home"), ".p2/pools.info");
-		if (Files.isRegularFile(bundlePools)) {
-			try (Stream<String> lines = Files.lines(bundlePools, getSharedBundlePoolEncoding())) {
-				return lines.map(Path::of).filter(Files::isDirectory).toList();
+		Path agentsInfoLocation = getAgentManagerLocation().resolve("agents.info");
+		if (Files.isRegularFile(agentsInfoLocation)) {
+			Charset encoding = getSharedAgentEncoding();
+			try (Stream<String> agentLocations = Files.lines(agentsInfoLocation, encoding)) {
+				List<Path> result = new ArrayList<>();
+				agentLocations.map(agentLocation -> Path.of(agentLocation, "pools.info")).filter(Files::isRegularFile)
+						.forEach(poolsInfoLocation -> {
+							try (Stream<String> poolLocations = Files.lines(poolsInfoLocation, encoding)) {
+								poolLocations.map(Path::of)
+										.filter(poolLocation -> Files.exists(poolLocation.resolve("artifacts.xml")))
+										.forEach(result::add);
+							} catch (Exception ex) {
+								LogHelper.log(Status.warning("The bundle pools load failed: " + poolsInfoLocation, ex));
+							}
+						});
+				return result;
 			} catch (Exception ex) {
-				LogHelper.log(Status.warning("The bundle pool load failed: " + bundlePools, ex));
+				LogHelper.log(Status.warning("The agents load failed: " + agentsInfoLocation, ex));
 			}
 		}
 		return List.of();
 	}
 
 	@SuppressWarnings("nls")
-	private static Charset getSharedBundlePoolEncoding() {
-		Path encodingInfo = Path.of(System.getProperty("user.home"), ".p2/encoding.info");
+	private static Path getAgentManagerLocation() {
+		return Path.of(System.getProperty("user.home"), ".eclipse/org.eclipse.oomph.p2");
+	}
+
+	@SuppressWarnings("nls")
+	private static Charset getSharedAgentEncoding() {
+		Path encodingInfo = getAgentManagerLocation().resolve("encoding.info");
 		if (Files.isRegularFile(encodingInfo)) {
 			try {
 				return Charset.forName(Files.readString(encodingInfo).trim());


bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
```diff
@@ -151,20 +150,31 @@ public static <T> IRepository<T> createMemoryComposite(IProvisioningAgent agent,
...
 			}
 		}
 		return List.of();
 	}
 
 	@SuppressWarnings("nls")
-	private static Charset getSharedBundlePoolEncoding() {
-		Path encodingInfo = Path.of(System.getProperty("user.home"), ".p2/encoding.info");
+	private static Charset getSharedAgentEncoding() {
+		Path encodingInfo = Path.of(System.getProperty("user.home"), ".eclipse/org.eclipse.oomph.p2/encoding.info");
```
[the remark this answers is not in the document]
merks: Yes, that avoid duplicating constants/logic.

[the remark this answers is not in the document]
merks: Done.

merks:
FYI, this change is necessary because Oomph as moved the location of the encoding.info file so at this point it's arguably a bug fix.

That said, of course I do appreciate careful reviews!

laeubi (approved):
Looks good.

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
```diff
@@ -151,20 +150,36 @@ public static <T> IRepository<T> createMemoryComposite(IProvisioningAgent agent,
...
+		Path agentsInfoLocation = getAgentManagerLocation().resolve("agents.info");
+		if (Files.isRegularFile(agentsInfoLocation)) {
+			Charset encoding = getSharedAgentEncoding();
+			try (Stream<String> agentLocations = Files.lines(agentsInfoLocation, encoding)) {
+				List<Path> result = new ArrayList<>();
+				agentLocations.map(agentLocation -> Path.of(agentLocation, "pools.info")).filter(Files::isRegularFile)
+						.forEach(poolsInfoLocation -> {
+							try (Stream<String> poolLocations = Files.lines(poolsInfoLocation, encoding)) {
+								poolLocations.map(poolLocation -> Path.of(poolLocation, "artifacts.xml"))
+										.filter(Files::isRegularFile).map(Path::getParent).forEach(result::add);
```
HannesWell, line 162: An alternative suggestion to have no back-and-fourth on the path.
```suggestion
								poolLocations.map(Path::of)
										.filter(poolLocation -> Files.exists(poolLocation.resolve("artifacts.xml")))
										.forEach(result::add);
```

HannesWell (approved):
Looks good. Thanks for the follow up.

Just one minor suggestion below (but probably also a matter of taste).

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/RepositoryHelper.java
merks: I had the **same** thought.  What's there is slightly more concise looking, but is not really how one would express it if one wasn't using lambdas...

merks: I take it back.  It's not really more concise.  I will change it.

HannesWell: Yes this is really a matter of taste, which may vary between persons or even in time at the same person 😅

Btw. I'm surprised that there is no constant in P2 for `artifacts.xml`?

merks: There 24 matches in the SDK workspace, including two in pdebuild, but only one place with private static constant that of course is only used local in that class:

org.eclipse.equinox.internal.p2.engine.ProfileMetadataRepository.ARTIFACTS_XML

iloveeclipse:
I haven't analyzed yet, but it seems this PR might caused some severe problem, look at the  test results: https://download.eclipse.org/eclipse/downloads/drops4/I20231111-0600/testResults.php

Unfortunately I only found errors that errors happen on startup, but no real log fie, see
https://download.eclipse.org/eclipse/downloads/drops4/I20231111-0600/testresults/consolelogs/ep430I-unit-macM1-java17_macosx.cocoa.aarch64_17_consolelog.txt

```
using launcher:  
An error has occurred. See the log file
/Users/genie.releng/workspace/AutomatedTests/ep430I-unit-macM1-java17/workarea/I20231111-0600/eclipse-testing/workspace/.metadata/.log.
```

iloveeclipse:
In this log it says it can't find pde.core?

https://download.eclipse.org/eclipse/downloads/drops4/I20231111-0600/testresults/ep430I-unit-macM1-java17_macosx.cocoa.aarch64_17/directorLogs/director-org.eclipse.ant.tests.core.log

```
Starting application: 1334
Installing org.eclipse.ant.tests.core 3.7.200.v20230920-0631.
Installing org.eclipse.test 3.5.400.v20231108-0838.
Installing org.eclipse.ant.optional.junit 3.3.400.v20230714-1309.
Installing org.eclipse.test.performance 3.19.400.v20231108-0838.
Installing org.eclipse.test.performance.win32 3.1.600.v20230714-1309.
Installation failed.
Cannot complete the install because one or more required items could not be found.
	Software being installed: Ant Core Test Plugin 3.7.200.v20230920-0631 (org.eclipse.ant.tests.core 3.7.200.v20230920-0631)
	Missing requirement: Ant Core Test Plugin 3.7.200.v20230920-0631 (org.eclipse.ant.tests.core 3.7.200.v20230920-0631) requires 'osgi.bundle; org.eclipse.pde.core 0.0.0' but it could not be found
```

merks:
I’m not sure how this can be in any way related.  This method would only be used only after resolution succeeds but resolution itself is failing.  And I’m very sure the method returns an empty list in an environment where oomph has not populated the user.home folder.

iloveeclipse:
I simply don't see other changes merged...

@laeubi : any tycho / surefire updates?

merks:
Note too in particular that only pde calls this method and only during target platform resolutions.

laeubi:
@iloveeclipse platform uses a released version of Tycho and as far as I know the eclipse testframework uses ant/pdebuild anyways (@akurtakov ?) so I have no clue how this change or Tycho can affect this, maybe a concurrent build or network problem?

merks:
Unless some PDE target platform resolution test is involved, it’s not even theoretically possible for the method to be called.

merks:
Probably it’s a good idea to kick the build again.

iloveeclipse:
> Probably it’s a good idea to kick the build again.

Let see if other builds will show same issue. If not, it was a glitch in the Matrix and we can restart Mac build.

iloveeclipse:
I see other builds were OK, so O will try to restart Mac build

iloveeclipse:
Next build was OK. So it was some infrastructure issue...

Issue #381, opened 2023-11-06
Not able to communicate through proxy


SebasHein:
I had a discussion about oomph not able to communicate through proxy anymore with Ed Merks already - see here: https://www.eclipse.org/forums/index.php/m/1861818/#msg_1861818

He suggests to open an issue here and to reproduce the behaviour with the latest eclipse build.

I downloaded I-Build of Eclipse from here https://download.eclipse.org/eclipse/downloads/drops4/I20231105-1800/

My environment is behind a proxy that requests to authentificate with user and password.
In the eclipse preferences I set the same settings as was working well with Eclipse 4-29 ( 2023-09).
Active Provider: Manual
HTTP and HTTPS: Host and Port, user and password according to what was working in former release.
The SOCKS is activated (checkbox is checked) but host, port etc. is empty.
No proxy bypasses except localhost and 127.0.0.1

I then tried to install new software with help → install new software
The update sites I tried were

- https://download.eclipse.org/eclipse/updates/4.30
- https://download.eclipse.org/releases/2023-12

I immediately get Proxy authentication required exceptions
org.eclipse.equinox.p2.core.ProvisionException: HTTP Proxy Authentication Required: https://download.eclipse.org/releases/2023-12/content.xml
	at org.eclipse.equinox.internal.p2.repository.CacheManager.createCache(CacheManager.java:250)
	at org.eclipse.equinox.internal.p2.metadata.repository.SimpleMetadataRepositoryFactory.getLocalFile(SimpleMetadataRepositoryFactory.java:69)
	at org.eclipse.equinox.internal.p2.metadata.repository.SimpleMetadataRepositoryFactory.load(SimpleMetadataRepositoryFactory.java:89)
	at org.eclipse.equinox.internal.p2.metadata.repository.MetadataRepositoryManager.factoryLoad(MetadataRepositoryManager.java:63)
	at org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.loadRepository(AbstractRepositoryManager.java:791)
	at org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.loadRepository(AbstractRepositoryManager.java:689)
	at org.eclipse.equinox.internal.p2.metadata.repository.MetadataRepositoryManager.loadRepository(MetadataRepositoryManager.java:110)
	at org.eclipse.equinox.internal.p2.metadata.repository.MetadataRepositoryManager.loadRepository(MetadataRepositoryManager.java:105)
[...]

PR #385, opened 2023-11-12
Require ecf http5 in core feature

laeubi:
FYI @akurtakov @merks @tjwatson requires PL/PMC approval

See
- https://github.com/eclipse-equinox/p2/issues/381
- https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/86


merks:
PMC +1 🚗

PR #386, opened 2023-11-20
BrandingIron: use java.nio.file.Files convenience with consistent UTF_8

jukzi:


bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/internal/p2/publisher/eclipse/BrandingIron.java
```diff
@@ -504,7 +506,7 @@ private void copyMacIni(ExecutablesDescriptor descriptor, File initialRoot, File
 
 		StringBuffer buffer;
 		try {
-			buffer = readFile(ini);
+			buffer = new StringBuffer(Files.readString(ini.toPath(), StandardCharsets.UTF_8));
```
laeubi, line 509: I can't find in the original that UTF-8 is sued but system encoding instead.

laeubi (commented):
Can you explain why UTF-8 is to be used here?

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/internal/p2/publisher/eclipse/BrandingIron.java
jukzi: `In Java SE 9, properties files are loaded in UTF-8 encoding`
https://docs.oracle.com/javase/9/intl/internationalization-enhancements-jdk-9.htm#JSINT-GUID-974CF488-23E8-4963-A322-82006A7A14C7

laeubi: An (eclipse) ini file is not a java properties file...

merks: Yes, be really careful how this is written!  I know for sure that on Windows this is being written using the default encoding, not UTF-8 because I was just testing concerns that I have about Java 21 and its defaults.  I ran into cases where the installed IDE won't launch and then I look at they differences:

image

Getting the encoding wrong is something you might not noticed until you installed into a folder with non-ASCII characters.  In the one case that works, the default system encoding is used and in the case that doesn't work, the debug launched installer I set to UTF-8 encoding.

I'm still not sure how this Java 21 stuff is going to work?  😱   Some of these files are read by the native launcher and I don't know what encoding it uses when reading these files nor what exactly it reads, but given the JRE is specified in the eclipse.ini, I matters what encoding is used...

jukzi: So we have at least 4 codes that read that file and interpret it differently?

org.eclipse.equinox.internal.frameworkadmin.equinox.EquinoxFwConfigFileParser.loadProperties(File)
org.eclipse.equinox.internal.frameworkadmin.equinox.utils.FileUtils.loadFile(File)
org.eclipse.equinox.internal.p2.publisher.eclipse.BrandingIron.copyMacIni(ExecutablesDescriptor, File, File, String)
https://github.com/eclipse-equinox/equinox/blob/8beb9ce0b1dea1176b06ad7300f2663ff6db4eff/features/org.eclipse.equinox.executable.feature/library/eclipseConfig.c#L146-L150

jukzi: i changed this commit to not touch the encoding

merks: This is what makes me so paranoid about suddenly things being UTF-8 on Windows where everything is currently some ISO-Latin type of encoding, at least on my OS.  But I know that other countries have different all kinds of other encodings ad their defaults...

merks: I'm not sure what you are suggesting.  In the past they are read and written using the default system encoding.   Both your changes change that to UTF-8 now.   If that's your intent, then you need to reconsider.  I think it should use Charset.defaultCharset() to preserve existing semantics.

merks: Hmm, I'm not sure when this branding iron thing runs.  Just at build time I suppose and then we are talking about the default encoding of the build-time process, so maybe it doesn't really matter to use UTF-8 always for this case...

jukzi:
@merks can this be merged?

merks:
I think it's only used during builds, so yes.

merks:
I assume now that the minor version needs an increment:

image

jukzi:
> I assume now that the minor version needs an increment:

version bump added. i wonder why build does not fail without.

PR #392, opened 2023-12-04
Wrapped the OperationCanceledException in InterruptedException

amartya4256:
https://github.com/eclipse-platform/eclipse.platform/pull/853

Wrapped the OperationCanceledException in InterruptedException

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui.discovery/src/org/eclipse/equinox/internal/p2/ui/discovery/operations/DiscoveryInstallOperation.java
+++ b/bundles/org.eclipse.equinox.p2.ui.discovery/src/org/eclipse/equinox/internal/p2/ui/discovery/operations/DiscoveryInstallOperation.java
@@ -93,7 +93,9 @@ public class DiscoveryInstallOperation implements IRunnableWithProgress {
 				monitor.done();
 			}
 		} catch (OperationCanceledException e) {
-			throw new InterruptedException();
+			InterruptedException interruptedException = new InterruptedException(e.getLocalizedMessage());
+			interruptedException.initCause(e);
+			throw interruptedException;
 		} catch (Exception e) {
 			throw new InvocationTargetException(e);
 		}


HeikoKlare:
I have not taken a look into the Equinox repositories for this pattern of exception encapsulation yet. But the proposed change seems to make sense to me, as it only adds some potentially useful information about an exception cause.

bundles/org.eclipse.equinox.p2.ui.discovery/src/org/eclipse/equinox/internal/p2/ui/discovery/operations/DiscoveryInstallOperation.java
```diff
@@ -93,7 +93,9 @@ public void run(IProgressMonitor progressMonitor) throws InvocationTargetExcepti
 				monitor.done();
 			}
 		} catch (OperationCanceledException e) {
-			throw new InterruptedException();
+			InterruptedException interruptedException = new InterruptedException(e.getLocalizedMessage());
```
HannesWell, line 96: Can't the catched exception directly be passed to the constructor of `InterruptedException`?

laeubi:
In this particular case please note that the exception is generated in checkCancel and then is catched and rewrapped this seems a bit odd...

HeikoKlare:
> In this particular case please note that the exception is generated in checkCancel and then is catched and rewrapped this seems a bit odd...

That's true. Taking a look at the code shows that the exception is only used as an internal, intermediate representation. The only consumer transforms it into a boolean right after:
https://github.com/eclipse-equinox/p2/blob/f0b03b499c283f07a634f19f43ed5a745dbcd4e4/bundles/org.eclipse.equinox.p2.ui.discovery/src/org/eclipse/equinox/internal/p2/ui/discovery/DiscoveryUi.java#L44-L56

So there is no benefit in adding the "causing" exception here. We also found the same pattern in eclipse.platform where I argued against changing these places where the exception is only used "internally": https://github.com/eclipse-platform/eclipse.platform/pull/853#issuecomment-1808352471

So I have to revert my previous statement: the proposed change should not be merged.

merks:
The change, while following the pattern of similar changes, does not provide actual value in this case.

PR #393, opened 2023-12-05
SimpleBundlesState: remove unused import

jukzi:
to reduce warnings


iloveeclipse (changes requested):
Version was already bumped https://github.com/eclipse-equinox/p2/commit/66d322424785fdbc09ede80ecba5ce2b038741ad

iloveeclipse:
> i removed version bump

Thanks, can you update PR in that case?

jukzi:
> Thanks, can you update PR in that case?

uups. done

PR #399, opened 2023-12-09
Replace StringBuffer by StringBuilder in p2.metadata

HannesWell:
This PR consists of two steps

1. Apply 'Convert `StringBuffer` to `StringBuilder`' clean-up to p2.metadata
2. Manually restore API compatibility after `StringBuffer` clean-up
Keep the old `StringBuffer` accepting method as default implementations that redirect to the `StringBuilder` impls.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.metadata/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.metadata/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.metadata;singleton:=true
-Bundle-Version: 2.8.100.qualifier
+Bundle-Version: 2.9.0.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.equinox.internal.p2.metadata;
--- a/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/BasicVersion.java
+++ b/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/BasicVersion.java
@@ -63,7 +63,7 @@ public abstract class BasicVersion extends Version {
 	public abstract String getQualifier();
 
 	/**
-	 * Appends the original for this version onto the <code>sb</code> StringBuffer
+	 * Appends the original for this version onto the <code>sb</code> StringBuilder
 	 * if present.
 	 * @param sb The buffer that will receive the raw string format
 	 * @param rangeSafe Set to <code>true</code> if range delimiters should be escaped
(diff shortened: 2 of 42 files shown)


vogella (commented):
LGTM, also great that the old methods are deprecated

merks:
FYI, even the most innocent changes can cause problems.  😱 

https://github.com/eclipse-cbi/p2repo-aggregator/commit/1c8405289fdb6a796ca0375e110aba4106c52e4e#diff-2b3345c61b2e1aacd5d16a21cf18212dec2d2d5f66041400161acaf9207fbd8c

The aggregator does rather deep integration with p2...

HannesWell:
> LGTM, also great that the old methods are deprecated

Thanks Lars. Yes and I'm happy this API usage does not prevent us from using the clean-up.
I also not marked it for removal since keeping it does not cause any problems (now).

> FYI, even the most innocent changes can cause problems. 😱
> 
> [eclipse-cbi/p2repo-aggregator@1c84052#diff-2b3345c61b2e1aacd5d16a21cf18212dec2d2d5f66041400161acaf9207fbd8c](https://github.com/eclipse-cbi/p2repo-aggregator/commit/1c8405289fdb6a796ca0375e110aba4106c52e4e#diff-2b3345c61b2e1aacd5d16a21cf18212dec2d2d5f66041400161acaf9207fbd8c)
> 

Unfortunately yes. 😞
But what was the problem here exactly? That the hash-code value changed in the P2 implementation?
I'm not sure but I would say one should not relay on an exact hash-code value, shouldn't one?

But do you think this change is still fine? Unless somebody uses the internal impls it should be.

merks:
Yes, the aggregator implements all the p2 APIs directly via modeled classes so contained a copy of the hash-code method.  Wicked evil.   Even just the overrides of equals on the generated modeled classes is also a violation of EMF's contract for EObjects:

https://github.com/eclipse-emf/org.eclipse.emf/blob/f8402dcec2a7a99fa83afd5e22635cc77fcd207a/plugins/org.eclipse.emf.ecore/src/org/eclipse/emf/ecore/EObject.java#L65-L66

Also wicked evil.

I think anything that's broken by the changes you propose can be fixed.  And, unlike EMF and Oomph, the aggregator is not constrained to keep working on older versions.  So don't worry, it's just an FYI that even the littlest things can break stuff.  (Speaking of little changes breaking things, I'm currently making sure that EMF's tests can run in the IDE after the slf4j changes from the previous release; it's always something.)

HannesWell:
That sounds all quite fragile. But well we now see the result of it. I'm not gonna ask why this is done🙉 


> I think anything that's broken by the changes you propose can be fixed. And, unlike EMF and Oomph, the aggregator is not constrained to keep working on older versions. So don't worry, it's just an FYI that even the littlest things can break stuff. 

OK, then I don't worry 😅

> (Speaking of little changes breaking things, I'm currently making sure that EMF's tests can run in the IDE after the slf4j changes from the previous release; it's always something.)

Is there a specific problem you encounter? Isn't it sufficient to ensure that the aries spifly bundle is started and you have e.g. slf4j-simple in the TP? This reminds me of https://github.com/eclipse-pde/eclipse.pde/pull/644.

PR #400, opened 2023-12-09
Migrate AutomaticUpdateScheduler from IStartup to OSGi EventHandler

HannesWell:
... for the `UIEvents.UILifeCycle.APP_STARTUP_COMPLETE` event.

This is a less Eclipse specific approach and also prevents the user from disabling the start-up task.
The automatic search for updates can still be disabled.

Based on the work in https://github.com/eclipse-platform/eclipse.platform.ui/pull/1362.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/src/org/eclipse/equinox/internal/p2/ui/sdk/scheduler/AutomaticUpdateScheduler.java
+++ b/bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/src/org/eclipse/equinox/internal/p2/ui/sdk/scheduler/AutomaticUpdateScheduler.java
@@ -21,6 +21,7 @@ import java.util.Date;
 import java.util.Random;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.e4.ui.workbench.UIEvents;
 import org.eclipse.equinox.internal.p2.core.helpers.ServiceHelper;
 import org.eclipse.equinox.internal.p2.garbagecollector.GarbageCollector;
 import org.eclipse.equinox.internal.p2.ui.sdk.scheduler.migration.MigrationSupport;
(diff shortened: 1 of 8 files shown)


bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/src/org/eclipse/equinox/internal/p2/ui/sdk/scheduler/AutomaticUpdateScheduler.java
```diff
@@ -56,7 +62,7 @@ public class AutomaticUpdateScheduler implements IStartup {
 	private IUpdateChecker checker;
 
 	@Override
-	public void earlyStartup() {
+	public void handleEvent(Event event) {
```
iloveeclipse, line 65: Shouldn't we check if automated updates are disabled **before** starting the job?

iloveeclipse:
With "Early startup" one could disable updates job. Now it is always started even if updates are not enabled. I think this is a regression.

HannesWell:
> With "Early startup" one could disable updates job. Now it is always started even if updates are not enabled. I think this is a regression.

There is a dedicated preference to control the automatic search for updates:
grafik

So yes now the job always runs, but the job still respects the preference so if you have disabled updates, it is basically a noop.

I can also refactor this to only run the job if the preference is enabled, but I'm not sure what's best in regard of IDE start-up.
Check the preference before running the job or within the job and potentially exit early.
Of course starting a job has an overhead but so does checking a preference.

laeubi:
@HannesWell @iloveeclipse should we further work on this or should we close this?

HannesWell:
> @HannesWell @iloveeclipse should we further work on this or should we close this?

I would still be in favor of having this and was waiting for the response of @iloveeclipse on my previous response:

> I can also refactor this to only run the job if the preference is enabled, but I'm not sure what's best in regard of IDE start-up speed. Check the preference before running the job or within the job and potentially exit early?
> Of course starting a job has an overhead but so does checking a preference.

iloveeclipse:
> Check the preference before running the job or within the job and potentially exit early?
> Of course starting a job has an overhead but so does checking a preference.

I would avoid starting job if possible, because compared with a single preference check extra job adds extra complexity for the overall IDE startup, uses shared resources (worker pool), adds async execution effects.

vogella:
@HannesWell could you move the preference check before starting the job? Would be nice to have this finished.

vogella:
> 2026-03 M1

Lets try 2026-09 M1

vogella:
API errors showing up in the full CI build are unrelated to this change, ,they come from other bundles.

merks:
I'm concerned that pushing multiple p2 changes at the same time will result in increment that will then conflict.

E.g., with this PR:

https://github.com/eclipse-equinox/p2/pull/1075

I assume one can always force push the original commit and then there will be new different version increments, but this approach does result in a much large set of builds and notifications.

Maybe we can control the flow a little bit so get one PR through for a given repository?

vogella:
Build already validated.

PR #402, opened 2023-12-10
Remove @noreference from InstallableUnit.isSingleton/getFilter

merks:


merks:
Guys, do you have any concerns about this?  There are downstream warnings that will be fixed and in the end, one can't really do useful things with p2 without taking this information into account, especially the getFilter information, that's needed here:

https://github.com/eclipse-tycho/tycho/issues/3197

Oomph and the CBI Aggregator need it to, so it's kind of silly how it is now.

HannesWell (approved):
No concerns from me.
The returned type are also all public API so this looks like a clean things.

I just wondered why this wasn't API before or if there are other ways to obtain these data, but from git-blame I only found:
https://github.com/eclipse-equinox/p2/commit/c30c95794fd8b9de3bd362e9164332ed1bd3ad24

With that I think it is fine to make this API.

laeubi (approved):

PR #409, opened 2023-12-14
No error on javadoc in p2 tests

akurtakov:


akurtakov:
Push as the real test will be by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1661

Issue #410, opened 2023-12-14
Provide a mechanism to fetch all workspace (PDE) bundle pools


merks:
PDE maintains a bundle pool in `<workspace>/.metadata/.plugins/org.eclipse.pde.core/.bundle_pool`.  In `org.eclipse.pde.internal.core.target.P2TargetUtils.findWorkspaceRepos(Set<URI>)` it tries to find such bundle pools in other workspaces by looking at the preferences.  Because Oomph maintains a resource with the locations of all installations and all workspace (and their associations) in `<user.home>/.eclipse/org.eclipse.oomph.setup/setups/locations.setup` this approach can be generalized.  Also, because Tycho can benefit from this information, providing a utility in p2, `org.eclipse.equinox.internal.p2.repository.helpers.RepositoryHelper.getWorkspaceBundlePools()`, analogous to `RepositoryHelper.getSharedBundlePools()` would be beneficial.

See also https://github.com/eclipse-equinox/p2/issues/373

PR #411, opened 2023-12-14
Support RepositoryHelper.getWorkspaceBundlePools()

merks:
Provide access to all the PDE bundle pools of all the known workspaces.

https://github.com/eclipse-equinox/p2/issues/410


laeubi (approved):
Very useful!

PR #416, opened 2023-12-18
Perform the mirror step in a batch

laeubi:
Currently the Mirroring is performing each step independently this can lead to considerable slow down as for example the
SimpleArtifactRepository performs a save on every adding of a descriptor.

This wraps the mirror step and update of properties into a batch operation to increase performance.


laeubi:
@merks It seems that one can't chain `executeBatch` for artifact repositories but I can't find usages of Mirroring beside MirrorApplication anyways where it does not batch execute anyways, but maybe Oomph is using that somewhere?

merks:
It looks like Oomph only uses that in a test like:
```
      Mirroring mirror = new Mirroring(sourceRepository, targetRepository, true);
      mirror.setCompare(false);
      mirror.setTransport(transport);
      mirror.setArtifactKeys(keys.toArray(new IArtifactKey[keys.size()]));

      IStatus result = mirror.run(true, false);
      P2CorePlugin.INSTANCE.coreException(result);
```

merks (approved):
Looks fine.

HannesWell (approved):

PR #420, opened 2023-12-28
Do not duplicate failed repository attempts

laeubi:
Currently it can happen that when the same repository is reported as a problem multiple times then the accumulated status show the same repository multiple times.

This do the following:

1) using a set to record failed URIs
2) check if the URI was failing before

as discovered here:
- https://github.com/eclipse-equinox/p2/pull/343

image

image


laeubi:
This does not solve the problem that it seems the same repository is contacted multiple times...

merks (approved):
Looks fine.

HannesWell (approved):

PR #421, opened 2023-12-30
Do not use PackageAdmin for refresh bundles

laeubi:
Currently there are two ways to refresh packages, one using the deprecated PackageAdmin and one using the FrameworkWiring.

This do the following:
- always use FrameworkWiring
- use collections over arrays
- remove some redundant checks and copy of collections

### DIFF
--- a/bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
+++ b/bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
@@ -19,6 +19,7 @@ import java.io.*;
 import java.net.*;
 import java.util.*;
 import java.util.concurrent.CountDownLatch;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.eclipse.equinox.internal.simpleconfigurator.utils.*;
 import org.eclipse.osgi.report.resolution.ResolutionReport.Entry.Type;
@@ -105,7 +106,7 @@ class ConfigApplier {
 		}
 
 		Set<Bundle> prevouslyResolved = getResolvedBundles();
-		Collection<Bundle> toRefresh = new ArrayList<>();
+		Collection<Bundle> toRefresh = new LinkedHashSet<>();
 		Collection<Bundle> toStart = new ArrayList<>();
 		if (exclusiveMode) {
 			toRefresh.addAll(installBundles(expectedState, toStart));
@@ -123,12 +124,19 @@ class ConfigApplier {
 			} else {
 				// In this case the platform is up, we should try to do an incremental resolve
 				// TODO consider removing this case because it can cause inconsistent results.
-				refreshPackages(toRefresh.toArray(new Bundle[toRefresh.size()]), manipulatingContext);
-				if (toRefresh.size() > 0) {
-					Bundle[] additionalRefresh = getAdditionalRefresh(prevouslyResolved, toRefresh);
-					if (additionalRefresh.length > 0)
-						refreshPackages(additionalRefresh, manipulatingContext);
+				Map<String, List<Bundle>> bundlesByBsn = Arrays.stream(manipulatingContext.getBundles())
+						.filter(bundle -> bundle.getSymbolicName() != null)
+						.collect(Collectors.groupingBy(Bundle::getSymbolicName));
+				// Prior to Luna the Equinox framework would refresh all bundles with the same
+				// BSN automatically. This is no longer the case for Luna or other framework
+				// implementations. Here we want to make sure all existing bundles with the
+				// same BSN are refreshed also.
+				Set<Bundle> allSameBSNs = new LinkedHashSet<>(); // maintain order and avoid duplicates
+				for (Bundle bundle : toRefresh) {
+					allSameBSNs.addAll(bundlesByBsn.getOrDefault(bundle.getSymbolicName(), List.of(bundle)));
 				}
+				refreshPackages(allSameBSNs);
+				refreshPackages(getAdditionalRefresh(prevouslyResolved, toRefresh));
 			}
 			if (deepRefresh) {
 				// when refreshing large sets of bundles the resolver sometimes take a choice
@@ -218,11 +226,10 @@ class ConfigApplier {
 				.distinct().filter(bundle -> !doNotRefresh.contains(bundle)).toList();
 	}
 
-	private Bundle[] getAdditionalRefresh(Set<Bundle> previouslyResolved, Collection<Bundle> toRefresh) {
+	private Collection<Bundle> getAdditionalRefresh(Set<Bundle> previouslyResolved, Collection<Bundle> toRefresh) {
 		// This is the luna equinox framework or a non-equinox framework.
 		// Use standard OSGi API.
 		final Set<Bundle> additionalRefresh = new HashSet<>();
-		final Set<Bundle> originalRefresh = new HashSet<>(toRefresh);
 		for (Bundle bundle : toRefresh) {
 			BundleRevision revision = bundle.adapt(BundleRevision.class);
 			if (bundle.getState() == Bundle.INSTALLED && revision != null && (revision.getTypes() & BundleRevision.TYPE_FRAGMENT) != 0) {
@@ -243,7 +250,7 @@ class ConfigApplier {
 				if (foundPayLoadReq) {
 					Collection<BundleCapability> candidates = frameworkWiring.findProviders(hostReq);
 					for (BundleCapability candidate : candidates) {
-						if (!originalRefresh.contains(candidate.getRevision().getBundle())) {
+						if (!toRefresh.contains(candidate.getRevision().getBundle())) {
 							additionalRefresh.add(candidate.getRevision().getBundle());
 						}
 					}
@@ -297,7 +304,7 @@ class ConfigApplier {
 				}
 			}
 		}
-		return additionalRefresh.toArray(new Bundle[additionalRefresh.size()]);
+		return additionalRefresh;
 	}
 
 	private BundleWiring getHostWiring(BundleWiring wiring) {
@@ -331,19 +338,7 @@ class ConfigApplier {
 				toRefresh.add(bundle);
 			}
 		}
-
-		CountDownLatch latch = new CountDownLatch(1);
-		FrameworkListener listener = event -> {
-			if (event.getType() == FrameworkEvent.PACKAGES_REFRESHED) {
-				latch.countDown();
-			}
-		};
-		frameworkWiring.refreshBundles(toRefresh, listener);
-		try {
-			latch.await();
-		} catch (InterruptedException e) {
-			// ignore
-		}
+		refreshPackages(toRefresh);
 	}
 
 	private Set<Bundle> getDoNotRefresh() {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -109,24 +109,20 @@ void install(URL url, boolean exclusiveMode) throws IOException {
...
 				// In this case the platform is up, we should try to do an incremental resolve
 				// TODO consider removing this case because it can cause inconsistent results.
-				refreshPackages(toRefresh.toArray(new Bundle[toRefresh.size()]), manipulatingContext);
-				if (toRefresh.size() > 0) {
-					Bundle[] additionalRefresh = getAdditionalRefresh(prevouslyResolved, toRefresh);
-					if (additionalRefresh.length > 0)
-						refreshPackages(additionalRefresh, manipulatingContext);
+				refreshPackages(toRefresh, manipulatingContext);
+				Collection<Bundle> additionalRefresh = getAdditionalRefresh(prevouslyResolved, toRefresh);
+				if (additionalRefresh.size() > 0) {
```
merks, line 114: !empty()

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -91,7 +91,7 @@ void install(URL url, boolean exclusiveMode) throws IOException {
 		}
 
 		Set<Bundle> prevouslyResolved = getResolvedBundles();
-		Collection<Bundle> toRefresh = new ArrayList<>();
+		Collection<Bundle> toRefresh = new LinkedHashSet<>();
```
merks, line 94: Are duplicates possible?

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
laeubi: Is there any advantage? I try to avoid using negations in conditions when possible as its often harder to understand.

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
laeubi: I don't think so, but this is passed to a method that performs "contains" checks and there the list was copied into a set (possibly to speedup the lookup) so I think it is better to build a set at the first place.

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
merks: The size() method can be O(n) in general.

https://stackoverflow.com/questions/27525369/understanding-the-difference-between-collection-isempty-and-collection-size

laeubi: That's quite hypothetical to me... this all uses local standard collection types (in this case `HashSet`) what implement there `isEmpty` as `size == 0`

merks: Even tools will complain about such uses (comparing size to 0):

laeubi: Actually the check was not required (anymore) anyways so I removed it now completely.

Beside that, I'm not such a great fan of such "tools" that try to be smart for rare cases and mirco optimizations, I think the only real world example is a [ConcurrentSkipListSet](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentSkipListSet.html) where checks for empty/size are meaningless anyways when we assume a high concurrent read/write system...

> Some will state that comparing size() to 0 is "objectively less readable." 

`size() == 0` is equal to `isEmpty()`

but for this case

`size() > 0` is is equal to `!isEmpty()` where one easily can miss the single `!` at the start ;-)

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -435,49 +416,22 @@ private boolean isFragment(Bundle current) {
 		return (revision != null) && ((revision.getTypes() & BundleRevision.TYPE_FRAGMENT) != 0);
 	}
 
-	private void refreshPackages(Bundle[] bundles, BundleContext context) {
-		if (bundles.length == 0 || packageAdminService == null)
+	private void refreshPackages(Collection<Bundle> bundles) {
+		if (bundles.isEmpty())
 			return;
```
HannesWell, line 422: If you change this, you could also add brace.
```suggestion
		if (bundles.isEmpty()) {
			return;
		}
```

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -435,49 +416,22 @@ private boolean isFragment(Bundle current) {
...
-				}
-			}
-		}
-
 		CountDownLatch latch = new CountDownLatch(1);
 		FrameworkListener listener = event -> {
 			if (event.getType() == FrameworkEvent.PACKAGES_REFRESHED) {
 				latch.countDown();
 			}
 		};
```
HannesWell, line 428: The listener could be inlined now.

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -435,49 +416,22 @@ private boolean isFragment(Bundle current) {
...
-	private void refreshPackages(Bundle[] bundles, BundleContext context) {
-		if (bundles.length == 0 || packageAdminService == null)
+	private void refreshPackages(Collection<Bundle> bundles) {
+		if (bundles.isEmpty())
 			return;
 
-		// Prior to Luna the Equinox framework would refresh all bundles with the same
-		// BSN automatically.  This is no longer the case for Luna or other framework
-		// implementations.  Here we want to make sure all existing bundles with the
-		// same BSN are refreshed also.
```
HannesWell, line 445: Is this not relevant anymore?

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
HannesWell: > Actually the check was not required (anymore) anyways so I removed it now completely.

Indeed. The code-block where getAdditionalRefresh() is called is only called if `toRefresh` is not empty.

> Beside that, I'm not such a great fan of such "tools" that try to be smart for rare cases and mirco optimizations, I think the only real world example is a [ConcurrentSkipListSet](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentSkipListSet.html) where checks for empty/size are meaningless anyways when we assume a high concurrent read/write system...

While I agree that the negation in `!isEmpty()` is not ideal and negations in general tend to make boolean expressions more difficult I would also argue that that writing `not empty` is more meaningful and more like one thinks about the algorithm (at least for me) than `size greater/not equal zero`.
About this specific case I can only tell that I was once bitten by exactly that problem with a filtering list views (like e.g. Guava provides it) where size() must traverse and test all elements in the list, while `isEmpty()` can be a short-circuit.
And in general, I personally like these 'tools' since they often give hints about how one can implement things simpler, more elegant or more performant and thus one can often learn from that and identify code that easaly that needs some rework. And it happened more than once to me that they discovered bugs in code I have just written.
But of course there are also arguable rules/suggestions made by these 'tools', but I digress.

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
laeubi: If it would be relevant and not already covered by `getAdditionalRefresh` then it should be handled there ...
But as this codepath is already marked with

`TODO consider removing this case because it can cause inconsistent results.`

I'm not very certain this is worth to worry about.

HannesWell: > If it would be relevant and not already covered by `getAdditionalRefresh` then it should be handled there 

Agree.

> ... But as this codepath is already marked with
> 
> `TODO consider removing this case because it can cause inconsistent results.`
> 
> I'm not very certain this is worth to worry about.

If in doubt I would worry about it. But if it is not necessary anymore it should be removed yes, but we should then be certain about that. :)

laeubi: I now added this back even though it seems only there for historic reasons, if one refreshes always all bundles with same BSN then maybe one should also refresh all providers of a package... in the end it then seems better to probably refresh the whole state always as indicated by the comment....

laeubi:
@tjwatson I also now added a flag so one do not needs to wait fro the refresh to finish or is this mandatory here?

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -19,6 +19,7 @@
 import java.net.*;
 import java.util.*;
 import java.util.concurrent.CountDownLatch;
+import java.util.stream.Collectors;
 import org.eclipse.equinox.internal.simpleconfigurator.utils.*;
 import org.osgi.framework.*;
 import org.osgi.framework.namespace.*;
```
tjwatson, line 25: Why is PackageAdmin still imported?

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -463,21 +439,12 @@ private void refreshPackages(Bundle[] bundles, BundleContext context) {
 				latch.countDown();
 			}
 		};
-		context.addFrameworkListener(listener);
-		packageAdminService.refreshPackages(allSameBSNs.toArray(new Bundle[0]));
-
+		frameworkWiring.refreshBundles(bundles, listener);
 		try {
 			latch.await();
```
tjwatson, line 444: How is this not blocking forever if you never add the listener.  This makes me believe there are no tests for any of these changes you make here.

tjwatson (changes requested):

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
laeubi: It still uses packagage admin on other places, but I wanted to start with one small thing here.

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
laeubi: `frameworkWiring.refreshBundles( ... )` accepts the listener as an argument here, I just copied it from the existing code you added here

 https://github.com/eclipse-equinox/equinox/commit/907ec00a829e2fe03f156089775c70ddac834d6e

or do I misunderstood you comment?

tjwatson: OK, got it.  I missed the listener being passed in.

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
HannesWell: > I now added this back even though it seems only there for historic reasons, 

Thanks. Looks better now.

> if one refreshes always all bundles with same BSN then maybe one should also refresh all providers of a package... in the end it then seems better to probably refresh the whole state always as indicated by the comment....

Yes, this could probably be improved or removed. But it should then be done completely and not only partly.

HannesWell (approved):

laeubi:
@tjwatson are your concerns addressed now?

tjwatson (approved):

akurtakov:
@laeubi looks like we missed that in the previous cycle. Would you please rebase and once PR succeeds merge as it's already approved?

laeubi:
I now rebased this, as it was a non trivial rebase I would suggest that others might look at this as well.

PR #422, opened 2024-01-03
Refresh a subset of bundles if they provide missing requirements

laeubi:
When refreshing large sets of bundles the resolver sometimes take a choice where a requirement can't be bound even though it is there and resolvable.

This adds some additional code to detect this situation and then refresh a smaller set of bundles that provide these until a max retry is reached, all bundles are resolved, or all providers have been tried to refresh already.

### DIFF
--- a/bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/Activator.java
+++ b/bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/Activator.java
@@ -43,7 +43,7 @@ import org.osgi.framework.*;
  * Otherwise, no uninstallation will not be done.
  */
 public class Activator implements BundleActivator {
-	public final static boolean DEBUG = false;
+	public final static boolean DEBUG = Boolean.getBoolean("equinox.simpleconfigurator.debug");
 
 	/**
 	 * If this property is set to true, simpleconfigurator will attempt to read
--- a/bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
+++ b/bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
@@ -19,16 +19,21 @@ import java.io.*;
 import java.net.*;
 import java.util.*;
 import java.util.concurrent.CountDownLatch;
+import java.util.stream.Stream;
 import org.eclipse.equinox.internal.simpleconfigurator.utils.*;
+import org.eclipse.osgi.report.resolution.ResolutionReport.Entry.Type;
 import org.osgi.framework.*;
+import org.osgi.framework.hooks.resolver.ResolverHookFactory;
 import org.osgi.framework.namespace.*;
 import org.osgi.framework.startlevel.BundleStartLevel;
 import org.osgi.framework.wiring.*;
 import org.osgi.resource.Namespace;
 import org.osgi.resource.Requirement;
 import org.osgi.service.packageadmin.PackageAdmin;
+import org.osgi.service.resolver.ResolutionException;
 
 class ConfigApplier {
+
 	private static final String LAST_BUNDLES_INFO = "last.bundles.info"; //$NON-NLS-1$
 	private static final String PROP_DEVMODE = "osgi.dev"; //$NON-NLS-1$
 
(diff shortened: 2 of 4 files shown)


bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -115,11 +120,91 @@ void install(URL url, boolean exclusiveMode) throws IOException {
...
+			}
+			// when refreshing large sets of bundles the resolver sometimes take a choice
+			// where a requirement can't be bound even though it is there and resolvable
+			// this code collects all those requirements and refresh the smaller set of
+			// bundles that provide these until a max retry is reached, all bundles are
+			// resolved, or all providers have been tried to refresh already
+			Set<Bundle> bundlesTried = new HashSet<>();
+			Collection<Bundle> provider;
+			Set<Bundle> doNotRefresh = getDoNotRefresh();
+			int maxtry = MAX_REFRESH_TRY;
```
tjwatson, line 132: Can you explain why this would need to be tried more than once?  If it didn't resolve after the first attempt to refresh providers, why would it work on subsequent retries?

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
```diff
@@ -115,11 +120,91 @@ void install(URL url, boolean exclusiveMode) throws IOException {
 					if (additionalRefresh.length > 0)
 						refreshPackages(additionalRefresh, manipulatingContext);
 				}
+			}
+			// when refreshing large sets of bundles the resolver sometimes take a choice
```
tjwatson, line 124: Do you have a scenario that exhibits this behavior?  Is there going to be a way to add a test for this?

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/Activator.java
```diff
@@ -43,7 +43,7 @@
  * Otherwise, no uninstallation will not be done.
  */
 public class Activator implements BundleActivator {
-	public final static boolean DEBUG = false;
+	public final static boolean DEBUG = Boolean.getBoolean("equinox.simpleconfigurator.debug");
```
tjwatson, line 46: I'm unsure why this doesn't use the `org.eclipse.osgi.service.debug.DebugOptions` service to get trace options for the simple configurator.  Probably because of the ill-fated dream to have p2 be able to run on other framework implementations.

tjwatson (commented):
The code is fine given what it is trying to do.  But I question if we can have a testcase that exhibits the behavior so we can add a test for this code path.

laeubi:
> The code is fine given what it is trying to do. But I question if we can have a testcase that exhibits the behavior so we can add a test for this code path.

I'm not yet sure in wich form such a test-case would have to be provided, can I somehow "export" a bundle-set (metadata-only) and use it as an input for a test with simple-configurator?

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/Activator.java
laeubi: I could think about that tracing might not be available at this stage? But not completely sure, that way you can at least enable it quite easy when starting Eclipse from commandline.

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/ConfigApplier.java
laeubi: Yes I have a product where I install a larger set of plugins into an older Eclipse version, but I'm not sure how this cane be transformed into a standalone test-case, do we have such facility somewhere?

I could think about exporting all bundles into a OSGi Repository and then call the resolver api with this repository as a source...

bundles/org.eclipse.equinox.simpleconfigurator/src/org/eclipse/equinox/internal/simpleconfigurator/Activator.java
tjwatson: We can leave that to another issue

tjwatson:
@laeubi you may not like this, but to move this forward I would like to have an option that defaults to the current behavior so we can enable this for the problematic scenarios.  I would rather do that now so that the ones impacted by this issue can try out the solution without risking others.  Once we get more confident in the approach then we can flip the default or get rid of the option.

laeubi:
@tjwatson do you like to suggest how this should be implemented? Maybe as simple as a system property to check?

tjwatson:
> @tjwatson do you like to suggest how this should be implemented? Maybe as simple as a system property to check?

Yes, but use `BundleContext.getProperty` instead of `System.getProperty`.

laeubi:
@tjwatson I now added this inside a block that is disabled by default and can be enabled with `equinox.simpleconfigurator.deeprefresh=true`

tjwatson (approved):
LGTM,

Thanks for putting the option to enable here.  As developers we should encourage committers and contributors to test with this enabled so we can evaluate enabling this by default in the future.

PR #428, opened 2024-01-14
Fix NPE in AbstractArtifactRepository#hashCode

laeubi:
if the location is null then the AbstractArtifactRepository#hashCode throws a NPE because it effectively uses the string representation of the URI.


akurtakov:
Thanks.

PR #429, opened 2024-01-16
Remove @noreference from IProvidedCapability

laeubi:
I think we can assumes these methods are "full API"


akurtakov:
I agree.

merks (approved):
Better late than never.

PR #431, opened 2024-01-19
Support properties supplied by an agent

laeubi:
Currently P2 can be controlled by a set of properties that mostly delegate to the Bundle Properties (what effectively is backed by system properties). In certain situations it would be more suitable that the IProvisioningAgent can control these (e.g. max threads, mirrors used or not, checksums enabled, ...).

This adds new (default implemented) methods to IProvisioningAgent to query it for a property where the default simply uses the Bundle properties if available or system properties in all other cases.


merks (approved):
Looks okay.

PR #434, opened 2024-01-22
remove Exceptions not actually thrown

jukzi:
eliminates problem markers and dead code


jukzi:
review/ submit any committer, please?

merks:
@jukzi  Such removals are only acceptable **assuming** no one has overridden the method, right?

jukzi:
> @jukzi Such removals are only acceptable **assuming** no one has overridden the method, right?

yes. if someone in my workspace had overriden it it would have been resulted in an error. As private packages they should not be used outside.

jukzi:
thanks, ed

PR #435, opened 2024-01-22
use try-with-resource where hinted

jukzi:


jukzi:
review/ submit any committer, please?

merks:
Thanks. 👍

PR #438, opened 2024-01-24
fix "hiding" warnings

jukzi:


jukzi:
@merks mind to review/submit?
compared to master removes dozens of warnings:
image

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/viewers/DeferredQueryContentProvider.java
```diff
@@ -34,7 +33,7 @@ public class DeferredQueryContentProvider extends ProvElementContentProvider {
 	Object currentInput;
 	HashMap<Object, Object> alreadyQueried = new HashMap<>();
 	HashSet<Object> queryCompleted = new HashSet<>();
-	AbstractTreeViewer viewer = null;
+	AbstractTreeViewer viewer2 = null;
```
merks, line 36: This is unused now.  I really didn't understand the point of the field before.  It was only written but never read.  Better to remove it.

merks: Other than this nit, it looks fine.  (So ugly here too that the things are package protected and not private, but I digress.)

It's also hard to review changes when there is so much formatting mixed in.  But that's also not your fault!

jukzi: done

merks:
@jukzi  Oh dear, a conflict.  Sorry.  😢

jukzi:
> @jukzi Oh dear, a conflict. Sorry. 😢

np, resolved

PR #439, opened 2024-01-25
Migrate p2 engine to agent properties

laeubi:
Now that agent properties are possible we should use them in the engine, to not rely on only static OSGi context properties.


merks (changes requested):
It seems okay.  But it causes one deprecation warning in Oomph for the phase set factory changes.  I can't actually get rid of it by using the new method because I need to be able to run in older IDEs.  I can't safely ignore the warning either because it's marked for removal.  I could use reflection, but that is also ugly and error prone.  So mostly I'm stuck with an ugly problem for which there are only ugly solutions.

There isn't a single iota of documentation to explain why I **must** use the new improved method.  In fact, the deprecation doesn't  even document which method should be used instead, nor even an indication when the removal will happen (no since).  There's nothing the in Javadoc  to suggest the agent could be null and why the agent is needed at all.  If there is going to be disruptive change, the documentation should be better.

laeubi:
@merks I was now able to rework this even without any deprecations!

merks (approved):
This looks less disruptive.

PR #446, opened 2024-02-04
Add optional, greedy and disabled by default requirement to the source

laeubi:
Currently it is quite cumbersome effort to include the source of a bundle with an installation even though there might be multiple demands, for example legal reasons or debugging.

This now adds a new greedy, optional requirement that is disabled by a filter so it is not effective by default but enables an install agent to specify a property of `org.eclipse.equinox.p2.includesource=true` to include sources of bundles in products whenever available.


merks (approved):
Looks very good and useful.

HannesWell:
Sounds also reasonable for me as well.
Would it make sense to have a constant for the filter so that code enabling it can reference the constant?

laeubi:
I now tested this and in general it works with Tycho but just requires some additional configuration at the moment, I also take the opportunity to fix javadoc errors and extract some constants here, so now this should be ready to submit.

PR #447, opened 2024-02-04
Adjust maven metadata to the ones actually published to central

laeubi:
Currently the project is configured to use the org.eclipse.equinox.p2 group id but it is actually published to org.eclipse.platform. That is both confusing and makes the metadata in p2 pointless and not allows to use local build artifacts (e.g. for Tycho).

This changes the group id and configures the update-consumer-pom to create a pom that can be used in a local maven setup similar to one published on maven central.

This is what is generated by Tycho for example:
```
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.eclipse.platform</groupId>
  <artifactId>org.eclipse.equinox.p2.publisher.eclipse</artifactId>
  <version>1.6.0-SNAPSHOT</version>
  <name>[bundle] Equinox Provisioning Publisher for Eclipse</name>
  <organization>
    <name>Eclipse.org - Equinox</name>
  </organization>
  <dependencies>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.p2.metadata.repository</artifactId>
      <version>1.5.300-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.simpleconfigurator.manipulator</artifactId>
      <version>2.3.100-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.frameworkadmin</artifactId>
      <version>2.3.100-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.p2.metadata</artifactId>
      <version>2.9.0-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.p2.artifact.repository</artifactId>
      <version>1.5.300-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.frameworkadmin.equinox</artifactId>
      <version>1.3.100-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.p2.core</artifactId>
      <version>2.11.0-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.p2.repository</artifactId>
      <version>2.8.100-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.p2.publisher</artifactId>
      <version>1.9.100-SNAPSHOT</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.bouncycastle</groupId>
      <artifactId>bcpg-jdk18on</artifactId>
      <version>1.77</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.bouncycastle</groupId>
      <artifactId>bcprov-jdk18on</artifactId>
      <version>1.77</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.tukaani</groupId>
      <artifactId>xz</artifactId>
      <version>1.9</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.p2.jarprocessor</artifactId>
      <version>1.3.300-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.osgi</groupId>
      <artifactId>org.osgi.service.prefs</artifactId>
      <version>1.1.2</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.simpleconfigurator</artifactId>
      <version>1.5.100-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
</project>
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@ bin/
 .META-INF_*
 pom.tycho
 .tycho-consumer-pom.xml
+.flattened-pom.xml
 apiAnalyzer-workspace/
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,2 +1,3 @@
 -Pbuild-individual-bundles
 -Dtycho.surefire.deleteWorkDir=true
+-Dtycho.localArtifacts=ignore
--- a/bundles/org.eclipse.equinox.p2.tests/pom.xml
+++ b/bundles/org.eclipse.equinox.p2.tests/pom.xml
@@ -8,7 +8,7 @@
 	<modelVersion>4.0.0</modelVersion>
 
 	<parent>
-		<groupId>org.eclipse.equinox.p2</groupId>
+		<groupId>org.eclipse.platform</groupId>
 		<artifactId>rt.equinox.p2</artifactId>
 		<version>${releaseVersion}${qualifier}</version>
 		<relativePath>../../</relativePath>
@@ -39,7 +39,7 @@
 	<dependencies>
 		<!-- The tests require the reconciler product -->
 		<dependency>
-			<groupId>org.eclipse.equinox.p2</groupId>
+			<groupId>${project.groupId}</groupId>
 			<artifactId>org.eclipse.equinox.p2.reconciler</artifactId>
 			<version>1.1.0-SNAPSHOT</version>
 			<type>pom</type>
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -2,7 +2,7 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.eclipse.equinox.p2</groupId>
+    <groupId>org.eclipse.platform</groupId>
     <artifactId>rt.equinox.p2</artifactId>
     <version>${releaseVersion}${qualifier}</version>
     <relativePath>../pom.xml</relativePath>
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <relativePath>../eclipse-platform-parent</relativePath>
   </parent>
 
-  <groupId>org.eclipse.equinox.p2</groupId>
+  <groupId>org.eclipse.platform</groupId>
   <artifactId>rt.equinox.p2</artifactId>
   <packaging>pom</packaging>
   <version>${releaseVersion}${qualifier}</version>
@@ -125,6 +125,18 @@
             <goals>
               <goal>p2-metadata</goal>
             </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-packaging-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>default-update-consumer-pom</id>
+            <configuration>
+				<skipPomGeneration>false</skipPomGeneration>
+			</configuration>
           </execution>
         </executions>
       </plugin>


mickaelistria:
@merks would the Maven publisher script require some adaptation with this change?

merks:
@mickaelistria 

Thanks for asking!  I expect this rule for mapping the BSN will kick in regardless of this PR's changes:

image

mickaelistria:
Thanks @laeubi for this change (I'm really happy to see more and more consistency here, ultimately, those are small steps towards just being able to do `mvn deploy` one day).
And thanks @merks for the check.

laeubi:
Just note that this metadata is just transient (means you only see it in your local repository!) it is not embedded in the jar or used anywhere else so I think @merks is correct that it has no influence at all.

merks:
@laeubi 

That's not entirely true.  Tycho records that information in the published repository metadata as `maven-*` properties:
```
<unit id='org.eclipse.equinox.p2.artifact.repository' version='1.5.300.v20240201-0843' generation='2'>
  <update id='org.eclipse.equinox.p2.artifact.repository' range='[0.0.0,1.5.300.v20240201-0843)' severity='0'/>
  <properties size='9'>
    <property name='df_LT.pluginName' value='Equinox Provisioning Artifact Repository Support'/>
    <property name='df_LT.providerName' value='Eclipse.org - Equinox'/>
    <property name='org.eclipse.equinox.p2.name' value='%pluginName'/>
    <property name='org.eclipse.equinox.p2.provider' value='%providerName'/>
    <property name='org.eclipse.equinox.p2.bundle.localization' value='plugin'/>
    <property name='maven-groupId' value='org.eclipse.equinox.p2'/>
    <property name='maven-artifactId' value='org.eclipse.equinox.p2.artifact.repository'/>
    <property name='maven-version' value='1.5.300-SNAPSHOT'/>
    <property name='maven-type' value='eclipse-plugin'/>
  </properties>
```

The mapping rules can even use that information, which it does as a last resort for bundle mappings:

image

Nevertheless, the conclusion remains correct.

laeubi:
@merks thats correct, I assume its not bas that the information there is now a little bit more correct than before (SNAPSHOT on release is still wrong)?

mickaelistria:
> Thats correct, I assume its not bad that the information there is now a little bit more correct than before

Yes.

> (SNAPSHOT on release is still wrong)?

They're the ones from Tycho to build a qualifier; the Maven publisher publishes without snapshots but we don't know it at the time we generate p2 metadata.
If you want to start getting rid of it, one solution can be to allow Tycho to append a qualifier to a release version, so the bundle could have Maven GAV x.y.z and p2 metadata x.y.z.qualifier and those would work together. Or another approach is to use the expanded qualifier instead of -SNAPSHOT in generated p2 metadata. but both approaches would also sometimes lead to incorrect result if we decide to publish Maven artifacts as snapshots (but actually instead of snapshots, and want to use those -SNAPSHOTs with the usual assumption they're the latest decent build).
The thing is that we usually "promote a release" without rebuilding it while Maven doesn't really understand that. It's probably a more philosophical question: Maven most likely assumes that was is released is the sources and builds binaries from it; while we in Eclipse SDK land tend to assume that we release binaries first (and tag the source when we are happy with the binaries).

merks:
I think everything is fine and good just the way it is now with no need for changes.  👍  It is an accurate reflection of exactly what is known in the source at the time of production.  (And generally folks don't publish to maven by replacing the SNAPSHOT with something else but rather strip it completely.)

merks:
Don't change anything.   🙈  There are bigger and better fish to fry.  😁

PR #452, opened 2024-02-07
Remove o.e.equinox.p2.releng

akurtakov:
Outdated, not working and not useful. It has been broken 4.25 to 4.28, now it stayed to 4.29 and hasn't been noticed in almost half a year. Aka this is more of a burden than help.


akurtakov:
OOM in Jenkins but I'll still merge as it gets less diffs in Tycho versions and one less useless project lurking around

PR #454, opened 2024-02-08
Get rid of p2 tests reconciler product

akurtakov:
Configuring tycho surefire with all the deps needed should be achievable via other less expensive means.


laeubi:
@akurtakov see
- https://github.com/eclipse-equinox/p2/pull/458

PR #455, opened 2024-02-08
Embed the required StateFactory implementations

laeubi:
Currently P2 has the BundleDescription in its API where the only implementation is in the discouraged osgi.compatibility fragment, but actually P2 does not really need a full implementation, it is just using that to parse manifest headers and generate metadata from it.

This embeds a stripped down version of the actual implementation in the p2.publisher.eclipse that allows to create implementation objects without rely on the compatibility fragment anymore.

This is currently work in progress and a way to track progress, the classes can most probably stripped down even more and currently depends on some equinox framework internals that probably should either be inline as well or an alternative should be chosen.


laeubi:
Beside the zip packaging problem this seem to not break anything @merks do you want to give it a try with Ooomph?

@tjwatson what do you think? I know it is not nice to copy the code, but it seems unlikely P2 will ever be able to move away from the API interface and at least we can strip down everything to the bare minimum P2 is actually using here (e.g. it never uses a `State` anyways and would make PDE the only consumer of the compatibility fragment.

merks:
It appears to cause no breakage for Oomph...

laeubi:
One option would be if we move the fragment to P2 instead of _copy_ one can simply *link* the sources here, see

- https://github.com/eclipse-equinox/equinox/discussions/510

bundles/org.eclipse.equinox.p2.publisher.eclipse/META-INF/MANIFEST.MF
```diff
@@ -33,12 +33,15 @@ Import-Package: org.eclipse.equinox.app;version="[1.0.0,2.0.0)",
  org.eclipse.equinox.simpleconfigurator.manipulator;version="[2.0.0,3.0.0)",
  org.eclipse.equinox.spi.p2.publisher,
  org.eclipse.osgi.framework.util,
+ org.eclipse.osgi.internal.framework,
```
tjwatson, line 36: Importing of internal packages should not be done for this.

tjwatson:
I'm not convinced making a copy here helps much with anything vs. using the APIs and implementation fragment as they are today.

laeubi:
> I'm not convinced making a copy here helps much with anything vs. using the APIs and implementation fragment as they are today.

The API will still be used (as we can't change this much) but the implementation (compatibility fragment) will not be required anymore as P2 don't really wants the "real" implementation, in fact is only uses that to parse the MANIFEST and provides that as a structured way to access the information.

So one can also reimplement that, but copy whats already there seems to be a faster approach.

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/statefactory/StateBuilder.java
```diff
@@ -0,0 +1,923 @@
...
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *     Danail Nachev -  ProSyst - bug 218625
+ *******************************************************************************/
+package org.eclipse.equinox.p2.publisher.eclipse.statefactory;
+
+import java.lang.reflect.Constructor;
+import java.util.*;
+import org.eclipse.osgi.internal.framework.EquinoxContainer;
```
tjwatson, line 19: Should not import this internal for the constant.  And you seem to have an `EquinoxContainer` class in this `statefactory` you can use.

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/statefactory/StateBuilder.java
```diff
@@ -0,0 +1,923 @@
...
+import org.eclipse.osgi.internal.framework.FilterImpl;
+import org.eclipse.osgi.internal.messages.Msg;
```
tjwatson, line 21: Should not use the frameworks internal NLS classs `Msg`

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/statefactory/GenericSpecificationImpl.java
```diff
@@ -0,0 +1,212 @@
...
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *     Rob Harrop - SpringSource Inc. (bug 247522)
+ *******************************************************************************/
+package org.eclipse.equinox.p2.publisher.eclipse.statefactory;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.eclipse.osgi.internal.framework.FilterImpl;
```
tjwatson, line 20: Should use `FrameworkUtil.createFilter` instead of using the internal `FilterImpl` of the framework.

tjwatson (changes requested):

bundles/org.eclipse.equinox.p2.publisher.eclipse/META-INF/MANIFEST.MF
laeubi: There are two places that are quite hard to replace, that is native code mappings and where special operations are performed on a filter... I'm checking if we can get rid of it.

laeubi:
@merks @tjwatson if no further concerns I think we should merge this to get some more confidence everything works and possibly fix things if needed.

merks:
FYI, I did check that this causes no compilation problem for Oomph so I think it's fine.  I did not test behavior...

HannesWell:
> > I'm not convinced making a copy here helps much with anything vs. using the APIs and implementation fragment as they are today.
> 
> The API will still be used (as we can't change this much) but the implementation (compatibility fragment) will not be required anymore as P2 don't really wants the "real" implementation, in fact is only uses that to parse the MANIFEST and provides that as a structured way to access the information.
> 
Can't we migrate P2 to use the code from in the 'new' Equinox container, i.e. away from the old equinox resolver APIs?

tjwatson (approved):
LGTM

laeubi:
> Can't we migrate P2 to use the code from in the 'new' Equinox container, i.e. away from the old equinox resolver APIs?

Of course we can migrate to "something else" but I don't really see much benefit in that:

1. We would break the API here so special consideration is required to allow consumers to move away
2. The new API must be doing the same as the current API (or very similar) without any immediate benefit

So given PDE has moved away from this package we can decide to completely embed it into P2 instead of invent a new API and then we can even remove more things.

PR #458, opened 2024-02-10
Remove the reconciler product

laeubi:
This removes the only for tests build reconcile product project, instead it is now part of the test bundle and used to create a p2 provisioned install.

The product is still needed because P2 here copies a real product and then performs operations on it, but at laest we save us the packaging of all environments, only the current running environment is ever used here.


akurtakov (approved):

akurtakov:
Thanks a lot!

Issue #461, opened 2024-02-12
UpdateSitePublisher -addJREIU argument doesn't add current JVM


iloveeclipse:
One of the issues found while investigating https://github.com/eclipse-equinox/p2/issues/450.

`eclipse -application org.eclipse.equinox.p2.publisher.UpdateSitePublisher  -addJREIU ...` doesn't add JRE "IU" for the currently used JRE but to the one set in `JREAction.DEFAULT_PROFILE` (currently Java 17).

So I'm unable to add "a.jre.javase" with Java 21 to the published update site content.xml and so can't install bundles that require Java 21 using plain SDK, even if everything runs on Java 21.

iloveeclipse:
Verified works with I20240212-1800. I can generate "a.jre.javase" 21 IU now via UpdateSitePublisher application.

PR #462, opened 2024-02-12
UpdateSitePublisher -addJREIU argument doesn't add current JVM

iloveeclipse:
Don't hard code "default" JRE version if we "guess one" and create some profile based on current JRE packages. 
Just read & use the version from the running JRE.

Fixes https://github.com/eclipse-equinox/p2/issues/461

### DIFF
--- a/bundles/org.eclipse.equinox.p2.publisher/src/org/eclipse/equinox/p2/publisher/actions/JREAction.java
+++ b/bundles/org.eclipse.equinox.p2.publisher/src/org/eclipse/equinox/p2/publisher/actions/JREAction.java
@@ -40,9 +40,12 @@ import org.eclipse.osgi.util.NLS;
 import org.osgi.framework.*;
 
 public class JREAction extends AbstractPublisherAction {
+
+	private static final int RUNTIME_VERSION = Runtime.version().feature();
+
 	private static final String DEFAULT_JRE_NAME = "a.jre"; //$NON-NLS-1$
-	private static final Version DEFAULT_JRE_VERSION = Version.parseVersion("17.0"); //$NON-NLS-1$
-	private static final String DEFAULT_PROFILE = "JavaSE-17"; //$NON-NLS-1$
+	private static final Version DEFAULT_JRE_VERSION = Version.parseVersion(RUNTIME_VERSION + ".0"); //$NON-NLS-1$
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Build fails with `org.eclipse.equinox.p2.tests: Error assembling JAR: Problem creating jar: Execution exception: Java heap space`.
is this something that changed recently?

laeubi:
Please add https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.mvn/jvm.config to this repository as well, it should help with that OOM errors as currently only 1GB of 4G is used by default.

iloveeclipse:
> Please add https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.mvn/jvm.config to this repository as well, it should help with that OOM errors as currently only 1GB of 4G is used by default.

We surely won't have that workaround be forever, is there a ticket for that so I can link the PR against?

laeubi:
> > Please add https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.mvn/jvm.config to this repository as well, it should help with that OOM errors as currently only 1GB of 4G is used by default.
> 
> We surely won't have that workaround be forever, is there a ticket for that so I can link the PR against?

If you mean we should stop development once we hit 1gb of memory demands then no there is no such ticket, beside that we don't get an refund if we only use 1/4 of available memory.

iloveeclipse:
> If you mean we should stop development once we hit 1gb of memory demands then no there is no such ticket, beside that we don't get an refund if we only use 1/4 of available memory.

I'm only asking for the root issue here, that requires now this workaround.

laeubi:
> > If you mean we should stop development once we hit 1gb of memory demands then no there is no such ticket, beside that we don't get an refund if we only use 1/4 of available memory.
> 
> I'm only asking for the root issue here, that requires now this workaround.

I have recently analyzed the build and didn't find any unusual memory consumption, so the root cause seem that we are use more features, doing more checks and so on takes more memory so we hit a limit. As long as we are not really hitting the limit (e.g. need more and more and finally getting killed) I won't analyze further. So if 2g works for the next 20years its a bit of waste of dev resources otherwise.

If you are interested, you can run the build local with -Xmx=1g (the current default) and let you produce a heapdump and look for possible optimizations in platform/jdt(I already made some) or tycho/maven (e.g [MNG-7592](https://issues.apache.org/jira/projects/MNG/issues/MNG-7592).

Issue #463, opened 2024-02-12
Jenkins builds fail with Java heap space


iloveeclipse:
See 
- https://github.com/eclipse-equinox/p2/pull/462 
- https://ci.eclipse.org/equinox/job/p2/job/PR-462/1/console

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:4.0.6-SNAPSHOT:package-plugin (default-package-plugin) on project org.eclipse.equinox.p2.tests: Error assembling JAR: Problem creating jar: Execution exception: Java heap space -> [Help 1]
[ERROR] 
```

PR #464, opened 2024-02-12
Workaround for Jenkins builds failing with "Java heap space" errors

iloveeclipse:
Fixes https://github.com/eclipse-equinox/p2/issues/463

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,0 +1,2 @@
+-XX:+PrintFlagsFinal
+-Xmx2g


iloveeclipse:
@laeubi : no idea what is wrong, but maven seem to read JVM argument as maven option:

```
Error:  Error executing Maven.
org.apache.commons.cli.ParseException: Unrecognized maven.config file entries: [mx2g]
```

mickaelistria:
I think you mean Jenkins, not Gerrit.

iloveeclipse:
> I think you mean Jenkins, not Gerrit.

Doesn't work either.

akurtakov:
"Please add https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.mvn/jvm.config to this repository as well, it should help with that OOM errors as currently only 1GB of 4G is used by default." but you put it in maven.config. It's well documented and explained in https://maven.apache.org/configure.html .

iloveeclipse:
> "Please add https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.mvn/jvm.config to this repository as well, it should help with that OOM errors as currently only 1GB of 4G is used by default." but you put it in maven.config. It's well documented and explained in https://maven.apache.org/configure.html .

I'm allowed to make mistakes, OK?

akurtakov:
> > "Please add https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.mvn/jvm.config to this repository as well, it should help with that OOM errors as currently only 1GB of 4G is used by default." but you put it in maven.config. It's well documented and explained in https://maven.apache.org/configure.html .
> 
> I'm allowed to make mistakes, OK?

Sure, that's why I point you to the detailed documentation for reference in the future.

PR #469, opened 2024-02-13
Jenkinsfile: enable eclipse qualityGate

jukzi:
error was:
"Repository miner is not configured, skipping repository mining Reference build recorder is not configured
Obtaining reference build from same job"


jukzi:
finds master now:
image

PR #470, opened 2024-02-14
Suppress warnings about deprecated Preferences in legacy Code

jukzi:


laeubi (changes requested):
Instead of suppress the warning one should consider to replace the deprecated code.

Issue #252, opened 2023-04-18
java.lang.UnsupportedOperationException in org.eclipse.equinox.internal.p2.extensionlocation.ExtensionLocationArtifactRepository


jukzi:
```
java.lang.UnsupportedOperationException
	at org.eclipse.equinox.internal.p2.extensionlocation.ExtensionLocationArtifactRepository.removeDescriptor(ExtensionLocationArtifactRepository.java:191)
	at org.eclipse.equinox.internal.p2.artifact.repository.MirrorRequest.perform(MirrorRequest.java:179)
	at org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository.getArtifact(SimpleArtifactRepository.java:773)
	at org.eclipse.equinox.internal.p2.artifact.repository.simple.DownloadJob.run(DownloadJob.java:64)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

```
please see original report: https://github.com/eclipse-platform/.github/issues/117

PR #471, opened 2024-02-15
MirrorRequest: catch UnsupportedOperationException #252

jukzi:
and resolve deprecated call

https://github.com/eclipse-equinox/p2/issues/252


laeubi:
Please note that this does not *ignores* the exception as suggested by @merks here:

https://github.com/eclipse-platform/.github/issues/117#issuecomment-1512681071

so it might lead to a download never happens even though we just can't remove the descriptor (for whatever reasons).

jukzi:
then would you mind to fix it accordingly?

laeubi:
See:
- https://github.com/eclipse-equinox/p2/pull/472

PR #472, opened 2024-02-18
Ignore the UnsupportedOperationException in case of descriptor removal

laeubi:
If removal of the errornous descriptor failed this can have different reasons:

1) it was never added because the target is not capable of downloading 2) the repository does not support removal of descriptors 3) some bug in the implementation

in any case we want to report the original error an not an error that says we can't remove the descriptor.


merks (approved):
It seems adding more logic is the only way to track down why we are removing a descriptor from an implementation that does not support removal at all.

PR #474, opened 2024-02-20
Correctly consider cardinality based on the directives

laeubi:
Currently the BundlesAction uses either 0 or 1 as the cardinality but actually a requirement can be a multi-cardinality as well.

This adds new dedicated methods to not duplicate the computation and a testcase that ensures all combination are covered and correctly translated.


laeubi:
`cardinality:=multiple` is not very common but we still should map this correctly, @tjwatson @merks should this go into RC1 or should we wait for the next release?

merks:
Can you summarize the impact.  I believe it just means that if there is a cardinality:=multiple the requirement will have max=MAX_INT.  Otherwise nothing is changed, correct?

laeubi:
Yes currently it is either `0..1` or `1..1`, but it is also possible that it is `0...n` and `1...n` if you look at the manifest I used for the testcase it should be more clear.

laeubi:
As no one raised concerns I'll merge this as soon as build completes.

PR #476, opened 2024-02-27
Nullpointer in AbstractRepositoryManager#handleRemoteIndexFile

laeubi:
I found that in some rare cases the handleRemoteIndexFile could run into a NPE if the Transport service can't be fetched.

This now checks first for null before using the transport and if it is null fall back to simple URL stream, if that also fails, create an empty index file.

@merks  I don't know how this could happen but seen this in a log while performing an update operation, but the method itself also seems merely used (only found two references), so instead of running into a NPE this seems better and hopefully then reveals the real root cause later on.

The stacktrace is:

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.equinox.internal.p2.repository.Transport.download(java.net.URI, java.io.OutputStream, org.eclipse.core.runtime.IProgressMonitor)" because the return value of "org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.getTransport()" is null
        at org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.handleRemoteIndexFile(AbstractRepositoryManager.java:742)
        at org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.loadIndexFile(AbstractRepositoryManager.java:736)
        at org.eclipse.equinox.internal.p2.repository.helpers.AbstractRepositoryManager.loadRepository(AbstractRepositoryManager.java:678)
        at org.eclipse.equinox.internal.p2.metadata.repository.MetadataRepositoryManager.loadRepository(MetadataRepositoryManager.java:110)
        at org.eclipse.equinox.internal.p2.metadata.repository.MetadataRepositoryManager.loadRepository(MetadataRepositoryManager.java:105)
        at org.eclipse.equinox.p2.ui.LoadMetadataRepositoryJob.doLoad(LoadMetadataRepositoryJob.java:126)
        at org.eclipse.equinox.p2.ui.LoadMetadataRepositoryJob.runModal(LoadMetadataRepositoryJob.java:110)
```

### DIFF
--- a/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/AbstractRepositoryManager.java
+++ b/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/AbstractRepositoryManager.java
@@ -739,12 +739,21 @@ public abstract class AbstractRepositoryManager<T> implements IRepositoryManager
 	private LocationProperties handleRemoteIndexFile(URI indexFileURI, IProgressMonitor monitor) {
 		ByteArrayOutputStream index = new ByteArrayOutputStream();
 		IStatus indexFileStatus = null;
-		indexFileStatus = getTransport().download(indexFileURI, index, monitor);
-		while (indexFileStatus.getCode() == IArtifactRepository.CODE_RETRY) {
-			indexFileStatus = getTransport().download(indexFileURI, index, monitor);
+		Transport transport = getTransport();
+		if (transport != null) {
+			indexFileStatus = transport.download(indexFileURI, index, monitor);
+			while (indexFileStatus.getCode() == IArtifactRepository.CODE_RETRY) {
+				indexFileStatus = transport.download(indexFileURI, index, monitor);
+			}
+			if (indexFileStatus != null && indexFileStatus.isOK())
+				return LocationProperties.create(new ByteArrayInputStream(index.toByteArray()));
+		} else {
+			try (InputStream openStream = indexFileURI.toURL().openStream()) {
+				return LocationProperties.create(openStream);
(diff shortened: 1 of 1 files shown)


laeubi:
It seems the issue is/was the ECF bundles where missing in my run config .... so I'm not sure if my fix still applies here but P2 could use a fallback service for Transports that simply uses URL.openStream even though this might not be very powerful its better than failing completely.

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/helpers/AbstractRepositoryManager.java
```diff
@@ -739,12 +739,21 @@ private LocationProperties loadIndexFile(URI location, IProgressMonitor monitor)
...
+		Transport transport = getTransport();
+		if (transport != null) {
+			indexFileStatus = transport.download(indexFileURI, index, monitor);
+			while (indexFileStatus.getCode() == IArtifactRepository.CODE_RETRY) {
+				indexFileStatus = transport.download(indexFileURI, index, monitor);
+			}
+			if (indexFileStatus != null && indexFileStatus.isOK())
+				return LocationProperties.create(new ByteArrayInputStream(index.toByteArray()));
+		} else {
+			try (InputStream openStream = indexFileURI.toURL().openStream()) {
```
merks, line 751: My big concern here is that this masks a severe configuration problem. A problem likely to crop up elsewhere and a problem that, if not fixed will, will disable all kinds of functionality such as proxy support.   Surely if one gets the index, something downstream will fail instead.

It would seem better to fail with some kind of message about the transport not being configured available and the like cause(s) for that.

PR #477, opened 2024-02-29
Absence of an unpack attribute must not result in BundleShapeAdvice

laeubi:
Currently the FeaturesAction assumes that the unpack attribute is always set, if that is not the case it leads to unexpected results because then unpack=true is assumed always. This is especially dangerous as PDE since 2023-12 release removes the attributes and Tycho assume they are false by default the same as PDE has done in the past when adding new items.

To mitigate this this adds an additional check to only evaluate the unpack if it is set, together with a testcase that covers all possible combinations.

I give +1 as project lead for this really late fix but I think it is urgent enough as it has unexpected side-effects for users using already released PDE that will likely be hard to discover and the previous state was to always specify this attribute so existing setup won't be affected. Waiting for one more release to fix this seems therefore not acceptable for me.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/FeaturesAction.java
+++ b/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/FeaturesAction.java
@@ -186,7 +186,7 @@ public class FeaturesAction extends AbstractPublisherAction {
 	private void createBundleShapeAdvice(Feature feature, IPublisherInfo publisherInfo) {
 		FeatureEntry entries[] = feature.getEntries();
 		for (FeatureEntry entry : entries) {
-			if (entry.isUnpack() && entry.isPlugin() && !entry.isRequires())
+			if (entry.unpackSet() && entry.isUnpack() && entry.isPlugin() && !entry.isRequires())
 				publisherInfo.addAdvice(new BundleShapeAdvice(entry.getId(), Version.parseVersion(entry.getVersion()), IBundleShapeAdvice.DIR));
 		}
 	}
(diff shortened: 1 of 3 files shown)


akurtakov:
The change looks good to me. I would like to get @merks review here too as we are really late in the game.

akurtakov (approved):

merks (approved):
That's great that the unpackSet method already existed!  This is indeed a very good and very minimal change.   We will avoid significant grief to include this in the 4.31 release

akurtakov:
For the record - PMC approved

laeubi:
I build started here: https://ci.eclipse.org/releng/job/Builds/job/I-build-4.31/162/

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/FeaturesAction.java
```diff
@@ -186,7 +186,7 @@ private void createAdviceFileAdvice(Feature feature, IPublisherInfo publisherInf
 	private void createBundleShapeAdvice(Feature feature, IPublisherInfo publisherInfo) {
 		FeatureEntry entries[] = feature.getEntries();
 		for (FeatureEntry entry : entries) {
-			if (entry.isUnpack() && entry.isPlugin() && !entry.isRequires())
+			if (entry.unpackSet() && entry.isUnpack() && entry.isPlugin() && !entry.isRequires())
```
tjwatson, line 189: It would seem there is a bug in `isUnpack` if we want the default to be false.  For a future release we should consider if this is correct for `isUnpack`:

```java
	public boolean isUnpack() {
		return (unpack != null && unpack.booleanValue());
	}
```

PR #482, opened 2024-03-15
Ensure the product *.ini is loaded and saved using the native encoding

merks:
The native launcher executable reads the product *.ini, e.g., eclipse.ini, using the native system encoding. Currently the *.ini is loaded and saved in the launched Java application using the system default encoding, but in Java 21 the default encoding is always UTF-8 so we need to be more careful to always use the native encoding regardless of the default encoding.


laeubi (approved):
Looks good.

PR #486, opened 2024-03-18
Enhance p2 director with "addJREIU" option.

raghucssit:
P2 Director Application fails to install some of the bundles which has requirements on Java-SE 21. The reason for this issue was missing java profile dependency during installation. This new option includes Jre IU as an extra installable Unit.

Fixes https://github.com/eclipse-equinox/p2/issues/484

PR #488, opened 2024-03-19
Update copyright date

merks:


merks:
This is only a test to see if the failure from https://github.com/eclipse-equinox/p2/pull/486 is reproducible in a different PR.

PR #493, opened 2024-03-28
Improve warning message with no digest found

laeubi:
Currently the message only mentions the artifact to download but not where it originates from, this makes it very hard to analyze and fix the issue (actually one needs to to run a debugger).

This enhances the message a tiny little bit to mention the artifact descriptors repository URL.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
@@ -94,7 +94,8 @@ public class RawMirrorRequest extends MirrorRequest {
 					IArtifactDescriptor.DOWNLOAD_CHECKSUM, Collections.emptySet());
 			if (steps.isEmpty()) {
 				LogHelper.log(new Status(IStatus.WARNING, Activator.ID,
-						NLS.bind(Messages.noDigestAlgorithmToVerifyDownload, artifactDescriptor.getArtifactKey())));
+						NLS.bind(Messages.noDigestAlgorithmToVerifyDownload, artifactDescriptor.getArtifactKey(),
+								artifactDescriptor.getRepository().getLocation())));
 			}
 			ProcessingStep[] stepArray = steps.toArray(new ProcessingStep[steps.size()]);
 			// TODO should probably be using createAndLink here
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/SimpleArtifactRepository.java
```diff
@@ -491,7 +491,8 @@ private OutputStream addPreSteps(ProcessingStepHandler handler, IArtifactDescrip
 		addChecksumVerifiers(descriptor, downloadChecksumSteps, skipChecksums, IArtifactDescriptor.DOWNLOAD_CHECKSUM);
 		if (downloadChecksumSteps.isEmpty() && !isLocal()) {
 			LogHelper.log(new Status(IStatus.WARNING, Activator.ID,
-					NLS.bind(Messages.noDigestAlgorithmToVerifyDownload, descriptor.getArtifactKey())));
```
merks, line 494: Notice that this call is guarded by `!isLocal()" because of course if the repository is local, this is not a download.

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/RawMirrorRequest.java
```diff
@@ -94,7 +94,8 @@ protected IStatus getArtifact(IArtifactDescriptor artifactDescriptor, OutputStre
 					IArtifactDescriptor.DOWNLOAD_CHECKSUM, Collections.emptySet());
 			if (steps.isEmpty()) {
 				LogHelper.log(new Status(IStatus.WARNING, Activator.ID,
-						NLS.bind(Messages.noDigestAlgorithmToVerifyDownload, artifactDescriptor.getArtifactKey())));
+						NLS.bind(Messages.noDigestAlgorithmToVerifyDownload, artifactDescriptor.getArtifactKey(),
```
merks, line 97: I was hoping that we could avoid calling this if the artifact is local.  See comment below...

merks (approved):
This looks fine.

We should consider if we can improve this to not log for local transfers, but this additional information should help make it clear when this is happening.

laeubi:
I think the reason is because RawMirrorRequest is always against a remote, at least while debugging it really tries to download from a remote.

PR #501, opened 2024-04-23
Unify Constants and modernize CollectAction

laeubi:
Currently CollectAction and Collect (Phase) both use a parameter 'artifactRequests' but both define them separately.

This unify the access to the constant and modernize the CollectAction to use collections instead of raw arrays.


merks (approved):
It does NOT NOT break Oomph on the CBI p2 Aggregator.  😄

PR #502, opened 2024-04-23
Improve message if artifact can not be downloaded from any repository

laeubi:
There are rare cases where one gets a message that an artifact can not be downloaded from any repository, this has two pitfalls:

1) You don't get an idea what repositories are actually contacted 2) Its hard to find out who actually requires this artifact

This now enhances an IArtifactRequest by a method to set a "context iu" why this artifact is requested. If the information is given, it searches through the available IUs and gives additional information why it is required and what repositories are queried.


laeubi:
@merks I made this now more "local" change.

merks (approved):
Just an observation.  The IU's id and version of the IU that requires the artifact typically provides identical information, as the artifact key and version.   Providing the first level of incoming requirements might beg for the next level of IU dependencies.

PR #505, opened 2024-04-24
Add a way to skip reference repositories

tivervac:
This goes hand in hand with the [PR](https://github.com/eclipse-pde/eclipse.pde/pull/1253/files) in PDE. If there's a way to achieve this without editing the public API in `IPlanner`, I would much prefer it.

Seeing how `FOLLOW_ARTIFACT_REPOSITORY_REFERENCES` was _always_ hardcoded to `true` in `ProvisioningContext`, I feel like someone wanted to implement something similar at some point, but didn't get around to it.

For versioning, I just followed the quick fixes, I hope this is correct.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.director/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.director/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.director;singleton:=true
-Bundle-Version: 2.6.400.qualifier
+Bundle-Version: 2.7.0.qualifier
 Bundle-Activator: org.eclipse.equinox.internal.p2.director.DirectorActivator
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/bundles/org.eclipse.equinox.p2.director/src/org/eclipse/equinox/internal/p2/director/SimplePlanner.java
+++ b/bundles/org.eclipse.equinox.p2.director/src/org/eclipse/equinox/internal/p2/director/SimplePlanner.java
@@ -480,7 +480,7 @@ public class SimplePlanner implements IPlanner {
 
 	@Override
 	public IProvisioningPlan getProvisioningPlan(IProfileChangeRequest request, ProvisioningContext context,
-			IProgressMonitor monitor) {
+			boolean includeReferences, IProgressMonitor monitor) {
 		ProfileChangeRequest pcr = (ProfileChangeRequest) request;
 		SubMonitor sub = SubMonitor.convert(monitor, ExpandWork);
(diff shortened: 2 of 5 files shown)


merks:
I’ll look in detail tomorrow. I just wonder how p2 director already supports the -followReferences option (off by default) with the current implementation. 

https://help.eclipse.org/latest/index.jsp?topic=/org.eclipse.platform.doc.isv/guide/p2_director.html

I also recall when oomph’s p2 director task didn’t follow references and I had change something to enable that.  That’s a few years back…

laeubi:
@tivervac the value is hardcoded in the constructor *but* code creating the provisioning context can still override that value by calling `setProperty(FOLLOW_ARTIFACT_REPOSITORY_REFERENCES, "false")` so I don't think we need a new constructor parameter but make the property public. 
@merks thats exactly how DirecorApplication works, but currently uses a *copy* of the constant `FOLLOW_ARTIFACT_REPOSITORY_REFERENCES`

Also note that there is another property `FOLLOW_REPOSITORY_REFERENCES` for *metadata* repositories what already is public.

merks:
This is specifically where the director application configures that:

https://github.com/eclipse-equinox/p2/blob/1b21340cef3b1cdcc22ccf687835be9e925cc344/bundles/org.eclipse.equinox.p2.director.app/src/org/eclipse/equinox/internal/p2/director/app/DirectorApplication.java#L1063-L1064

So I don't think this PR is needed because one can always achieve the goal by configuring the context properties.

laeubi:
> So I don't think this PR is needed because one can always achieve the goal by configuring the context properties.

As mentioned the property is private, we should make it public instead of maintain just another copy at PDE soon.
Beside that, it might be worth to have a method `ProvisioningContext#setFollowReferences(boolean)` that simply sets both properties to true/false to cover the common case and make this more obvious.

merks:
Let's not pollute the API with convenience methods that then must interact properly with the property settings.  In any case, that's not what this PR does.

Note that this is public:

- https://github.com/eclipse-equinox/p2/blob/1b21340cef3b1cdcc22ccf687835be9e925cc344/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/p2/engine/ProvisioningContext.java#L98

But this is not:

- https://github.com/eclipse-equinox/p2/blob/1b21340cef3b1cdcc22ccf687835be9e925cc344/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/p2/engine/ProvisioningContext.java#L110

That one could and probably should be made public and that would avoid the copy in the director:

image

But that's also not what this PR does.   So we don't need this PR but a different PR to make the constant public seems helpful.

tivervac:
Settings the property works for my usecase \o/

merks:
@tivervac 

Thanks for taking the time to contribute to such a complex project where it's all too easy to be thrown off by the endless details!

tivervac:
Replaced by https://github.com/eclipse-equinox/p2/pull/506/

PR #506, opened 2024-04-25
Make ProvisioningContext.FOLLOW_ARTIFACT_REPOSITORY_REFERENCES public to avoid duplication

tivervac:
It was duplicated in DirectorApplication, making it hard to follow the property around the code.

The original inspiration for this came from https://github.com/eclipse-equinox/p2/pull/505


bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/p2/engine/ProvisioningContext.java
```diff
@@ -107,7 +107,15 @@ public boolean contains(IArtifactRepository element) {
 	 */
 	public static final String CHECK_AUTHORITIES = "org.eclipse.equinox.p2.director.checkAuthorities"; //$NON-NLS-1$
 
-	private static final String FOLLOW_ARTIFACT_REPOSITORY_REFERENCES = "org.eclipse.equinox.p2.director.followArtifactRepositoryReferences"; //$NON-NLS-1$
+	/**
+	 * Instructs the {@link PhaseSetFactory#PHASE_COLLECT collect phase} to check
```
laeubi, line 111: I would suggest to reuse the documentation from `FOLLOW_REPOSITORY_REFERENCES` and simply replace 'metadata repositories' with 'artifact repositories'.

merks:
@laeubi 

The service segment has been incremented twice in this release cycle:

image

The `@since` in this PR is not okay because this was added for 2.10.200 not for 2.10.  So we really should increase this to 2.11.0.

Also the package constraint where this constant is used says 2.2.0 because that's the current version of the package, but that version hasn't changed here and so that version of the package does not necessary have the public constant.  So probably that should be increased too, probably to 2.11.0 like the bundle.

laeubi:
@merks  I think the package should become `2.3.0` now according to semver as this is a client compatible change and I don't see why it must be aligned with the bundle version (what we didn't do before).

tivervac:
Since I'm just trying to resolve the reference issue in general, I've added the [isEnabled](https://github.com/eclipse-pde/eclipse.pde/pull/1253#discussion_r1578895129) changes to this PR. If you prefer, I can split it up into two PRs.

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/p2/repository/IRepositoryReference.java
```diff
@@ -47,4 +47,15 @@ public interface IRepositoryReference {
...
+
+	/**
+	 * Returns whether this reference is enabled or not
+	 *
+	 * @see IRepository#ENABLED
+	 * @return whether this reference is enabled or not
+	 * @since 2.9
+	 */
+	default boolean isEnabled() {
+		return (getOptions() & IRepository.ENABLED) == IRepository.ENABLED;
```
laeubi, line 59:
```suggestion
		return (getOptions() & IRepository.ENABLED) != 0;
```

tivervac:
I've bumped `p2.engine` to 2.3.0 from `2.2.0`. I'm holding off on the 2.11 story until consensus is achieved. Personally, I expect that 2.11 bundle is necessary, but that a 2.3 package suffices.

merks:
> I've bumped `p2.engine` to 2.3.0 from `2.2.0`. I'm holding off on the 2.11 story until consensus is achieved. Personally, I expect that 2.11 bundle is necessary, but that a 2.3 package suffices.

I’m good with that.

bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/p2/engine/ProvisioningContext.java
```diff
@@ -107,7 +107,17 @@ public boolean contains(IArtifactRepository element) {
 	 */
 	public static final String CHECK_AUTHORITIES = "org.eclipse.equinox.p2.director.checkAuthorities"; //$NON-NLS-1$
 
-	private static final String FOLLOW_ARTIFACT_REPOSITORY_REFERENCES = "org.eclipse.equinox.p2.director.followArtifactRepositoryReferences"; //$NON-NLS-1$
+	/**
+	 * Instructs the provisioning context to follow artifact repository references
+	 * when providing queryables for obtaining metadata and artifacts. When this
```
laeubi, line 112: I think this only affects artifact (and not metadata) repositories?

merks: You did tell him to reuse the documentation of the other constant which is documented in this same way.  😱

laeubi: Reading it again, it is not really "wrong" just a bit confusing, essentially it says then when "when providing queryables for obtaining metadata and artifacts" is created this influences only the former (or later in this case).

laeubi:
@tivervac thanks for the imropvement.

PR #507, opened 2024-04-29
VersionRange for root features is too tight when expanded version given

laeubi:
Currently the (disabled by default) requirement for a root feature uses an exact version range when the feature version is expanded (e.g. Tycho) but the purpose of a root feature is that it can be updated what such a strict range makes it impossible (if activated).

This now adds a new IVersionRangeAdvice that allows to modify the result of a product included requirements version range.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.tests/src/org/eclipse/equinox/p2/tests/publisher/actions/ProductActionTest.java
+++ b/bundles/org.eclipse.equinox.p2.tests/src/org/eclipse/equinox/p2/tests/publisher/actions/ProductActionTest.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Optional;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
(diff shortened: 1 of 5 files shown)


bundles/org.eclipse.equinox.p2.tests/src/org/eclipse/equinox/p2/tests/publisher/actions/ProductActionTest.java
```diff
@@ -452,11 +471,15 @@ public void testInstallModeRootFeature() throws Exception {
...
 		assertTrue(installModeRootFeatureRequirement.isPresent());
-		IMatchExpression<IInstallableUnit> filter = installModeRootFeatureRequirement.get().getFilter();
+		IRequiredCapability capability = installModeRootFeatureRequirement.get();
+		IMatchExpression<IInstallableUnit> filter = capability.getFilter();
 		assertNotNull(filter);
 		String filterString = filter.getParameters()[0].toString();
 		assertEquals("(|(branded.product.install.mode.root=true)(org.eclipse.equinox.p2.install.mode.root=true))",
 				filterString);
+		VersionRange versionRange = capability.getRange();
+		assertTrue("Version Range is to strict: " + versionRange,
```
merks, line 481: to -> too

merks (approved):
Looks fine.

I questioned if we should upper bound with the next major version, but I do feature really respect semantic versioning and even if they do, what safety would this really buy us?  So I think only a lower bound is okay.

laeubi:
I think the best would be if one can configure it in the product editor like it is possible with feature dependencies, this would probably be even better than the current "root feature" approach. But that's more for a future enhancement.

PR #509, opened 2024-05-02
Adding capability to support PKI

JavaJoeS:
Update with option to install PKI from marketplace, org.eclipse.core.pki

### DIFF
--- a/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
+++ b/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
@@ -107,7 +107,7 @@ public class RepositoryTransport extends Transport {
 			}
 		} catch (NoSuchAlgorithmException e) {
 			// TODO Auto-generated catch block
-			e.printStackTrace();
+			// No TLS is installed, so let the process work as it would have with no mod
 		}
 		AuthenticationInfo loginDetails = null;
 		URI secureToDownload;
--- a/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
+++ b/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
@@ -107,11 +107,7 @@ public class RepositoryTransport extends Transport {
 			}
 		} catch (NoSuchAlgorithmException e) {
 			// TODO Auto-generated catch block
-<<<<<<< Upstream, based on origin/master
 			// No TLS is installed, so let the process work as it would have with no mod
-=======
-			e.printStackTrace();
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
```diff
@@ -95,7 +97,18 @@ public RepositoryTransport(IProvisioningAgent agent) {
...
 		boolean useJREHttp = false;
+		try {
+			// DELMARVA - is org.eclipse.core.pki installed and configured
+			if (SSLContext.getDefault().getProvider().getName().contains("TLS")) { //$NON-NLS-1$
+				useJREHttp = true;
+				Activator.getDefault().useJREHttpClient();
+			}
+		} catch (NoSuchAlgorithmException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
```
mickaelistria, line 110: Please use `ILog.log().error(e.getMessage(), e)`

JavaJoeS: @mickaelistria  I noticed that most of the other try catch blocks just throw and dont print anything out.  There are no other ILog messages in that bundle.  Maybe better to stick with the picture show and just throw?

JavaJoeS: @mickaelistria   Looking again, the software is supposed to be transparent and not required, so perhaps btter to not do anything in the catch block.  Software does not even need to know an exception may have occurred.

JavaJoeS:
@mickaelistria Please recheck..

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/RepositoryTransport.java
merks: Please 🙏 try to keep a PR to a single commit. Amend and force push further changes. Then you won’t be asked to squash the commits later.

laeubi:
> Update with option to install PKI from marketplace, org.eclipse.core.pki

I must confess I can't connect that statement to the actual code, the code seem to enable `useJREHttp`  if the default SSL context contains TLS (what else could it be), how is this connected to `org.eclipse.core.pki`? Why is this doen with P2 transport directly and not with ECF where the actual connection is made?

JavaJoeS:
@laeubi When there is no PKI enabled the default SSLContext is not TLS.

laeubi (changes requested):
Please rebase, squash commits and add documentation somewhere why / what this change does, forcing JRE http does not look "right" just because TLS is enabled (?)

laeubi:
Closing this due to missing feedback from the author, if this is still required please reopen with appropriate explanation and at best some test.

PR #517, opened 2024-05-16
Cache loaded repos in ProvisioningContext

laeubi:
Currently if fetching updates the provision context is queried multiple times leading to the repositories loaded multiple times, even though this is usually faster as repo are cached this still produces unnecessary workload, especially if there is one failed repository it will be tried to be loaded over and over again.

This now caches the repositories loaded on first access to speed that up.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/p2/engine/ProvisioningContext.java
+++ b/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/p2/engine/ProvisioningContext.java
@@ -208,18 +208,22 @@ public class ProvisioningContext {
 	}
 
 	private Set<IMetadataRepository> getLoadedMetadataRepositories(IProgressMonitor monitor) {
-		IMetadataRepositoryManager repoManager = agent.getService(IMetadataRepositoryManager.class);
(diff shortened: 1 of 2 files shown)


merks:
I guess this same problem doesn't arise for ProvisioningContext.getLoadedArtifactRepositories(IProgressMonitor)?

One concern I have here is that one can call org.eclipse.equinox.p2.engine.ProvisioningContext.setMetadataRepositories(URI...) in the API and result of getLoadedMetadataRepositories depends on that value so probably it would be most proper to clear the cache when that setter is called.

laeubi:
The ProvisionContext is kind of messy... e.g. it says

> The provisioning context has a distinct lifecycle, whereby the metadata and artifact repositories to be used are determined when the client retrieves retrieves the metadata queryable.  Clients should not reset the list of metadata repository cations or artifact repository locations once the metadata queryable has been retrieved.

in fact the artifact repositories are only loaded fully when the metadata is queried and so on.

merks (approved):
Looks fine.

repositorie  is spelled wrong, but it was that way before. 😱

laeubi:
Even though I think it is the right thing to do I'll rather close this now as it requires adjusting of existing test some people might rely on this behavior.

PR #519, opened 2024-05-19
Always use full-named o.e.equinox.launcher artifact

HannesWell:
For for Mac OS X there was no equinox-launcher fragment for CARBON with full name, i.e. with <ws>.<os>.<arch> suffix, but only a org.eclipse.equinox.launcher.carbon.macosx fragment. But since Eclipse 4.2, released 2012, CARBON is not supported anymore at all [1]. Therefore the code adjust to that special naming schema, which also imposed difficulties or new supported architectures on Macos should be removed.

See also https://github.com/eclipse-equinox/equinox/pull/622

[1] - https://download.eclipse.org/eclipse/updates/4.2/R-4.2-201206081400

### DIFF
--- a/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/EquinoxExecutableAction.java
+++ b/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/EquinoxExecutableAction.java
@@ -143,9 +143,7 @@ public class EquinoxExecutableAction extends AbstractPublisherAction {
 			String ws = config[0];
 			String os = config[1];
 			String arch = config[2];
-			String launcherFragment = EquinoxLauncherCUAction.ORG_ECLIPSE_EQUINOX_LAUNCHER + '.' + ws + '.' + os;
-			if (!(Constants.OS_MACOSX.equals(os) && !Constants.ARCH_X86_64.equals(arch)))
-				launcherFragment += '.' + arch;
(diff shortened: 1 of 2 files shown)


HannesWell:
@merks, @laeubi do you think this is something we should have for the upcoming summer release?
This would then probably allow us to remove the short-named launcher fragment for mac (see https://github.com/eclipse-equinox/equinox/pull/622). But I would leave the actual removal for the next release-cycle.

merks (approved):
It's rather weird and inconsistent in the current state:

image

So indeed it seems better to fix it sooner rather than later.

---

This logic is used primarily by Tycho so when Tycho starts using the new version then you can get rid of the older short name.  As a result, when you get rid if it, everyone building products against the latest platform builds must use a new release to Tycho, or their builds will fail, right?

HannesWell:
> So indeed it seems better to fix it sooner rather than later.

Agree.

> 
> As a result, when you get rid if it, everyone building products against the latest platform builds must use a new release to Tycho, or their builds will fail, right?

That's right. But I think in general it's hard to expect that old tools work with new content because we can't predict the future. For example PDE warns if the TP content is newer than the runtime.
But to have at least a three-months transition period I think it would be best to have this in now and remove the short-named launcher fragment for the next release.

> This logic is used primarily by Tycho so when Tycho starts using the new version then you can get rid of the older short name.

For the record, here is the pendant in https://github.com/eclipse-tycho/tycho/pull/3860. I have to admit that I'm not absolutely sure which change will be the one that influence the product assemble, but I assume its this one.

PR #524, opened 2024-06-11
Prevent NPE when no artifact repository is given

laeubi:
Currently RawMirrorRequest can fail if no repository is given in the descriptor and it want to log a message.

P2 should be gracefully in this case and simply report the location as being unknown.


merks:
Of course the usual version bump games kick in...

merks (approved):

laeubi:
@merks sorry for the delay, yes it seems soem version bumps are still required.

merks (approved):
Looks fine.

PR #534, opened 2024-08-01
Remove org.tukaani.xz from org.eclipse.equinox.p2.core.feature

merks:
- There's no reason to lock in a specific version via the feature.


merks:
Darned heap space error.  I'm so sick of how much of my time is wasted by such errors.  I could buy memory chips for the cost of my wasted time. 😱

PR #535, opened 2024-08-02
Compile Error fix for bouncycastle 1.78.1

MohananRahul:


bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/provisional/p2/repository/DefaultPGPPublicKeyService.java
```diff
@@ -742,9 +742,12 @@ private static abstract class GPGPubringSupplier implements Supplier<PGPPublicKe
 
 		public GPGPubringSupplier(Path pubring) {
 			this.pubring = pubring;
-			
+			try {
 				keyRingCollection = new PGPPublicKeyRingCollection(Collections.emptyList());
-			
+			} catch (Exception e) {
+				// Cannot happen for an empty collection.
+				throw new RuntimeException(e);
```
laeubi, line 749: If it can not happen throw Assertion Error with explanation message, or if we just think it *should* not happen one should first check if `e` is a Runtime Exception already before wrapping it again.

MohananRahul: Would this approach be okay? 
catch (Exception e) {
    if (e instanceof RuntimeException) {
        throw (RuntimeException) e;
    } else {
        throw new RuntimeException("Wrapping exception", e); 
    }

laeubi: A message like `throw new RuntimeException("Unexpected exception while PGPPublicKeyRingCollection", e);` would be better but basically yes.

bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/provisional/p2/repository/DefaultPGPPublicKeyService.java
```diff
@@ -742,9 +742,12 @@ private static abstract class GPGPubringSupplier implements Supplier<PGPPublicKe
 
 		public GPGPubringSupplier(Path pubring) {
 			this.pubring = pubring;
-			
+			try {
 				keyRingCollection = new PGPPublicKeyRingCollection(Collections.emptyList());
-			
+			} catch (Exception e) {
+				// Cannot happen for an empty collection.
+				throw new RuntimeException("Unexpected exception while PGPPublicKeyRingCollection",e);
```
laeubi, line 749: I would still check if e is a runtime exception first.

MohananRahul: While upgrading bouncycastle to 1.78.1 in 4.23 maintenance: 
This is exception I am getting,

[ERROR] 	keyRingCollection = new PGPPublicKeyRingCollection(Collections.emptyList());
[ERROR] 	                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Unhandled exception type IOException
[ERROR] /lts/git-eclipse/R4_23_maintenance/rt.equinox.p2/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/provisional/p2/repository/DefaultPGPPublicKeyService.java:[746] 
[ERROR] 	keyRingCollection = new PGPPublicKeyRingCollection(Collections.emptyList());
[ERROR] 	                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Unhandled exception type PGPException

PR #542, opened 2024-09-11
Update org.eclipse.equinox.p2.core.feature to 1.7.400

merks:
- There is a new version of org.eclipse.equinox.preferences.

https://github.com/eclipse-equinox/equinox/pull/663


laeubi:
Should we probably remove the inclusion? From P2 POV this is more a third party lib that is already pulled in by dependencies?

merks:
Yes, it's not clear what's best here.  As you can see, it's a growing list of transitive inclusions in any case.

PR #543, opened 2024-09-18
back-port: Remember referenced repositories as system repositories in…

gireeshpunathil:
…stead of user

Currently there is a quite surprising behavior, that if a user adds an updatesite that contains a referenced repository, that after an update-check additional repositories are visible. Even worse these are now used additionally to check for updates and if they contain other references these are also added and so on. This can result not only in a long list of sites the user never has added and has no clue where they are coming from but even pulling in unwanted or conflicting updates.

This now distinguishes two cases:

1) A repository is discovered by the RepositoryAction it is handled as a user added repository and becomes visible
2) A repository is discovered by a reference in that case it is handled as a system repository and not becomes visible

that way the list of user visible repositories stay clean from referenced repositories and unexpected side effects on update checks.


gireeshpunathil:
back port of https://github.com/eclipse-equinox/p2/pull/512. I have a customer asking for it.

tjwatson (approved):
LGTM

PR #545, opened 2024-09-18
Ensure that a jar:file: URI is considered local for artifact requests

merks:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2265


merks:
The pauseAndResume test failures on Windows/Mac are not related to this change.

PR #550, opened 2024-09-26
Internal code simplificatons in equinox.p2.publisher.eclipse

HannesWell:


HannesWell:
New warning is a false positive and reported to JDT in
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3025

merks:
Given p2's lack of defined APIs, you really should assume that when you change things you will break other interesting things even if they aren't API.  Minimally consider there are friends for a reason:

image

PR #558, opened 2024-10-12
Remove unnecessary specific entries in build.properties

HannesWell:


HannesWell:
> This pull request changes some projects for the first time in this development cycle.

Something went really wrong this time when assembling the commit message and I haven't noticed it before submitting it :/
https://github.com/eclipse-equinox/p2/commit/e55cc8fbadd1a6f8d3d76156536114fe2cc7968f

This was the run:
https://github.com/eclipse-equinox/p2/actions/runs/11304868644/job/31443754803

I try to investigate it.

HannesWell:
It turned out the problem existed from the very beginning of the versions-check workflow:
https://github.com/eclipse-equinox/p2/commit/ec65256121182aaae6df5f1726ee2338243423ec

But the cause is clear: All builds of p2 print the jvm-options at the moment:
https://github.com/eclipse-equinox/p2/blob/e55cc8fbadd1a6f8d3d76156536114fe2cc7968f/.mvn/jvm.config#L1

I created https://github.com/eclipse-equinox/p2/pull/559 to fix that for now, but will also enhance the workflow to not rely on printout but write the sdk-version property in a file. This requires more steps but is more robust.

HannesWell:
> > This pull request changes some projects for the first time in this development cycle.
> 
> Something went really wrong this time when assembling the commit message and I haven't noticed it before submitting it :/ [e55cc8f](https://github.com/eclipse-equinox/p2/commit/e55cc8fbadd1a6f8d3d76156536114fe2cc7968f)

That problem is now fixed with https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2450.

PR #559, opened 2024-10-12
[Build] Specify options for the build JVM only in Jenkins pipeline

HannesWell:
The options where introduced for all builds of the p2 repo via https://github.com/eclipse-equinox/p2/pull/464

Having it enabled can cause problems in GitHub workflows that are sensitive to the printout.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,2 +1,0 @@
--XX:+PrintFlagsFinal
--Xmx2g
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,15 +14,18 @@ pipeline {
 	}
 	stages {
 		stage('Build') {
+			environment {
+				MAVEN_OPTS ='-XX:+PrintFlagsFinal -Xmx2g'
+			}
 			steps {
-				wrap([$class: 'Xvnc', useXauthority: true]) {
+				xvnc(useXauthority: true) {
 					sh """
 					mvn clean verify --batch-mode --fail-at-end -Dmaven.repo.local=$WORKSPACE/.m2/repository \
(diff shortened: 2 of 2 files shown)


laeubi:
> Having it enabled can cause problems in GitHub workflows that are sensitive to the printout.

Should then not better the GitHub workflows be fixed? I'm not sure that one should depend on any special output of the build pipeline...

HannesWell:
> > Having it enabled can cause problems in GitHub workflows that are sensitive to the printout.
> 
> Should then not better the GitHub workflows be fixed?

I think both is appropriated, as mentioned in https://github.com/eclipse-equinox/p2/pull/558#issuecomment-2408539014

For the GH workflows to build P2 e.g. the Xmx config is not necessary.

> I'm not sure that one should depend on any special output of the build pipeline...

In hindsight this wasn't the best idea, but I followed to example given to me in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352#discussion_r1762288096 without thinking about this case. :)

laeubi:
> For the GH workflows to build P2 e.g. the Xmx config is not necessary.

I think its always good to have Jenkins + Github as close as possible to verify things on different platforms.

HannesWell:
> > For the GH workflows to build P2 e.g. the Xmx config is not necessary.
> 
> I think its always good to have Jenkins + Github as close as possible to verify things on different platforms.

In general yes, but since these options were just added to fix failures on Jenkins I'm not sure if it might be better to just apply it to Jenkins. Especially since it now clutters all other builds.

What are the opinions of the others?
To fix the version-increment workflow this is not necessary anymore.

HannesWell:
Ok, let's keep it as it is for now.

PR #565, opened 2024-10-18
Support empty lower and upper bounds in P2 VersionRange

HannesWell:
This enables support for version range specifications where the lower or upper bound is not specified and the delimiting bracket is directly next to the separating colon. For example `[1,)` or `(,1]` but also `(,)` can then be parsed successfully.

The type of supported brackets for an empty or unspecified bound is something that's not inherently obvious. From a mathematical perspective an empty bound can be considered equivalent to infinity. And in math may not say inclusive infinity. From that perspective only `[1,infinity)` respectively `[1,)` would be permitted, but not `[1,]`. But technically infinity has a specific value for P2 versions, i.e. `Version.MAX_VERSION`. So if one really wants to cover all possible values, the upper bound still has to be inclusive, i.e. `]`. For the lower bound the situation is similar, where `0.0.0` or the 'empty'-version is the smallest one possible.

Alternatively we could treat no lower bound simply as a synonym for `0.0.0` and no upper bound as synonym for `Version.MAX_VERSION` and consider the bounds as specified?

### DIFF
--- a/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/p2/metadata/VersionRange.java
+++ b/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/p2/metadata/VersionRange.java
@@ -157,11 +157,11 @@ public class VersionRange implements Serializable {
 		} else {
 			fmt = VersionFormat.OSGI_FORMAT;
 		}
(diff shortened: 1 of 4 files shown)


merks (approved):
This looks good.

Issue #568, opened 2024-11-18
FileInfoReader deadlock


jukzi:
Maybe an infinite loop waiting for org.eclipse.equinox.internal.p2.transport.ecf.FileInfoReader.barrier
waitOnSelf() does not synchronize the read on that while org.eclipse.equinox.internal.p2.transport.ecf.FileInfoReader.handleRemoteFileEvent(IRemoteFileSystemEvent) does synchronize for write

_Originally posted by @jukzi in https://github.com/eclipse-pde/eclipse.pde/issues/1481#issuecomment-2483087630_
            

original reports:
https://github.com/eclipse-pde/eclipse.pde/issues/1481
https://github.com/eclipse-pde/eclipse.pde/issues/1160

laeubi:
This code looks a bit suspicious on the other hand its there since 16 years now, so there are likely better ways today to solve this.

One part is also this one

https://github.com/eclipse-equinox/p2/blob/65348d689980d1e11708f934f1767f056e62907d/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/FileInfoReader.java#L150

it only notifies **one** thread, not all threads waiting... beside that it seem to be blocked in `job.join()` so has no chance to check the condition anyways...

laeubi:
Thread State / Dump would be more interesting than CPU sampling.

jukzi:
> Thread State / Dump would be more interesting than CPU sampling. e.g. why does PDE hinders the IDE to start at all?

see https://github.com/eclipse-pde/eclipse.pde/issues/1481#issue-2668550957

laeubi:
From that stacktrace it seem to hang in a Java Editor that triggers init of classpath while showing a view (what triggers PDE to do something).

Still I'm wondering why all this happens *before* the UI is displayed....

laeubi:
I have now analyzed this in more details, and the problem is `IDEApplication.start(IApplicationContext)` does suspend the `JobManager`, so no jobs are ever scheduled. Therefore the condition will never be meet and the code is running endlessly. It seems that this is particular happen with targlets (I can't reproduce this with plain PDE).

I tired to do some fixing for this in the `FileInfoReader` itself, but the problem is that ECF uses a `job` internally as well (that then will never run of course) so it does not help here much.

laeubi:
The class has been overhauled now to not using any jobs or simple synchronization primitives so I think its safe to close this issue.

[...]

PR #569, opened 2024-11-18
MT hardening FileInfoReader #568

jukzi:
The value of the barrier (true/false) was never read. It was only checked for !=null. The value had been read in waitOnSelf() without any multithreading semantics like synchronize/volatile. Instead now we use a thread safe AtomicBoolean. The synchronization blocks are still used for wait()/notifyAll()

https://github.com/eclipse-equinox/p2/issues/568

### DIFF
--- a/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/FileInfoReader.java
+++ b/bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/FileInfoReader.java
@@ -14,6 +14,7 @@ package org.eclipse.equinox.internal.p2.transport.ecf;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.ecf.core.*;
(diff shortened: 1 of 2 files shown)


laeubi (changes requested):
The purpose of the group of `Atomic` primitives is that they don't require any synchronization at all (and some tools will complain that this is code-smell).

Why not use a real barrier (e.g. `CountDownLatch`) here that do not require syncronization.

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/FileInfoReader.java
```diff
@@ -60,14 +61,13 @@ protected IStatus run(IProgressMonitor monitor) {
...
-	 * file transfer were a custom job can be supplied.
-	 * TODO: log an issue for ECF.
+	 * Waits until request is processed (barrierReached.get()). This is a bit of a
+	 * hack, as it would be better if the ECFBrowser worked in similar fashion to
+	 * file transfer were a custom job can be supplied. TODO: log an issue for ECF.
 	 */
 	private void waitOnSelf() {
 		schedule();
-		while (barrier[0] == null) {
+		while (!barrierReached.get()) {
```
laeubi, line 70: I must confess that I can't get what is the purpose of this loop... if I schedule a job **once** how can I **join** it more than once?

@merks any clue whats the purpose of this?

jukzi:
CountDownLatch can not be reset. CyclicBarrier can be reset but not read/mixed with the wait loop that askes for cancel. If you find any better solution please do. Meanwhile this could improve the situation.

bundles/org.eclipse.equinox.p2.transport.ecf/src/org/eclipse/equinox/internal/p2/transport/ecf/FileInfoReader.java
```diff
@@ -34,21 +35,21 @@
...
+	private final AtomicBoolean barrierReached = new AtomicBoolean();
+	private volatile IRemoteFile[] remoteFiles;
+	private volatile IRemoteFileSystemRequest browseRequest;
 
 	@Override
 	protected IStatus run(IProgressMonitor monitor) {
-		synchronized (barrier) {
-			while (barrier[0] == null) {
+		synchronized (barrierReached) {
+			while (!barrierReached.get()) {
```
laeubi, line 50: This changes behavior before any value TRUE/FALSE would have ended the loop now it loops forever if set to false.

merks:
We're in the RC2 phase so **nothing** will get done for this release. 

I'm really uncomfortable changing something here simply because we theorize that it might be broken with no reproducible way to test that we've actually fixed what we think is the problem rather than simply changed the behavior a bit and hoping for the best that this somehow helps something. How many times has even the most simple, innocent change broken something?  I've lost count.   But that's just my opinion and perhaps it's simply paranoia...

jukzi:
I don't want this for any RC but M1. Simply adding a synchronize around an endless loop would harm more then help: The only use of synchronize in that class is the notify/wait mechanism  - that does not exist in waitOnSelf().

laeubi:
> Simply adding a synchronize around an endless loop would harm more then help:

I didn't suggested to put it _around_ the loop ...

> The only use of synchronize in that class is the notify/wait mechanism - that does not exist in waitOnSelf().

Then please fix up the PR/Commit message this PR seem to do something different then.

jukzi:
> Maybe it would be better to have two PRs for two things as well.

@laeubi please provide such.

PR #573, opened 2024-11-23
Revert "Use CI Friendly versions for the P2 maven build"

HannesWell:
This reverts commit 612c5fa6fc41602aaa59624e36c06c58acaea863.

Using this pattern breaks the `Update to next release ` workflow:
- https://github.com/eclipse-equinox/p2/actions/runs/11973081501
- https://github.com/eclipse-tycho/tycho/pull/3900#issuecomment-2495468974

This reverts it to use versions the default way again. Once Tycho is fixed we can revert this again.
Or we just keep it then, since due more usage of tycho-pomless the number of locations to change is very low now and the `Update to next release ` workflow does it automatically any ways.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.tests/pom.xml
+++ b/bundles/org.eclipse.equinox.p2.tests/pom.xml
@@ -10,7 +10,7 @@
 	<parent>
 		<groupId>org.eclipse.platform</groupId>
 		<artifactId>rt.equinox.p2</artifactId>
-		<version>${releaseVersion}${qualifier}</version>
+		<version>4.34.0-SNAPSHOT</version>
 		<relativePath>../../</relativePath>
 	</parent>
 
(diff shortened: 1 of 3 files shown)


laeubi (changes requested):
Please use an older Tycho release then or a SNAPSHOT version after it is fixed. Reverting a 2 year old commit seem not to be a proper fix.

HannesWell:
It's indeed more of a workaround.
But with the current state of tycho-pomless usage and the tools we have now the advantage of using ci-friendly versions does not seem to be as helpful as it used to be. 
And I also see an advantage in using the same 'pattern' as everywhere else.

laeubi:
> But with the current state of tycho-pomless usage and the tools we have now the advantage of using ci-friendly versions does not seem to be as helpful as it used to be.

Just because it is not leveraged do not mean its not useful, e.g. even at the current state only one property needs to be updated (the parent version) and one can perform a release with proper versions in the metadata.

Also without using this feature the regression in Tycho would probably not be discovered.

> And I also see an advantage in using the same 'pattern' as everywhere else.

Actually the goal is to use CI-Friendly versions everywhere P2 is just the first project that was converted to act as a demonstration/incubator.

HannesWell:
Update was now done manually.

PR #574, opened 2024-11-24
Update for release 4.35

HannesWell:
As suggested in https://github.com/eclipse-equinox/p2/pull/573, this applies the changes necessary for release preparation manually, that cannot be done automatically at the moment due to a bug in Tycho. See
- https://github.com/eclipse-tycho/tycho/pull/3900#issuecomment-2495468974


akurtakov:
New warnings are due to third party dependency deprecations.

PR #576, opened 2024-11-29
Correct improper usage of the SubMonitor

ptziegler:
From the documentation:
> Since it is illegal to call beginTask on the same IProgressMonitor
> more than once, the same instance of IProgressMonitor must not be
> passed to convert more than once.

The proper way is to replace calls to IProgressMonitor.beginTask(...) with SubMonitor.convert(...), keep the SubMonitor as a local variable and use that from now on.

This amends commit 9009085b72e9507fdc4b7d0adb87883adefc5637.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.discovery/src/org/eclipse/equinox/internal/p2/discovery/Catalog.java
+++ b/bundles/org.eclipse.equinox.p2.discovery/src/org/eclipse/equinox/internal/p2/discovery/Catalog.java
@@ -73,28 +73,24 @@ public class Catalog {
 
 		final int totalTicks = 100000;
 		final int discoveryTicks = totalTicks - (totalTicks / 10);
-		monitor.beginTask(Messages.Catalog_task_discovering_connectors, totalTicks);
-		try {
-			for (AbstractDiscoveryStrategy discoveryStrategy : discoveryStrategies) {
-				if (monitor.isCanceled()) {
-					status.add(Status.CANCEL_STATUS);
-					break;
-				}
-				discoveryStrategy.setCategories(newCategories);
-				discoveryStrategy.setItems(newItems);
-				discoveryStrategy.setCertifications(newCertifications);
-				discoveryStrategy.setTags(newTags);
-				try {
-					discoveryStrategy.performDiscovery(SubMonitor.convert(monitor, discoveryTicks / discoveryStrategies.size()));
-				} catch (CoreException e) {
-					status.add(new Status(IStatus.ERROR, DiscoveryCore.ID_PLUGIN, NLS.bind(Messages.Catalog_Strategy_failed_Error, discoveryStrategy.getClass().getSimpleName()), e));
-				}
+		SubMonitor subMonitor = SubMonitor.convert(monitor, Messages.Catalog_task_discovering_connectors, totalTicks);
+		for (AbstractDiscoveryStrategy discoveryStrategy : discoveryStrategies) {
+			if (subMonitor.isCanceled()) {
+				status.add(Status.CANCEL_STATUS);
+				break;
 			}
-
-			update(newCategories, newItems, newCertifications, newTags);
-		} finally {
-			monitor.done();
+			discoveryStrategy.setCategories(newCategories);
+			discoveryStrategy.setItems(newItems);
+			discoveryStrategy.setCertifications(newCertifications);
+			discoveryStrategy.setTags(newTags);
+			try {
+				discoveryStrategy.performDiscovery(subMonitor.newChild(discoveryTicks / discoveryStrategies.size()));
+			} catch (CoreException e) {
+				status.add(new Status(IStatus.ERROR, DiscoveryCore.ID_PLUGIN, NLS.bind(Messages.Catalog_Strategy_failed_Error, discoveryStrategy.getClass().getSimpleName()), e));
+			}
 		}
+
+		update(newCategories, newItems, newCertifications, newTags);
 		return status;
 	}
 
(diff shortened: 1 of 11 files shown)


ptziegler:
I noticed that the target definition resolution usually gets stuck at 47%, which I believe is partially caused by a lot of funny stuff being done to the progress monitor...

image

vogella (commented):
I'm just looking via my phone but I assume that the done() call (and finally block) can be removed. See https://www.eclipse.org/articles/Article-Progress-Monitors/article.html

vogella:
See section 5.7

ptziegler:
> I'm just looking via my phone but I assume that the done() call (and finally block) can be removed. See https://www.eclipse.org/articles/Article-Progress-Monitors/article.html

Technically yes, but I think my mistake was that the call to done still needs to be called for the main monitor. Meaning the calls in the finally block should remain `monitor.done()`.

> It is not necessary to call done() on the result, but the caller is responsible for calling done() on the argument.

bundles/org.eclipse.equinox.p2.discovery/src/org/eclipse/equinox/internal/p2/discovery/Catalog.java
```diff
@@ -73,10 +73,10 @@ public IStatus performDiscovery(IProgressMonitor monitor) {
 
 		final int totalTicks = 100000;
 		final int discoveryTicks = totalTicks - (totalTicks / 10);
-		monitor.beginTask(Messages.Catalog_task_discovering_connectors, totalTicks);
+		SubMonitor subMonitor = SubMonitor.convert(monitor, Messages.Catalog_task_discovering_connectors, totalTicks);
 		try {
 			for (AbstractDiscoveryStrategy discoveryStrategy : discoveryStrategies) {
-				if (monitor.isCanceled()) {
```
vogella, line 79: Is this extra check necessary? We do a split (which auto checks for cancellations ) just a few lines later

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/query/QueryableUpdates.java
```diff
@@ -43,16 +43,17 @@ public IQueryResult<IInstallableUnit> query(IQuery<IInstallableUnit> query, IPro
...
 		int totalWork = 2000;
-		monitor.beginTask(ProvUIMessages.QueryableUpdates_UpdateListProgress, totalWork);
+		SubMonitor subMonitor = SubMonitor.convert(monitor, ProvUIMessages.QueryableUpdates_UpdateListProgress,
+				totalWork);
 		IPlanner planner = ui.getSession().getProvisioningAgent().getService(IPlanner.class);
 		try {
 			Set<IInstallableUnit> allUpdates = new HashSet<>();
 			for (IInstallableUnit unit : iusToUpdate) {
-				if (monitor.isCanceled())
+				if (subMonitor.isCanceled())
```
vogella, line 52: Same here

bundles/org.eclipse.equinox.p2.discovery/src/org/eclipse/equinox/internal/p2/discovery/Catalog.java
laeubi: > Is this extra check necessary? We do a split (which auto checks for cancellations ) just a few lines later

This is not completely the same. In one case you can silently return in the other an exception is thrown.

laeubi:
> It is not necessary to call done() on the result, but the caller is responsible for calling done() on the argument

I think this refers to the "top most caller" that is the one that created the monitor / acquired it for the first time. For a Job I would expect that the job framework is calling `done()` after the job is finished.

bundles/org.eclipse.equinox.p2.discovery/src/org/eclipse/equinox/internal/p2/discovery/Catalog.java
vogella: Split also throws this exception

vogella:
Yes to @laeubi statement of the done call. If you don't create the monitor you do not have to call done

vogella:
@ptziegler we have also tracing to find incorrect usage of progress monitoring, see the article (search for tracing).

Years ago @kthoms and I a spend lot of time trying to clean the usage up in platform and pde but there were to many to fix them all .

bundles/org.eclipse.equinox.p2.discovery/src/org/eclipse/equinox/internal/p2/discovery/Catalog.java
ptziegler: That's true. But given that the method is supposed to return gracefully upon cancellation, rather than throwing an exception, I should use `newChild()` rather than `split()`.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/query/QueryableUpdates.java
ptziegler: This whole `if` check can go, given that the `OperationCanceledException` is explicitly caught. In both cases, `Collector.emptyCollector()` is returned.

ptziegler:
I've removed the try-finally blocks that only serve to call `monitor.done()`. I've also replaced most calls to `SubMonitor.split()` with `SubMonitor.newChild()` because most methods are expected to return gracefully and therefore an OperationCanceledException must not be thrown.

> we have also tracing to find incorrect usage of progress monitoring, see the article (search for tracing).

Looks interesting. Perhaps I can give it a go later on. It wouldn't surprise me if there are other places where the SubProgressMonitor was replaced too eagerly.

vogella:
@ptziegler is this ready from your side?

ptziegler:
> @ptziegler is this ready to be merged from your side?

Yes. From my side, I'm done.

vogella:
Build works fine with GH  actions. The Jenkins build is also successful but fail after the build in the publish bit, which seems unrelated.

05:45:15  [INFO] ------------------------------------------------------------------------
05:45:15  [INFO] BUILD SUCCESS
05:45:15  [INFO] ------------------------------------------------------------------------
05:45:15  [INFO] Total time:  20:56 min (Wall Clock)
05:45:15  [INFO] Finished at: 2024-12-03T10:45:15Z
05:45:15  [INFO] ------------------------------------------------------------------------
[Pipeline] }

vogella:
Thanks @ptziegler, looking forward for tomorrow I-build improved target platform resolution

Issue #567, opened 2024-11-15
Gathering certificates can get stuck if the server doesn't respond


danthe1st:
Someone on Stack Overflow [experienced an issue](https://stackoverflow.com/q/79193830/10871900) where installing updates got stuck [waiting for the certificates](https://github.com/eclipse-equinox/p2/blob/65348d689980d1e11708f934f1767f056e62907d/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java#L266) (which was introduced with #235).

`httpResponse.get()` might wait forever if [the request](https://github.com/eclipse-equinox/p2/blob/65348d689980d1e11708f934f1767f056e62907d/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java#L254-L256) takes forever which is possible because no timeout is set and [`HttpClient` blocks forever by default](https://docs.oracle.com/en/java/javase/21/docs/api/java.net.http/java/net/http/HttpRequest.Builder.html#timeout(java.time.Duration)). I didn't find any other uses of `HttpRequest.newBuilder()` in p2.

I suggest adding a timeout to this and I am willing to contribute that (unless someone else really wants to do that).

Are there any opinions on a good timeout? I would suggest something along the line of 2-5 seconds since the HTTP request taking more time would indicate some sort of issue in which case not accepting the certificate is probably fine.

merks:
Here are some timeout values used by ECF which is used by p2. 

![image](https://github.com/user-attachments/assets/3435d57c-ec52-49e2-b6de-afc32299caed)

Having a default value that can be modified via a system property would be good I think.

The tricky question is how to handle a timeout, with a retry like in other places in p2?  And, most importantly, how is the failure to identify/retrieve the host certificate presented to the user?

Is this some type of problem with the host not supporting a head request?  Is there a way I can reproduce the problem.  Sometime is a little fishy if we can load the repository but we cannot get a certificate...

danthe1st:
> Is this some type of problem with the host not supporting a head request?

I think it happens if the server accepts the connection but infinitely waits for the response so it would only happen if something is wrong with the server (but it should still not lead to it being stuck forever). I don't know what exactly caused it for the user who experienced it but I guess the server causing the issue probably had some issue with `HEAD` requests or similar.

> The tricky question is how to handle a timeout, with a retry like in other places in p2?

[...]

PR #577, opened 2024-11-30
Add timeout to gathering host certificates #567

danthe1st:
As explained in #567, it is currently possible that Eclipse waits infinitely for a host certificate in case the server accepting the request but not responding after that. This issue [caused the plugin installation of a Stack Overflow user to get stuck](https://stackoverflow.com/q/79193830/10871900).

This PR adds a timeout and retrying using the properties `org.eclipse.equinox.p2.engine.requestTimeout` and `org.eclipse.equinox.p2.engine.requestRetries`.

### How to test
- Create a server that accepts incoming connections but never responds, e.g. `nc -l 8080 -k` on Linux or using Java code similar to the following (not using TLS for simplicity):
  ```java
  try(ServerSocket servSock = new ServerSocket(8080)){
      while(true){
          try(Socket sock = servSock.accept();
              BufferedReader br = new BufferedReader(new InputStreamReader(sock.getInputStream()))){
              br.lines().forEach(System.out::println);
          }
      }
  }
  ```
- In the Eclipse application, go to `Preferences` > `Install/Update` > `Trust` > `Authorities`
- Change `http` to `Allow`
- Click `Add`
- Enter `http://localhost:8080`

Without this change, Eclipse sends one HTTP request to this server, never closes the request and [the thread is blocked forever](https://github.com/eclipse-equinox/p2/blob/65348d689980d1e11708f934f1767f056e62907d/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java#L266).

With this change, it sends multiple (by default 3) requests to the server and then skips the host.

### Other changes

I also added a warning log when gathering certificates throws an exception. I made sure the current thread is reinterrupted when an `InterruptedException` is thrown (and not logging anything since that seems unnecessary).

### Notes

Since I didn't find a property that I felt was actually applicable to reuse for the timeout and retry count, I created new ones. I decided on a default of 2 retries (i.e. 3 requests in total) with a timeout (not the connection timeout but the timeout of [the request](https://docs.oracle.com/en/java/javase/21/docs/api/java.net.http/java/net/http/HttpRequest.Builder.html#timeout(java.time.Duration)), i.e. when waiting for a response after connecting) of 5 seconds. I didn't include a backoff time between requests but I could still add that if wanted.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.core/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.core;singleton:=true
-Bundle-Version: 2.12.200.qualifier
+Bundle-Version: 2.13.0.qualifier
 Bundle-Activator: org.eclipse.equinox.internal.p2.core.Activator
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/IProvisioningAgent.java
+++ b/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/IProvisioningAgent.java
@@ -165,4 +165,32 @@ public interface IProvisioningAgent {
 		return Boolean.parseBoolean(getProperty(key, Boolean.toString(defaultValue)));
 	}
 
+	/**
+	 * Returns an agent bound property as an int, the default implementation
+	 * delegates to {@link IProvisioningAgent#getIntProperty(String, int)} with
+	 * <code>0</code> as the default.
+	 *
+	 * @since 2.13
+	 */
+	default int getIntProperty(String key) {
+		return getIntProperty(key, 0);
+	}
+
+	/**
+	 * Returns an agent bound property as an int, the default implementation
+	 * delegates to {@link IProvisioningAgent#getProperty(String)} with the given
+	 * default as a String and parses the result as by
+	 * {@link Boolean#parseBoolean(String)}.
+	 *
+	 * @since 2.13
+	 */
+	default int getIntProperty(String key, int defaultValue) {
+		String propertyValue = getProperty(key, Integer.toString(defaultValue));
+		try {
+			return Integer.parseInt(propertyValue);
+		} catch (NumberFormatException e) {
+			return defaultValue;
+		}
+	}
+
 }
--- a/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java
+++ b/bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java
@@ -16,6 +16,7 @@ import java.net.http.*;
 import java.net.http.HttpRequest.BodyPublishers;
 import java.net.http.HttpResponse.BodyHandlers;
 import java.security.cert.Certificate;
+import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
@@ -25,6 +26,7 @@ import java.util.stream.Collectors;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.equinox.internal.p2.core.helpers.LogHelper;
 import org.eclipse.equinox.internal.p2.engine.EngineActivator;
 import org.eclipse.equinox.internal.p2.engine.Messages;
 import org.eclipse.equinox.p2.core.*;
@@ -52,6 +54,10 @@ public class AuthorityChecker {
 	private static final Pattern HIERARCHICAL_URI_PATTERN = Pattern
 			.compile("((?:[^/:]+):(?://[^/]+|///|/)?)([^?#]*)([#?].*)?"); //$NON-NLS-1$
 
+	private final int requestTimeoutMillis;
+
+	private final int maxRequestRetries;
+
 	private final IProvisioningAgent agent;
 	private final ProvisioningContext context;
 	private final IProfile profile;
@@ -75,6 +81,9 @@ public class AuthorityChecker {
 		this.ius = ius;
 		this.artifacts = artifacts;
 		this.profile = profile;
+		requestTimeoutMillis = agent.getIntProperty("org.eclipse.equinox.p2.engine.certificateRequestTimeout", 5000);//$NON-NLS-1$
+		maxRequestRetries = agent.getIntProperty("org.eclipse.equinox.p2.engine.certificateRequestRetries", //$NON-NLS-1$
+				3);
 	}
 
 	IStatus start(IProgressMonitor monitor) {
@@ -235,7 +244,7 @@ public class AuthorityChecker {
 		return filteredAuthorities;
 	}
 
-	public static Map<URI, List<Certificate>> getCertificates(Collection<? extends URI> uris,
+	public Map<URI, List<Certificate>> getCertificates(Collection<? extends URI> uris,
 			IProgressMonitor monitor) {
 		var certificates = new TreeMap<URI, List<Certificate>>();
 		var authorities = new TreeMap<URI, List<Certificate>>();
@@ -247,13 +256,14 @@ public class AuthorityChecker {
 		return certificates;
 	}
 
-	public static void gatherCertificates(Map<URI, List<Certificate>> authorities, IProgressMonitor montior) {
+	public void gatherCertificates(Map<URI, List<Certificate>> authorities, IProgressMonitor montior) {
 		var client = HttpClient.newBuilder().build();
 		var requests = authorities.keySet().stream().collect(Collectors.toMap(Function.identity(), uri -> {
 			try {
-				return Optional.of(client.sendAsync(
-						HttpRequest.newBuilder().uri(uri).method("HEAD", BodyPublishers.noBody()).build(), //$NON-NLS-1$
-						BodyHandlers.ofString()));
+				var request = HttpRequest.newBuilder().uri(uri).timeout(Duration.ofMillis(requestTimeoutMillis))
+						.method("HEAD", BodyPublishers.noBody()) //$NON-NLS-1$
+						.build();
+				return Optional.of(sendHttpRequestOrRetry(client, request, maxRequestRetries));
 			} catch (RuntimeException ex) {
 				return Optional.<CompletableFuture<HttpResponse<String>>>ofNullable(null);
 			}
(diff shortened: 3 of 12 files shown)


danthe1st:
Do I also need to update the versions in the `feature.xml` files (Jenkins fails with `Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.10:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.equinox.p2.core.feature: Only qualifier changed for (org.eclipse.equinox.p2.core.feature.feature.jar/1.7.400.v20241130-2028). Expected to have bigger x.y.z than what is available in baseline (1.7.400.v20241019-1812)`)?
Since I haven't contributed here before, I don't want to do this before hearing from anyone first as that seems like it could easily introduce merge conflicts so I wouldn't change it if it wasn't necessary.

danthe1st:
I hope the way I updated the versions in the `MANIFEST.MF` and `feature.xml` files is correct (I haven't contributed code to Eclipse repos before).

merks:
I can confirm that PR behaves described:

image

You've done a good job figuring out what's needed in terms of version increments for manifests and features.

bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java
```diff
@@ -52,6 +54,11 @@ public class AuthorityChecker {
 	private static final Pattern HIERARCHICAL_URI_PATTERN = Pattern
 			.compile("((?:[^/:]+):(?://[^/]+|///|/)?)([^?#]*)([#?].*)?"); //$NON-NLS-1$
 
+	private static final int REQUEST_TIMEOUT_MS = Integer.getInteger("org.eclipse.equinox.p2.engine.requestTimeout", 5000); //$NON-NLS-1$
```
merks, line 57: I think being a little bit more explicit about the what is being requested would be good.
```suggestion
	private static final int REQUEST_TIMEOUT_MS = Integer.getInteger("org.eclipse.equinox.p2.engine.certificateRequestTimeout", 5000); //$NON-NLS-1$
```

bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java
```diff
@@ -52,6 +54,11 @@ public class AuthorityChecker {
...
+
+	private static final int MAX_REQUEST_RETRIES = Integer.getInteger("org.eclipse.equinox.p2.engine.requestRetries", //$NON-NLS-1$
```
merks, line 59:
```suggestion
	private static final int MAX_REQUEST_RETRIES = Integer.getInteger("org.eclipse.equinox.p2.engine.certificateRequestRetries", //$NON-NLS-1$
```

bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java
```diff
@@ -52,6 +54,11 @@ public class AuthorityChecker {
...
+			2);
```
merks, line 60: I know the value of 2 leads to 3 retries, but just looking at the property name, that's confusing.
```suggestion
			3);
```

bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java
```diff
@@ -270,16 +278,29 @@ public static void gatherCertificates(Map<URI, List<Certificate>> authorities, I
...
+							Messages.AuthorityChecker_GatherCertificatesFailure, e));
 				}
 			});
 		}
 	}
 
+	private static CompletableFuture<HttpResponse<String>> sendHttpRequestOrRetry(HttpClient client,
+			HttpRequest request, int retriesLeft) {
+		var future = client.sendAsync(request, BodyHandlers.ofString());
+		if (retriesLeft > 0) {
```
merks, line 298: Of course if we change the value from 2 to 3 above we should make sure to change this to respect that number.  It took me a minute to realize the like 297 creates on requested and that the recursive calls create two more.
```suggestion
		if (retriesLeft > 1) {
```

merks (changes requested):
Overall it's nice work and much appreciated!  Just a few very tiny nits and I think we're good to go.

danthe1st:
> Overall it's nice work and much appreciated! Just a few very tiny nits and I think we're good to go.

I have committed the requested changes. I assume the GitHub Actions failure (`Component 'Provisioning Admin UI RCP (Incubation)' in the baseline 'workspace' is disposed [main]`) on Windows before is unrelated to my changes?

merks (approved):
Looks good now.  Could you please squash this into a single commit and force push it?

bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java
```diff
@@ -52,6 +54,13 @@ public class AuthorityChecker {
 	private static final Pattern HIERARCHICAL_URI_PATTERN = Pattern
 			.compile("((?:[^/:]+):(?://[^/]+|///|/)?)([^?#]*)([#?].*)?"); //$NON-NLS-1$
 
+	private static final int REQUEST_TIMEOUT_MS = Integer
```
laeubi, line 57: Instead of system properties, the best is to fetch these from the `IProvisioningAgent`, this has several advantages e.g. you can configure this per agent and it automatically fall back to system properties.

bundles/org.eclipse.equinox.p2.engine/src/org/eclipse/equinox/internal/p2/engine/phases/AuthorityChecker.java
```diff
@@ -52,6 +54,13 @@ public class AuthorityChecker {
...
+			.getInteger("org.eclipse.equinox.p2.engine.certificateRequestTimeout", 5000); //$NON-NLS-1$
+
+	private static final int MAX_REQUEST_RETRIES = Integer.getInteger(
```
laeubi, line 60: Same here.

bundles/org.eclipse.equinox.p2.core/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.core;singleton:=true
-Bundle-Version: 2.12.200.qualifier
+Bundle-Version: 2.13.100.qualifier
```
merks, line 5: This is almost right.  This should be 2.13.0.
```suggestion
Bundle-Version: 2.13.0.qualifier
```

bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/IProvisioningAgent.java
```diff
@@ -165,4 +165,32 @@ default boolean getBooleanProperty(String key, boolean defaultValue) {
...
+	/**
+	 * Returns an agent bound property as an int, the default implementation
+	 * delegates to {@link IProvisioningAgent#getIntProperty(String, int)} with
+	 * <code>0</code> as the default.
+	 *
+	 * @since 2.13
+	 */
+	default int getIntProperty(String key) {
+		return getIntProperty(key, 0);
+	}
```
merks, line 177: This does no harm but it's arguably overkill because what are the odds that we will need an integer property that default to 0.

danthe1st: So I guess I should remove it? I added this to imitate boolean properties.

bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/IProvisioningAgent.java
```diff
@@ -165,4 +165,32 @@ default boolean getBooleanProperty(String key, boolean defaultValue) {
...
+	/**
+	 * Returns an agent bound property as an int, the default implementation
+	 * delegates to {@link IProvisioningAgent#getProperty(String)} with the given
+	 * default as a String and parses the result as by
+	 * {@link Boolean#parseBoolean(String)}.
+	 *
+	 * @since 2.13
+	 */
+	default int getIntProperty(String key, int defaultValue) {
+		String propertyValue = getProperty(key, Integer.toString(defaultValue));
```
merks, line 188: Also a nit, but I suppose one could call getProperty without a default value and if the propertyValue comes back null return the default. I don't mind either way, so don't feel compelled to change it.

danthe1st: In that case, `Integer.parseInt` would throw a NPE leading to the default value being returned.

merks (changes requested):
Only the version of  bundles/org.eclipse.equinox.p2.core/META-INF/MANIFEST.MF should be fixed.  The rest is at your discretion.

Thanks for being so super responsive!  You're awesome!!

bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/IProvisioningAgent.java
merks: It's not wrong how it is, but if you checked  getProperty(key) for returning null, you'd not have to convert defaultValue to a string and then parse that string.  As I said it's really a nit.

bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/IProvisioningAgent.java
merks: Good point about getBoolean.  As such, I think it's fine and good to leave it.  Best it all be parallel.

bundles/org.eclipse.equinox.p2.core/META-INF/MANIFEST.MF
danthe1st: Done.

bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/IProvisioningAgent.java
merks: Enough nit picking.  It's okay like this.

bundles/org.eclipse.equinox.p2.core/META-INF/MANIFEST.MF
merks: Very good.

merks:
When the build passes, it's ready to merge.  🥇

merks (approved):

merks:
Thanks again!!  ❤️

PR #578, opened 2024-12-03
DownloadManager - remove unnecessary done() call on ProgressMonitor

vogella:
See https://www.eclipse.org/articles/Article-Progress-Monitors/article.html


vogella:
@ptziegler a full text search shows tons of .done() in the repo.

Can you have a quick look at this trivial change?

ptziegler (approved):
This method already has a SubMonitor, which explicitly says that `done()` doesn't need to be called. So that looks fine to me.

vogella:
Thanks @ptziegler

PR #579, opened 2024-12-03
Remove unnecessary done() call on ProgressMonitor

vogella:
See https://www.eclipse.org/articles/Article-Progress-Monitors/article.html


vogella:
@ptziegler please have a quick look.

ptziegler (approved):
CompositeArtifactRepository is calling `done()` on the SubMonitor, so that should be fine as well and CompositeArtifactRepositoryFactory is also only actively using the SubMonitor, making it the responsibility of the callert to call `done()`. This should then also be fine.

PR #583, opened 2024-12-18
Version bump for new stream

akurtakov:
As reported at
https://download.eclipse.org/eclipse/downloads/drops4/I20241218-0320/buildlogs/reporeports/reports/versionChecks.html


akurtakov:
Jennkins reports same number of successful tests as master build.

PR #586, opened 2024-12-19
Ensure DownloadJob.run(IProgressMonitor) provides monitor task details

ptziegler:
IProgressMonitor.slice() only propagates the work to its parent monitor, but not any details.

Follow-up to https://github.com/eclipse-equinox/p2/pull/576

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/DownloadJob.java
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/DownloadJob.java
@@ -15,6 +15,8 @@
 package org.eclipse.equinox.internal.p2.artifact.repository.simple;
 
 import java.util.LinkedList;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
 import org.eclipse.core.runtime.*;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.equinox.p2.repository.artifact.IArtifactRequest;
@@ -24,11 +26,11 @@ public class DownloadJob extends Job {
 
 	private final LinkedList<IArtifactRequest> requestsPending;
 	private final SimpleArtifactRepository repository;
-	private final IProgressMonitor masterMonitor;
+	private final SubMonitor masterMonitor;
 	private final MultiStatus overallStatus;
 
 	DownloadJob(String name, SimpleArtifactRepository repository, LinkedList<IArtifactRequest> requestsPending,
-			IProgressMonitor masterMonitor, MultiStatus overallStatus) {
+			SubMonitor masterMonitor, MultiStatus overallStatus) {
 		super(name);
 		setSystem(true);
 		this.repository = repository;
@@ -56,8 +58,7 @@ public class DownloadJob extends Job {
 			if (masterMonitor.isCanceled())
 				return Status.CANCEL_STATUS;
 			// process the actual request
-			SubMonitor subMonitor = SubMonitor.convert(masterMonitor.slice(1), 1);
-			IStatus status = repository.getArtifact(request, subMonitor);
+			IStatus status = repository.getArtifact(request, new ThreadSafeProgressMonitor(1));
 			if (!status.isOK()) {
 				synchronized (overallStatus) {
 					overallStatus.add(status);
@@ -67,5 +68,76 @@ public class DownloadJob extends Job {
 
 		jobMonitor.done();
 		return Status.OK_STATUS;
+	}
+
+	/**
+	 * Wrapper around the general {@link IProgressMonitor} to make it thread safe.
+	 * All methods are wrapped within a {@link ReentrantLock} to ensure that only
+	 * one {@link Thread} can notify the {@link #masterMonitor}.
+	 */
+	private class ThreadSafeProgressMonitor implements IProgressMonitor {
+		private static final ReentrantLock LOCK = new ReentrantLock();
+		private final IProgressMonitor monitor;
+
+		private ThreadSafeProgressMonitor(int ticks) {
+			this.monitor = masterMonitor.newChild(1);
+		}
+
+		@Override
+		public void worked(int ticks) {
+			threadSafe(() -> monitor.worked(ticks));
+		}
+
+		@Override
+		public void internalWorked(double ticks) {
+			threadSafe(() -> monitor.internalWorked(ticks));
+		}
+
+		@Override
+		public void beginTask(String name, int totalWork) {
+			threadSafe(() -> monitor.beginTask(name, totalWork));
+		}
+
+		@Override
+		public void done() {
+			threadSafe(() -> monitor.done());
+		}
+
+		@Override
+		public void setCanceled(boolean value) {
+			threadSafe(() -> monitor.setCanceled(value));
+		}
+
+		@Override
+		public void setTaskName(String name) {
+			threadSafe(() -> monitor.setTaskName(name));
+		}
+
+		@Override
+		public void subTask(String name) {
+			threadSafe(() -> monitor.subTask(name));
+
+		}
+
+		@Override
+		public boolean isCanceled() {
+			return threadSafe(() -> monitor.isCanceled());
+		}
+
+		private void threadSafe(Runnable runnable) {
+			threadSafe(() -> {
+				runnable.run();
+				return null;
+			});
+		}
+
+		private <T> T threadSafe(Supplier<T> supplier) {
+			LOCK.lock();
+			try {
+				return supplier.get();
+			} finally {
+				LOCK.unlock();
+			}
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


vogella:
Not sure what is causing this, maybe JDT is re-deploying?

org.apache.maven.InternalErrorException: Internal error: java.lang.RuntimeException: org.apache.maven.plugin.PluginResolutionException: Plugin org.eclipse.tycho:tycho-compiler-plugin:4.0.10 or one of its dependencies could not be resolved:
	Could not find artifact org.eclipse.jdt:ecj:jar:3.41.0-SNAPSHOT in eclipse-snapshots (https://repo.eclipse.org/content/repositories/eclipse-snapshots/)

vogella:
Reported here https://github.com/eclipse-platform/eclipse.platform.swt/issues/1683

ptziegler:
But Equinox is using Tycho 4.0.10... Why is a _release_ depending on a _snapshot_? That doesn't look right to me.

merks:
@ptziegler 

This is cleaner indeed and also fixes the problem.  (With Oomph one can combine the project setup for p2 and Oomph itself to have a development environment where I can easy test changes to both.)

It looks like some build infrastructure problems kicked in...

merks:
FYI, this version of ecj was just build a short while ago:

https://repo.eclipse.org/content/repositories/eclipse-snapshots/org/eclipse/jdt/ecj/3.41.0-SNAPSHOT/

So while the build says it can't be found, you can see that it really does exist there...

merks:
Don't bother forcing new builds until the infrastructure problem is fixed.  

Thanks for your efforts so far! 

I will push appropriate buttons to restart builds when things look to be in better shape later in the day (I hope).

ptziegler:
> Don't bother forcing new builds until the infrastructure problem is fixed.

I simply noticed that line 933 was still referencing `monitor`, not `subMonitor`, which I just updated. 😅

merks:
Oh. Fixing is allowed.  🤣

merks:
I created this issue:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5454

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/DownloadJob.java
```diff
@@ -56,8 +56,7 @@ protected IStatus run(IProgressMonitor jobMonitor) {
 			if (masterMonitor.isCanceled())
 				return Status.CANCEL_STATUS;
 			// process the actual request
-			SubMonitor subMonitor = SubMonitor.convert(masterMonitor.slice(1), 1);
-			IStatus status = repository.getArtifact(request, subMonitor);
+			IStatus status = repository.getArtifact(request, masterMonitor.newChild(1));
```
laeubi, line 59: Please note that submonitor is not threadsafe! Also now the job monitor do not recive any updates.

It might therefore be better to pass in an own implementation that:

1. syncronize on the master before update it
2. also send all events to the job monitor

merks: I'm trying to understand the nature of the comment.  Even before the change that introduced problem (call to slice) it was a submonitor already then too:

image

Was your comment intended for the change in the other file?

https://github.com/eclipse-equinox/p2/pull/586/files#diff-9a34bd9b50e863d2faa5988b40185db6a3a7e144ef93175808e9e44c19394840

Note that I did test that I see logged information in the installer with these changes, so I'm not sure why the job monitor is not receiving updates...

Maybe we shouldn't be changing this at all?   It's very surprising to me that every change has a different gotcha.  It's still not clear if anything was actually improved...

ptziegler: Note that the code I changed was a botched migration to the SubMonitor, which started this whole ordeal. Originally, the now deprecated SubProgressMonitor was used.

image

https://github.com/eclipse-equinox/p2/commit/9009085b72e9507fdc4b7d0adb87883adefc5637#diff-98743f36a52a7f919e0b5cb2ce651e1a89f1f3a7abf1c343e58a29ee9340e2b5

> syncronize on the master before update it

Without synchronization, some ticks might indeed get lost due to parallelism. But I wonder if this is really worth the additional effort... `done()` is explicitly called on the SubMonitor, so any remaining ticks will be consumed once the download is done, so it's at least not a critical problem.

> also send all events to the job monitor

Wouldn't this effectively mean both jobs show the same task?

merks: If the assertion "The subMonitor created in the run() method is only used on this job's thread, i.e., on the thread that is calling run()" a true assertion?  I thought was was the case, but I've been known to be wrong.

I would have though only the thread safety of the masterMonitor is a potential concern...

ptziegler:
I've created a thread-safe implementation of the IProgressMonitor which should synchronize with the master monitor. As mentioned, the jobMonitor is not notified as this was never the case before.

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/DownloadJob.java
laeubi: @ptziegler has correctly explained it sorry for being late.

One should also keep in mind that newChild/split clear the work acquired by a previous call, so if you do:

```
IProgressMonitor m1 = monitor.split(10);
IProgressMonitor m2 = monitor.split(10);
```

then you have one Monitor (`m1`) that is already done, so only the last job started could report actual progress!

Reporting good and consistent progress is a very hard topic indeed and Eclipse makes its even harder some times due to historical reasons.

The "slice" was introduced by me a while ago specifically for the case where one do work in parallel Threads (usually streams or jobs) that still perform a common task but want to report overall progress. Adapted to this situation here for example one would have one (of unknown origin) Monitor and set its task to "Downloading files", and the amount of work to the number of files. Now I start a thread for each file and pass it a slice of one and this thread can now use safely the monitor (but needs to call done possibly) to report its overall download progress while everything is aggregated into the original one, that might observe the number of running Threads and update the message (e.g. with "Downloading files 5 / 10").

Another reason for slice is to notify the main monitor about that progress was distributed (what is impossible with SubMonitor at the moment) so an UI can show something like this:

grafik
 _Image from [here](https://stackoverflow.com/questions/2270863/displaying-progressbar-for-file-upload/2311153)_

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/DownloadJob.java
```diff
@@ -56,8 +58,7 @@ protected IStatus run(IProgressMonitor jobMonitor) {
 			if (masterMonitor.isCanceled())
 				return Status.CANCEL_STATUS;
 			// process the actual request
-			SubMonitor subMonitor = SubMonitor.convert(masterMonitor.slice(1), 1);
-			IStatus status = repository.getArtifact(request, subMonitor);
+			IStatus status = repository.getArtifact(request, new ThreadSafeProgressMonitor(1));
```
laeubi, line 61: I though that there are multiple jobs, so doe not need the "master monitor" already be thread safe?!?

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/SimpleArtifactRepository.java
```diff
@@ -914,11 +914,12 @@ public IStatus getArtifacts(IArtifactRequest[] requests, IProgressMonitor monito
...
 		} else {
 			// initialize the various jobs needed to process the get artifact requests
-			monitor.beginTask(NLS.bind(Messages.sar_downloading, Integer.toString(requests.length)), requests.length);
+			SubMonitor subMonitor = SubMonitor.convert(monitor,
+					NLS.bind(Messages.sar_downloading, Integer.toString(requests.length)), requests.length);
 			try {
 				DownloadJob jobs[] = new DownloadJob[numberOfJobs];
 				for (int i = 0; i < numberOfJobs; i++) {
-					jobs[i] = new DownloadJob(Messages.sar_downloadJobName + i, this, requestsPending, monitor,
+					jobs[i] = new DownloadJob(Messages.sar_downloadJobName + i, this, requestsPending, subMonitor,
```
merks, line 922: As I understand it, the master monitor is now created here and has become a submonitor.  So it this a/the threading concern?

bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/DownloadJob.java
merks: Yes, this is why I asked the question I asked yesterday.  The monitor here is used only in this thread I believe.

laeubi:
@ptziegler @merks I have now created a PR that I think solves the problem more generally and has the advantage that individual download jobs report proper progress and messages:

- https://github.com/eclipse-equinox/p2/pull/587

merks:
Superseded by https://github.com/eclipse-equinox/p2/pull/587

PR #589, opened 2025-01-04
Create reproducible properties files

Zlika:
There is an ongoing effort to make Tycho produce reproducible artifacts. Some non-reproducible bits come from equinox-p2. This PR makes the properties files created by equinox-p2 reproducible (remove the timestamp and sort the lines).

### DIFF
--- a/bundles/org.eclipse.equinox.frameworkadmin.equinox/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.frameworkadmin.equinox/META-INF/MANIFEST.MF
@@ -17,7 +17,8 @@ Import-Package: org.eclipse.equinox.frameworkadmin;version="[2.0.0,3.0.0)",
  org.osgi.framework.startlevel;version="1.0.0",
  org.osgi.service.log;version="1.3.0",
  org.osgi.service.startlevel;version="1.0.0",
- org.osgi.util.tracker;version="1.3.0"
+ org.osgi.util.tracker;version="1.3.0",
+ org.eclipse.equinox.internal.p2.core.helpers
 Export-Package: org.eclipse.equinox.internal.frameworkadmin.equinox;x-friends:="org.eclipse.equinox.p2.publisher,org.eclipse.equinox.p2.publisher.eclipse,org.eclipse.equinox.simpleconfigurator.manipulator",
  org.eclipse.equinox.internal.frameworkadmin.equinox.utils;x-friends:="org.eclipse.equinox.p2.publisher.eclipse"
 Require-Bundle: org.eclipse.equinox.common;bundle-version="3.18.0"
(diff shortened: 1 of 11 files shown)


merks:
I think adding dependencies for a utility isn't a great idea for something like this:

image

I also wonder if all these places really even affect a build and I wonder if the loss of information such as the time of creation is desired in all these places.  How does the above file specifically affect the build result?  In general can you elaborate on why all these locations need reproducible properties files?

Zlika:
Hello @merks . I didn't find a good (better) location for such utility class. Is there a better location for such common code?
I don't know if all the places affect the build. I'm sure that this location affects the build: EquinoxFwConfigFileParser.saveProperties(). It is called during the materialize-products goal of the tycho-p2-director-plugin plugin and creates a config.ini file in the generated product, which is not reproducible Properties file.
I don't know if other parts of the equinox-p2 code generate properties files during the tycho build, so the simplest solution is to fix the problem everywhere. This timestamp is always useless and only prevents a reproducible behavior: if you really need to know the creation time of files, why would you put a timestamp only inside a couple of properties files and not inside all the files you create? It's just a bad design of java.util.Properties.

laeubi:
Sometimes its better to copy some util code instead of create new dependencies. Also minimal changes are often easier to merge, so maybe just concentrate on the required part. Sometimes in the past we even have copied code from P2 to easier maintain requirements of Tycho vs P2.

That said, I also noticed that if one installs something with P2 the order of `<productname>.ini` changes, I this case I would better like to retain ordering instead of alphabetic ordering though...

merks:
Yes, I'm not sure where is a better place.    Making copies is also not pretty.  Somewhere higher up in equinox would make it more reusable.  I wonder if any of these locations produce results that appears in a build:

image

I did some hunting in a downloaded Eclipse SDK zip and there are quite some files with timestamps in them...

laeubi:
`equinox.common` seems a good place maybe and from there it could be reused by Tycho as well on the next release cycle (of course also from P2...). Because of that I think it would be good to maybe sync the impl with the current Tycho one that already allows to get rid of some boilerplate code when writing properties as well.

merks:
Oh, did you know the following exists, trying to solve the same problem.

[org.eclipse.core.internal.preferences.SortedProperties.entrySet()](https://github.com/eclipse-equinox/equinox/blob/91eebf3abb2f678a4ab9d25a88d2087f63100ab2/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/SortedProperties.java#L38-L51)

Note the big comment in Java 21 where these things are saved:

image

That comment makes me suspicious about the problem we are solving which appears to be not a problem for Java 21.

In Java 17 it is/was like this:

image

For this the override works.

Maybe it's okay to say reproducible builds require to build using Java 21 and we simply change thing?

merks:
There's even this for the date thing using `java.properties.date`:

image

merks:
The `java.properties.date` is documented in:

https://docs.oracle.com/en/java/javase/21/docs/api/system-properties.html

The sorted order is documented here:

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Properties.html

image

So my take is that if you use the existing frameworks as currently implement running Tycho/Maven with on Java 21, that you will get the reproducibility needed without needing to make changes in these frameworks.

Zlika:
Ok, let's cancel this PR.

merks:
Thank you nevertheless for your efforts.

PR #590, opened 2025-01-05
Add an ArtifactManager to allows unified managing of artifacts

laeubi:
Caching can greatly improve performance, basically P2 has two concepts
of where caching can occur, the transport and artifacts that can
originate from different sources. Currently there is only a CacheManger
for handling the first case and nothing for the second, maybe even more
important case. Because of this there are different approaches out in
the wild e.g. PDE Bundle Pools, Oomph shared P2 pools, and Tycho is
using its own cache in the m2 repository, but all of them suffer from
the fact that there is no real convenient entry point so different
things needs to be enhanced an managed.

This now adds a new abstraction service the ArtifactManager that allows
to intercept the process of downloading a logical artifact from a given
location and possibly providing an equivalent one that is already
present locally. There is also a default implementation that simply
delegates to the transport directly as before.

This also includes a commit from 
- https://github.com/eclipse-equinox/p2/pull/483

to add a component property type for the service name that is then used here.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/spi/AgentServiceName.java
+++ b/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/spi/AgentServiceName.java
@@ -23,14 +23,14 @@ import org.osgi.service.component.annotations.ComponentPropertyType;
 /**
  * This component property type can be used to annotate a declarative service
  * component that provides an {@link IAgentServiceFactory} to provides the
- * required {@link IAgentServiceFactory#PROP_CREATED_SERVICE_NAME}.
+ * required {@link IAgentServiceFactory#PROP_AGENT_SERVICE_NAME}.
  *
  * @since 2.13
  */
 @Retention(CLASS)
 @Target(TYPE)
 @ComponentPropertyType
-public @interface AgentServicename {
+public @interface AgentServiceName {
 
 	public static final String PREFIX_ = "p2."; //$NON-NLS-1$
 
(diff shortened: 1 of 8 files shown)


merks (approved):
Looks generally fine.  I've not looked in great details.

bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/p2/core/spi/AgentServiceName.java
```diff
@@ -23,14 +23,14 @@
...
- * required {@link IAgentServiceFactory#PROP_CREATED_SERVICE_NAME}.
+ * required {@link IAgentServiceFactory#PROP_AGENT_SERVICE_NAME}.
  *
  * @since 2.13
  */
 @Retention(CLASS)
 @Target(TYPE)
 @ComponentPropertyType
-public @interface AgentServicename {
+public @interface AgentServiceName {
```
HannesWell, line 33: I think the change is reasonable, I just wanted to mention that renamings that just change the case of some letters in a file name cause funny side-effects on Windows because the file-system is case-insensitive and then git gets confused.
But one just has to get along with that.

laeubi: As the class is actually new (I just renamed it in the PR for the sake of testing) it hopefully do not confuses windows to much :-)

merks: I appreciate the improved name.  It looks like a typo before.  😕

PR #591, opened 2025-01-05
Remove references to all of ECF's ssl bundles and features

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2725

### DIFF
--- a/features/org.eclipse.equinox.p2.core.feature/feature.xml
+++ b/features/org.eclipse.equinox.p2.core.feature/feature.xml
@@ -21,11 +21,9 @@
 
    <requires>
       <import feature="org.eclipse.ecf.core.feature" version="1.4.0" match="compatible"/>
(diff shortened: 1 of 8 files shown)


merks:
BTW, `org.eclipse.ecf.provider.filetransfer.httpclient.ssl` was removed from all products, but that hasn't existed for a very long time:

image

That makes me wonder about how meaningful/useful these products are.

merks:
@akurtakov 

We could try to delete the products folder, but that begs the next question about the examples which are probably only there for the products and hence pointless without the products.

image

merks:
I think the example bundles can all go too:

image

They're mentioned as friends and in some *.launch configurations.

And there appears to be come crappy, old documentation, but that's probably not the only crappy old documentation.

merks:
The failure looks like an unrelated glitch and the same tests passed for all the GitHub actions:

image

PR #592, opened 2025-01-05
Remove all example bundles and products and references to them

merks:


vogella:
+1, I also wanted to do this in the past but feared that this was not desired.

@merks please add the "why are the examples removed" explanation to the commit message.

merks:
Merging now because only the commit message has changed.

PR #594, opened 2025-01-05
Migration to more consistent property name for provided service

laeubi:
Currently we use the property `p2.agent.servicename` to identify the provided service, but the constant is named `PROP_CREATED_SERVICE_NAME`. So beside the inconsistency of created <> agent it shows that the "name" part should actually be an own segment (like in AgentServiceFactory). This also affects the naming of component property types

Because of this we introduce here a new constant `PROP_AGENT_SERVICE_NAME` with the value 'p2.agent.service.name' and deprecate the old constant while remain support for both properties.


merks (approved):
That's nicer.

laeubi:
@HannesWell I'm a bit confused why `org.eclipse.equinox.p2.tests.ui` should require any version bumps here?!?

laeubi:
It seems for whatever reason the bump is required as also this fails here:

https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-2728/1/console

so I'll include the version bump with this PR even though it looks unrelated.

PR #596, opened 2025-01-05
Add an ArtifactManager to allows unified managing of artifacts

laeubi:
Caching can greatly improve performance, basically P2 has two concepts of where caching can occur, the transport and artifacts that can originate from different sources. Currently there is only a CacheManger for handling the first case and nothing for the second, maybe even more important case. Because of this there are different approaches out in the wild e.g. PDE Bundle Pools, Oomph shared P2 pools, and Tycho is using its own cache in the m2 repository, but all of them suffer from the fact that there is no real convenient entry point so different things needs to be enhanced an managed.

This now adds a new abstraction service the ArtifactManager that allows to intercept the process of downloading a logical artifact from a given location and possibly providing an equivalent one that is already present locally. There is also a default implementation that simply delegates to the transport directly as before,

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/META-INF/MANIFEST.MF
@@ -69,7 +69,8 @@ Import-Package: javax.xml.parsers,
  org.osgi.service.prefs;version="1.1.1",
  org.w3c.dom,
  org.xml.sax;resolution:=optional
-Service-Component: OSGI-INF/org.eclipse.equinox.p2.artifact.repository.xml
+Service-Component: OSGI-INF/org.eclipse.equinox.internal.p2.artifact.repository.DefaultArtifactManagerServiceFactory.xml,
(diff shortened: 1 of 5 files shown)


mickaelistria:
That seems interesting. Can you please link to some draft PRs (in PDE, or Tycho, or whatever) showcasing how such a new service could be leveraged?
Would it be possible to consider a smarter implementation even directly in p2?

laeubi:
Tycho for example has [MirroringArtifactProvider](https://github.com/eclipse-tycho/tycho/blob/39c7c2f633d5f8372bdb250c8fa17272acebdf10/tycho-core/src/main/java/org/eclipse/tycho/p2/repository/MirroringArtifactProvider.java) where we put a lot of effort and require special setup to have a cache of artifacts what could better be implemented in this way. PDE has similar with its bundle pool where it cache already fetched artifacts.

> Would it be possible to consider a smarter implementation even directly in p2?

Everything is possible, currently there is no P2 intrinsic caching mechanism for artifacts. I just found that everyone seems to have slightly different demands on how caching works, so the most challenging would be to define a common cache infrastructure that fits all use-cases.

PR #597, opened 2025-01-05
Migrate ArtifactRepositoryComponent to DS annotations and property type

laeubi:


laeubi:
There must be something wrong on windows that I don't see :-\

laeubi:
If we use the generated XML by PDE everything is fine, so this seems to be an issue with bnd ds generator :-\

PR #602, opened 2025-01-08
Replace global waiting flag by local latch, remove browseRequest field

laeubi:
Currently there is a class instance shared field used for final waiting but actually it could be handled locally in the method, this also removes the need for keeping the browseRequest around.


vogella (commented):
Nit: Can't we replace LogHelper with ILog?

laeubi:
> Nit: Can't we replace LogHelper with ILog?

Possible yes... currently I just move things around to get rid of instance fields and reduce the complexity of the class.

PR #603, opened 2025-01-08
Inline eventhandling and remove need for exception field

laeubi:
Currently error handling is performed with a shared class field on two different places.

This now inline the event handling in a way it does not need a global field and handles the exception locally.


vogella:
Thanks for your work @laeubi 

It is impressive to watch how many changes needs / can be done in this single file.

laeubi:
The class was quite convoluted, so I decided that many smaller PRs are better than one that try to fix them all :-)

vogella:
> The class was quite convoluted, so I decided that many smaller PRs are better than one that try to fix them all :-)

Totally my style too to do small changes incremental, so +1 for this approach

PR #606, opened 2025-01-09
Only format edited lines in p2.transport.ecf

vogella:


vogella:
The windows test errors look unrelated to the save action change.

PR #611, opened 2025-01-14
Test new Api tools in verification build

laeubi:
This should show two errors in p2.core (and maybe more)

grafik


laeubi:
Seems to work :+1: 

grafik

PR #613, opened 2025-01-16
Inner class GradientInfo can be static

vogella:


laeubi:
@vogella are you fine with holding this PR back for a while I'd like to use it for some testing of automation if you don't mind!

laeubi:
The workflow is online:

- https://github.com/eclipse-equinox/p2/pull/643

and should soon push an automatic PR for this change!

laeubi:
The automatic created code cleanup PR is now here, closing this one:
- https://github.com/eclipse-equinox/p2/pull/650

vogella:
Nice to see this automatic workflow enhancement, still merging this one, as we typically merge the first PR which suggest a change.

PR #614, opened 2025-01-18
Add workflow that checks for bundle dependencies version ranges

laeubi:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/.github/workflows#checkdependenciesyml


laeubi:
CI checks wont verify this.

PR #642, opened 2025-01-21
Record Package Attributes and Directives on exported packages

laeubi:
Currently P2 lacks support for package attributes and directives that can lead to wrong selection of providers. One issue is that the information on exported packages are not recorded at all in the P2 metadata and therefore can not be used in slicer or planner operations.

As a first step this now records the information on the provider side by transforming attributes and directives into properties of the package capability.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
+++ b/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
@@ -31,6 +31,7 @@ import org.eclipse.core.runtime.*;
 import org.eclipse.equinox.frameworkadmin.BundleInfo;
 import org.eclipse.equinox.internal.p2.core.helpers.LogHelper;
 import org.eclipse.equinox.internal.p2.metadata.ArtifactKey;
+import org.eclipse.equinox.internal.p2.metadata.ProvidedCapability;
 import org.eclipse.equinox.internal.p2.publisher.Messages;
 import org.eclipse.equinox.internal.p2.publisher.eclipse.GeneratorBundleInfo;
 import org.eclipse.equinox.internal.p2.publisher.eclipse.bundledescription.BundleDescriptionBuilder;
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
```diff
@@ -1151,4 +1170,51 @@ private void addPublishingErrorToFinalStatus(Throwable t, File bundleLocation) {
...
+				}
+			} else if (isValidValueType(prop)) {
+				target.put(prop, value);
+			}
+		}
+	}
+
+	private static boolean isValidValueType(Object prop) {
+		if (prop instanceof List<?>) {
+			int idx = 0;
```
merks, line 1200: This variable is not used.

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
```diff
@@ -1151,4 +1170,51 @@ private void addPublishingErrorToFinalStatus(Throwable t, File bundleLocation) {
...
+			return true;
+		}
+		return isValidScalarType(prop);
+	}
+
+	private static boolean isValidScalarType(Object scalar) {
+		Class<?> clazz = scalar.getClass();
+		Stream<Class<?>> supportedClasses = Stream.of(Version.class, String.class, Long.class, Integer.class,
+				Short.class, Byte.class, Double.class, Float.class, Boolean.class, Character.class);
+		if (supportedClasses.noneMatch(t -> t.isAssignableFrom(clazz))) {
```
merks, line 1215: We could just return the value of the condition.

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
```diff
@@ -1151,4 +1170,51 @@ private void addPublishingErrorToFinalStatus(Throwable t, File bundleLocation) {
...
+				}
+			}
+			return true;
+		}
+		return isValidScalarType(prop);
+	}
+
+	private static boolean isValidScalarType(Object scalar) {
+		Class<?> clazz = scalar.getClass();
+		Stream<Class<?>> supportedClasses = Stream.of(Version.class, String.class, Long.class, Integer.class,
```
merks, line 1213: Might it be better to create a static final constant constant for the list?

merks (changes requested):
I'm slightly uncomfortable recording data for which we have no explicit planned uses at this point.  That being said, other than a bit of bloat in the content metadata, I don't see any harm and could image this information being useful in some way in the future.

When testing manually I see things like uses recorded:

image

The toString is slightly ugly, but that's always been that way.

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
mickaelistria: Streams are usually not reusable since they have "terminal" operations closing them. You'd get a `java.lang.IllegalStateException: stream has already been operated upon or closed`.
Using Streams as storage is a bad idea; Collections are better for it. Directly using stream is better when processing the stream immediately, when there is no chance of it being referenced anywhere else (because it's not accessible or because same expression would lead to a new stream instance); but here it's not the case.

laeubi:
> When testing manually I see things like uses recorded

uses is an standard directive and actually something very important that P2 currently also not take into account.

merks:
> > When testing manually I see things like uses recorded
> 
> uses is an standard directive and actually something very important that P2 currently also not take into account.

Yes, I should have been clear that this seems potentially useful.  😁

laeubi:
Jenkins seem to have timing issues, github is fine.

PR #686, opened 2025-02-01
[Build] Reuse 'releaseVersion' property from eclipse-platform-parent

HannesWell:
Don't redefine the `releaseVersion` from the `eclipse-platform-parent/pom.xml` with a slightly different value.

This reduces the number of required changes in the preparation of new release-cycles and unblocks
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2799

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,10 +26,9 @@
 
   <properties>
     <tycho.scmUrl>scm:git:https://github.com/eclipse-equinox/p2.git</tycho.scmUrl>
-    <releaseVersion>4.35.0</releaseVersion>
     <!-- Defines the default Qualifier if no format is given-->
     <!-- can be overriden with -Dtycho.buildqualifier.format=yyyyMMddHHmm for a dynamic qualifier or -Dqualifier=abcd for a static value -->
-    <qualifier>-SNAPSHOT</qualifier>
+    <qualifier>.0-SNAPSHOT</qualifier>
     <!-- disabled for now, should be enabled later! -->
     <addMavenDescriptor>false</addMavenDescriptor>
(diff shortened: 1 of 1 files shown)


laeubi:
The variable in the pom is for [CI friendly versions](https://maven.apache.org/guides/mini/guide-maven-ci-friendly.html) to work correctly here (I'm a bit surprised that inherit the value from the parent works) and it is only allowed to contain a constant string (at least for Maven 3.9).

I'm a bit confused why the build succeeds anyways, but we should simply add the constant to the qualifier part `<qualifier>.0-SNAPSHOT</qualifier>` or use one of the other CI variables for example `<revision>.0</revision>` so the version becomes `<version>${releaseVersion}${revision}${qualifier}</version>`

HannesWell:
> The variable in the pom is for [CI friendly versions](https://maven.apache.org/guides/mini/guide-maven-ci-friendly.html) to work correctly here (I'm a bit surprised that inherit the value from the parent works) and it is only allowed to contain a constant string (at least for Maven 3.9).

The build here is running with maven `3.9.9`. I assumed that it doesn't matter where the variable is defined, but I have also not looked into that topic deeply. I'm just happy that it works :)

>  but we should simply add the constant to the qualifier part `<qualifier>.0-SNAPSHOT</qualifier>` 

Sure that makes the change even simpler. Applied it.

HannesWell:
Btw. I checked the logs and the resolved version is as expected:
```
[INFO] -----------------< org.eclipse.platform:rt.equinox.p2 >-----------------
[INFO] Building rt.equinox.p2 4.35.0-SNAPSHOT                            [1/52]
```

PR #688, opened 2025-02-02
Add BundlesAction.createBundleIU(File) for easy creation of IU

laeubi:
It is often useful to get some information about a file and BundlesAction is the P2 way to gather such data, but at the moment it is really cumbersome and requires a lot of calls and handling of cases to get to that point.

This now adds a new method BundlesAction.createBundleIU(File) that ease this task.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
+++ b/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
@@ -201,6 +201,30 @@ public class BundlesAction extends AbstractPublisherAction {
 		return MetadataFactory.createInstallableUnit(cu);
 	}
 
+	/**
+	 * Attempts to read the given file (or folder) as a bundle and creates an
+	 * {@link IInstallableUnit} describing it.
+	 * 
+	 * @param file the file to read as a bundle
+	 * @return an {@link Optional} describing the result, if anything goes wrong an
+	 *         empty optional is returned
+	 */
+	public static Optional<IInstallableUnit> createBundleIU(File file) {
+		try {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
```diff
@@ -201,6 +201,30 @@ static IInstallableUnit createBundleConfigurationUnit(String hostId, Version cuV
...
+			BundleDescription bundleDescription = BundlesAction.createBundleDescription(file);
+			if (bundleDescription != null) {
+				IArtifactKey key = BundlesAction.createBundleArtifactKey(bundleDescription.getSymbolicName(),
+						bundleDescription.getVersion().toString());
+				PublisherInfo publisherInfo = new PublisherInfo();
+				publisherInfo.setArtifactOptions(IPublisherInfo.A_INDEX);
+				return Optional.ofNullable(BundlesAction.createBundleIU(bundleDescription, key, publisherInfo));
+			}
+		} catch (BundleException | IOException | RuntimeException e) {
+			return Optional.empty();
```
merks, line 223: I guess you prefer to return the same result as line 225 to avoid just an empty block or an empty block with a comment?

merks (approved):
Looks okay.

bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
laeubi: Yes especially for debugging an empty block is annoying as you can't set a breakpoints there (as far as I know) so I thought it would be good just in case one unexpectedly gets an empty optional to have these two cases.

PR #689, opened 2025-02-12
Ensure that AbstractRepository.properties cannot be replaced

merks:
- Make AbstractRepository.properties final.
- Modify AbstractRepository.setProperties to do a clear and a putAll.
- Make the only method that accesses AbstractRepository.properties without synchronization, AbstractRepository.getProperty(String), also be synchronized.

https://github.com/eclipse-tycho/tycho/issues/4709


laeubi (approved):
Looks good :+1:

PR #694, opened 2025-03-03
Update for release 4.36

MohananRahul:


akurtakov:
Standalone build will not work until first 4.36 I-build.

PR #736, opened 2025-03-07
Improve Double-Checked Locking (DCL) for variable *session*.

JVS94:
The previous code does not ensure a singleton instance of the variable `session`.


merks (commented):
Looks correct to me.  

@laeubi   What do you think?

iloveeclipse (approved):

laeubi (commented):
I think it would be better to simply make the method synchronized, I don't expect this to be a call time hotspots that any "tricks" will help here.

JVS94:
It is also acceptable to change directly to a synchronized method.

bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/src/org/eclipse/equinox/internal/p2/ui/sdk/scheduler/AutomaticUpdatePlugin.java
```diff
@@ -184,11 +184,12 @@ private void createImageDescriptor(String id, ImageRegistry reg) {
 		reg.put(id, desc);
 	}
 
-	public ProvisioningSession getSession() {
+	public synchronized ProvisioningSession getSession() {
 		if (session == null) {
-			synchronized (this) {
-				IProvisioningAgent agent = ServiceHelper.getService(getContext(), IProvisioningAgent.class);
-				session = new ProvisioningSession(agent);
+				if (session == null) {
```
laeubi, line 189: this is checked twice now

merks:
Please squash to a single commit. Merge commits are not accepted.

PR #784, opened 2025-04-02
VersionParser: make Exception message more meaningful

chrisrueger:
This PR helps debugging when there was an invalid version string passed.

I wasted hours trying to figure out a failing build because an invalid version string was passed in a <plugin> definition of a p2 feature.xml But the Exception unfortunatelly does not tell what the wrong string was.

## Example

```
BUILD FAILED
java.lang.IllegalArgumentException
        at org.eclipse.equinox.internal.provisional.p2.core.VersionParser.parseInto(VersionParser.java:71)
        at org.eclipse.equinox.internal.provisional.p2.core.Version.create(Version.java:138)
        at org.eclipse.equinox.internal.provisional.p2.core.Version.parseVersion(Version.java:162)
        at org.eclipse.equinox.p2.publisher.eclipse.FeatureEntry.<init>(FeatureEntry.java:51)
        at org.eclipse.equinox.internal.p2.publisher.eclipse.FeatureManifestParser.processPlugin(FeatureManifestParser.java:211)
        at org.eclipse.equinox.internal.p2.publisher.eclipse.FeatureManifestParser.startElement(FeatureManifestParser.java:244)
```

because a feature.xml contained `version="@org.slf4j.api-version@"` which was because of another bug where a token replacement failed. 

```
<plugin
         id="org.slf4j.api"
         download-size="0"
         install-size="0"
         version="@org.slf4j.api-version@"
         unpack="false"/>
```

But it was very hard find this needle in the haystack.

After the change the Exception would have a message like this:

`Expected letter at position 0 in version: @org.slf4j.api-version@`

instead of no message. 
With this information I can then do a fulltext search in my codebase for the string `@org.slf4j.api-version@` to find the file to fix the problem.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/VersionParser.java
+++ b/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/VersionParser.java
@@ -83,7 +83,7 @@ public abstract class VersionParser {
 		}
 
 		if (!isLetter(c)) {
-			throw new IllegalArgumentException();
+			throw new IllegalArgumentException(NLS.bind(Messages.expected_letter_at_position, pos, version));
 		}
 
 		if (version.startsWith(Version.RAW_PREFIX, pos)) {
--- a/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/Messages.java
+++ b/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/Messages.java
@@ -55,6 +55,8 @@ public class Messages extends NLS {
 
 	public static String enum_defined_more_then_once;
 
+	public static String expected_letter_at_position;
+	
 	public static String expected_orignal_after_colon_0;
 
 	public static String expected_orignal_after_slash_0;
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/VersionParser.java
```diff
@@ -83,7 +83,7 @@ public static Version parse(String version, int start, int maxPos) throws Illega
 		}
 
 		if (!isLetter(c)) {
-			throw new IllegalArgumentException();
+			throw new IllegalArgumentException("Expected letter at position " + pos + " in version: " + version); //$NON-NLS-1$
```
laeubi, line 86: I know it might is a bit odd but as we have other exception messages localized, we should do this here as well (see for example line 131)

laeubi (commented):
This really seems to be missed as other cases have meaningful messages. For consistency reasons I just think the message should be externalized.

bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/VersionParser.java
chrisrueger: Done

laeubi (approved):
Many Thanks!

PR #795, opened 2025-05-03
Simplify URI to string conversion and avoid URIUtil.toURL()

HannesWell:
Calling `new URL(uri.toString()).toExternalForm())` instead of `uri.toString()` does not seem to have a benefit.

And `BlobStore.fileFor()` should return an absolute URI and therefore `URI.toURL()` should be save too.


merks (approved):
Goodness knows what problem all that complexity was supposed to address

HannesWell:
I see (what seems to be) the same test-failures on Windows on the master, so I think we can assume they are unrelated.

PR #797, opened 2025-05-05
Add SVGs to org.eclipse.equinox.p2.ui bundles

Michael5601:
This PR adds SVGs for all icons in the bundles `org.eclipse.equinox.p2.ui`, `org.eclipse.equinox.p2.ui.admin`,
`org.eclipse.equinox.p2.ui.admin.rcp`,
`org.eclipse.equinox.p2.ui.discovery`,
`org.eclipse.equinox.p2.ui.importexport`,
`org.eclipse.equinox.p2.ui.sdk` and `org.eclipse.equinox.p2.ui.sdk.scheduler` except for the following as it is not available as SVG yet:

org.eclipse.equinox.p2.ui.importexport/icons/obj16/install-handler.svg
______________________________________________________________________
I deleted the constants `T_TOOL`, `FIND_CLEAR` and `FIND_CLEAR_DISABLED` in class `org.eclipse.equinox.internal.p2.ui.discovery.DiscoveryImages` as the specified GIFs do not exist and the constants are not used. The bundle is exported as `x-internal=true`.
_________________________________________________________________________
I changed one icon path `icons/install_wiz.png` in the class  `org.eclipse.equinox.internal.p2.ui.sdk.scheduler.migration.MigrationWizard` to `icons/wizban/install_wiz.svg` as the former path did not lead to any icon in the specified bundle `org.eclipse.equinox.p2.ui`.

See also this [PR](https://github.com/eclipse-platform/eclipse.platform/pull/1797) for more information.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui.admin.rcp/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.ui.admin.rcp/META-INF/MANIFEST.MF
@@ -12,3 +12,4 @@ Require-Bundle: org.eclipse.ui,
 Bundle-ActivationPolicy: lazy
 Export-Package: org.eclipse.equinox.internal.p2.ui.admin.rcp;x-internal:=true
 Automatic-Module-Name: org.eclipse.equinox.p2.ui.admin.rcp
+Require-Capability: eclipse.swt;filter:="(image.format=svg)"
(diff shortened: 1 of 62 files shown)


Michael5601:
@HannesWell It would be very kind of you if you can have a look at this PR and [PR](https://github.com/eclipse-m2e/m2e-core/pull/1985). These are the last two PRs for integrating SVGs (apart from the mylyn bundles that still have problems). It would be great if we could have them in the new Eclipse release.

merks:
Let's just move forward and fix any problems that might show up.  😀

Michael5601:
> Let's just move forward and fix any problems that might show up. 😀

Thank you @merks  :)

PR #824, opened 2025-05-12
Use absolute paths to invoke reg and find

merks:
- Tests are failing depending on what is on the system PATH environment variable, i.e., the script find the wrong find executable, the one from git bash rather than the windows native find.
- Do the same for `reg`.


merks:
The three failing tests for the windows native registry stopped failing.

HannesWell:
Should we try to revert this to see if this is also fixed by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3042?

merks:
It’s a real problem that the script assumes the right things are on the path first.  Launch eclipse from gitbash and this touch point will fail. How many other places make bad assumptions I don’t know but best to fix the assumption than make the environment conform to the assumption.

PR #825, opened 2025-05-12
[Build] Try run mavenBuild-workflow with CMD shell on Windows

HannesWell:
Test if executing the build with CMD under Windows helps to fix the test failures:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3042

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,4 +11,4 @@ on:
 
 jobs:
   build:
-    uses: eclipse-platform/eclipse.platform.releng.aggregator/.github/workflows/mavenBuild.yml@master
+    uses: HannesWell/eclipse.platform.releng.aggregator/.github/workflows/mavenBuild.yml@mavenBuild-win-cmd


merks:
The failure looks like some cleanup that was not properly completed 

> 21:51:41    Missing requirement: org.eclipse.sdk.tests.feature.group 4.36.0.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.jsch.tests 0.0.0' but it could not be found

it’s always a good plan to have and sdk workspace so you see everything when one deletes something.  I suppose the I-build might fail too.

HannesWell:
> The failure looks like some cleanup that was not properly completed

Indeed. Should be fixed with
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3043

HannesWell:
This is not necessary anymore as it confirmed that the P2 build test failures are fixed with
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3042

Thank you for the hint!

PR #863, opened 2025-06-06
Replace ECF feature imports with bundle imports

merks:
- Only import the bundles that are not required by any p2 bundle, i.e., the ones that provide the extensions to the underlying framework that are directly required by p2.


manandbytes:
> There is a valid ECA on file for ...

Not sure why it complains in https://api.eclipse.org/git/eca/status/gh/eclipse-equinox/p2/863 :thinking:

merks:
Yes, the ECA is broken.

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6246

You can safely ignore it because you can see in my profile that I have many commits:

image

Issue #570, opened 2024-11-19
[Ubuntu] "Always Trust Everything Confirmation" window is hidden


iuliavasii:
Environment:
- Ubuntu 20.04
- gtk 3.24.20
- Eclipse 2023.12

Steps to reproduce:
- try to install any signed/unsigned packages using "Install New Software"
- the wizard will take you to "Do you trust content" window (attached), added here: https://eclipse.dev/eclipse/news/4.28/platform.php#trusted-authorities
- click on "Always trust all authorities" / "Always trust all content"

Note that "Always Trust Everything Confirmation" window (attached) pops-up in the back of the current window, user is not able to click on it and the product becomes unusable.

Notes:
- Using "Select All" instead of "Always trust all authorities" / "Always trust all content" doesn't show any pop-up and therefore the installation finishes successfully.
- I was able to reproduce it in Eclipse 2024-12 as well.
![Screenshot 2024-11-19 165719](https://github.com/user-attachments/assets/6b4bb9ec-a434-4026-b555-4b51c5177977)
![Screenshot 2024-11-19 165828](https://github.com/user-attachments/assets/51561413-ec3c-44b5-ac1b-285685917efe)

merks:
This sounds similar to

akurtakov:
Both Ubuntu and Eclipse versions are way too old. Would you please try whether this happens with latest Ubuntu LTS and Eclipse 2024-09?

iuliavasii:
I've tried these combinations:
- Eclipse 2024-09 on Ubuntu 20.04 -> is reproducible
- Ubuntu 22.04 -> not reproducible (I used Eclipse 2023-12, but I believe the Eclipse version doesn't matter here)

danthe1st:
I am facing something that seems to be the same issue but I can reproduce it on Ubuntu 24.10 with Eclipse 2025-04:

When Eclipse 2025-03 is installed, it currently shows this popup suggesting the installation of Eclipse 2025-06 M1:
![Image](https://github.com/user-attachments/assets/6c2c5800-8de1-41fb-96cc-32d5fecb6869)

When clicking on "Install in this IDE", it opens the Eclipse Updater. When I'm installing the Update on an "Eclipse IDE for java and Enterprise Java Developers" installation, I get to the following:

![Image](https://github.com/user-attachments/assets/133bf334-26ec-47ca-a808-057dcb78242a)

In the background, there is another modal asking about signatures for Wild Web Developer:

[...]

PR #872, opened 2025-06-19
Use SWT.ON_TOP flag for MessageDialog created in TrustCertificateDialog

trancexpress:
When clicking on the "Always trust all authorities" checkbox, with Eclipse running on GTK+, the created MessageDialog is hidden by the "main" dialog of TrustCertificateDialog. After this, the "main" dialog cannot be interacted with using the mouse pointer, giving the impression that Eclipse is frozen and unusable.

This change adds the SWT.ON_TOP flag to MessageDialog dialogs created in TrustCertificateDialog, so that the question dialog appears above the TrustCertificateDialog dialog - that way the question dialog is seen by the user and can be interacted with normally.

Fixes: #570


trancexpress:
@akurtakov , @merks , it would be great if you can review the change.

Ed, could you check the change on Windows? I don't have a Windows machine.

merks:
It looks fine and it will be easier for me to test in further (in Oomph with the installer) when it's merged.

Thanks for investing the effort!!  💟

Issue #873, opened 2025-06-20
Look for alternatives of SWT.ON_TOP in TrustCertificateDialog


trancexpress:
See discussion here: https://github.com/eclipse-equinox/p2/issues/570#issuecomment-2988901110

We should check for alternatives of `SWT.ON_TOP`, since the flag requires extra handling for any further dialog we open from `TrustCertificateDialog`.

merks:
I suggest just trying to eliminate it because I tested that and it works fine for me on Windows.  We don't strictly need the dialog to remain on top of other windows if some other window is given focus but since that too worked fine on Windows I gave it no further thought.  We should change it in both dialogs if this work well on Linux:

![Image](https://github.com/user-attachments/assets/3bf47832-018e-471b-b76a-bc83a67fbd3d)

Thanks again for being proactive!!!

trancexpress:
From my POV we can keep the modal part of the dialog behavior, its useful for it to always be on top. @iloveeclipse ?

PR #874, opened 2025-06-20
Use SWT APPLICATION_MODAL instead of ON_TOP for TrustCertificateDialog

trancexpress:
Fixes: #873

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/TrustCertificateDialog.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/TrustCertificateDialog.java
@@ -82,7 +82,7 @@ public class TrustCertificateDialog extends SelectionDialog {
 	public TrustCertificateDialog(Shell parentShell, Object input) {
 		super(parentShell);
 
-		setShellStyle(SWT.DIALOG_TRIM | SWT.MODELESS | SWT.RESIZE | SWT.MAX | SWT.ON_TOP | getDefaultOrientation());
+		setShellStyle(SWT.DIALOG_TRIM | SWT.RESIZE | SWT.MAX | SWT.APPLICATION_MODAL | getDefaultOrientation());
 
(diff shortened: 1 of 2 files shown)


trancexpress:
@merks could you check the behavior now? On Linux it looks good:

[equinox_p2_gh873_trust_certificates_dialog_application_modal.webm](https://github.com/user-attachments/assets/3d4f8b46-fd5b-4636-90cd-5dc1a9b9edbc)

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/TrustCertificateDialog.java
```diff
@@ -853,9 +854,6 @@ private static class QuestionDialog extends MessageDialog {
 
 		public QuestionDialog(Shell shell, String title, String message, String labelAccept, String labelReject) {
```
trancexpress, line 855: We can also revert the addition of this class from my POV, I added it due to the handling for `SWT.ON_TOP`.

merks:
Could you please 🙏 also change this accordingly:

https://github.com/eclipse-equinox/p2/blob/7afe2522d4dabd0aaf06f290686911367964a1d3/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/TrustAuthorityDialog.java#L86

I'm sure it will have the same problem.  Installing from https://www.ironbloods.at/viplugin/update-site would let you test that, but I have no concerns about that being tested because that to I tested in Windows with the flag removed.

Thank you, thank you.

merks:
@iloveeclipse 

Thank you for questioning the need for ON_TOP!

merks (approved):
I retested on Windows with APPLICATION_MODAL with the installer and that works okay, even if there is a dialog open when these dialogs open.  A tiny nit, I noticed that SWT.MODELESS  is just zero, maybe on all OSes?

merks:
@trancexpress 

You're the best.  🏆

PR #876, opened 2025-06-20
Default implement IRepositoryManager.loadRepository(URI, IProgressMonitor)

HannesWell:
This simplifies the implementations and makes it easier to discover what default value is used for 'flags'.


merks (approved):
Looks good.

PR #881, opened 2025-06-23
Eliminate producing the "manifest" touchpoint data

merks:
- There is no consumer (touch point action) of this information and whatever old clients might have done something 14 years ago are surely long gone.


merks:
Maybe, but I thought first to test if those tests fail or any tests fail.

In the end one is allowed to put in arbitrary touch point information and there is no strict requirement for it be meaningful or consumed...

PR #883, opened 2025-07-12
Don't serialize ITouchpointType.NONE

merks:
- The implementation of InstallableUnit.getTouchpointType() always returns a non-null result returning ITouchpointType.NONE when the underlying field InstallableUnit.touchpointType is null so there is no need to serialize this.
- Also MetadataFactory.createTouchpointType(String, Version) has specialized handling for deserliazing ITouchpointType.NONE such that it makes no difference to deserialize the value versus there being no value specified.


laeubi (approved):
Less is better, but you should check if Tycho has similar like this as we write metadata there as well.

PR #884, opened 2025-07-14
Enable Manifest cleanups for P2 repository

laeubi:
As it worked for Equinox since a while now and P2 also offers a lot of API it seems good to have it enabled here as well.


laeubi:
Build will not test it

PR #923, opened 2025-07-24
ProfileMetadataRepository: surface profile properties as repo properties

merks:
This helps surface the information via existing API:

- org.eclipse.equinox.p2.repository.IRepository.getProperties()
- org.eclipse.equinox.p2.repository.IRepository.getProperty(String)


merks:
For example. the actual profile repository metadata does have properties:
```
<?xml version='1.0' encoding='UTF-8'?>
<?profile version='1.0.0'?>
<profile id='epp.package.committers' timestamp='1751986139744'>
  <properties size='7'>
    <property name='org.eclipse.equinox.p2.cache' value='D:\Users\test39\eclipse'/>
    <property name='org.eclipse.update.install.features' value='true'/>
    <property name='org.eclipse.equinox.p2.roaming' value='true'/>
    <property name='org.eclipse.equinox.p2.environments' value='osgi.os=win32,osgi.arch=x86_64,osgi.ws=win32,osgi.nl=en_US'/>
    <property name='org.eclipse.equinox.p2.installFolder' value='D:\Users\test39\eclipse'/>
    <property name='eclipse.touchpoint.launcherName' value='eclipse'/>
    <property name='org.eclipse.equinox.p2.cache.extensions' value='file:/D:/Users/test39/eclipse/.eclipseextension|file:/D:/Users/test39/eclipse/configuration/org.eclipse.osgi/229/data/listener_1925729951/'/>
  </properties>
```

PR #927, opened 2025-07-27
Remove tycho-source-plugin's feature-source goal

merks:
- With the removal of the  source features from the SDK feature, this is no longer needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,31 +104,6 @@
     <plugins>
       <plugin>
         <groupId>org.eclipse.tycho</groupId>
-        <artifactId>tycho-source-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -102,18 +102,6 @@
...
-        <executions>
-          <execution>
-            <id>feature-source</id>
-            <goals>
-              <goal>feature-source</goal>
-            </goals>
-          </execution>
-        </executions>
-      </plugin>
       <plugin>
```
akurtakov, line 105: attach-p2-metadata is there only for the source feature. Please remove it.

merks: Thanks!

PR #938, opened 2025-08-30
Do not expose unstarted service, stop started service leak

basilevs:
There is a potential for client to see service before they are started.
Dynamically registered services may be never stopped causing resource leaks.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/ProvisioningAgent.java
+++ b/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/ProvisioningAgent.java
@@ -95,9 +95,11 @@ public class ProvisioningAgent implements IProvisioningAgent, ServiceTrackerCust
 	@Override
 	public void registerService(String serviceName, Object service) {
 		checkRunning();
-		agentServices.put(serviceName, service);
 		if (service instanceof IAgentService) {
 			((IAgentService) service).start();
+		}
+		if (agentServices.put(serviceName, service) instanceof IAgentService prevService) {
+			prevService.stop();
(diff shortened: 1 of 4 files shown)


basilevs:
I guess I'll add this to a larger PR.

bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/ProvisioningAgent.java
```diff
@@ -95,10 +95,16 @@ private void checkRunning() {
...
 		checkRunning();
-		agentServices.put(serviceName, service);
 		if (service instanceof IAgentService) {
 			((IAgentService) service).start();
 		}
+		if (agentServices.put(serviceName, service) instanceof IAgentService prevService) {
+			prevService.stop();
+			if (prevService == service) {
+				agentServices.remove(serviceName, prevService);
+				throw new IllegalArgumentException("Double registration"); //$NON-NLS-1$
```
laeubi, line 105: Is it anywhere mentioned that this is not allowed? Also it would be more an illegalstate than an illegal argument and the message should be more descriptive.

laeubi: > If double registration was allowed, IAgentService.start() would be invoked twice for the same object.

This is an implementation detail and can of course be avoided, so the question is if anywhere is documented that registering the same service twice is forbidden (at least it was not before)

basilevs: I've extended the Javadoc.

PR #947, opened 2025-09-17
Don't call pack() on not yet visible table on Linux

iloveeclipse:
Something goes wrong on SWT Column/Table GTK code if column is being packed on not yet shown table. So let pack() the column on a next call on Linux.

Fixes https://github.com/eclipse-equinox/p2/issues/946


laeubi:
> Something goes wrong on SWT Column/Table GTK code if column is being packed on not yet shown table.

Should this not fixed at SWT instead on each call site of pack?

iloveeclipse:
> > Something goes wrong on SWT Column/Table GTK code if column is being packed on not yet shown table.
> 
> Should this not fixed at SWT instead on each call site of pack?

If one knows where and how to fix, for sure. I don't know.

merks:
Yes, I'm not happy to see such changes in client code that worked well in the past.  This doesn't feel right.  How many other place are affected?

merks (commented):
Does line `securedColumn.getColumn().pack()` at line 621 have the same problem when turning to the other tab?

This feels like treating a symptom and hoping the disease has no impact elsewhere.  ☹️

PR #950, opened 2025-09-25
Fix info message using double negation

laeubi:


merks (approved):
Most useless and yet unavoidable log message ever!

PR #968, opened 2025-11-13
Add a P2SSLContextFactory and convert to DS annotations

laeubi:
FYI @merks


merks:
@sratz 

I have tested this this PR has the intended effect.

image

The existing default client works in this case.

merks:
@iloveeclipse 

I kicked another build just to see, but the failure looks a bit like something needs to be touched to avoid a baseline replacement that makes something be missing because an older feature is being used where a newer one is required.

iloveeclipse:
> I kicked another build just to see, but the failure looks a bit like something needs to be touched

No, it looks like infra issue. We have now some other fully unrelated errors.

https://ci.eclipse.org/releng/job/Builds/job/I-build-4.38/119/console

```
[ERROR] Cannot resolve dependencies of project eclipse.platform.releng.tychoeclipsebuilder:eclipse-junit-tests:eclipse-repository:4.38.0-SNAPSHOT
15:05:44  [ERROR]  with context {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, org.eclipse.jdt.buildtime=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true}
15:05:44  [ERROR]   Software being installed: eclipse-junit-tests 4.38.0.qualifier
15:05:44  [ERROR]   Missing requirement: org.eclipse.sdk.tests.feature.group 4.38.100.v20251113-1016 requires 'org.eclipse.equinox.p2.iu; org.eclipse.sdk.tests [4.38.0.v20251113-0510,4.38.0.v20251113-0510]' but it could not be found
15:05:44  [ERROR]   Cannot satisfy dependency: eclipse-junit-tests 4.38.0.qualifier depends on: org.eclipse.equinox.p2.iu; org.eclipse.sdk.tests.feature.group 0.0.0: See log for details
```

laeubi:
> From a technical consistency perspective, I think it should be removed here.

It all depends on how the API contract is defined, due to how ECF currently works I assumed it always wants a initialized instance so the factory is responsible to supply such thing.

laeubi:
> Is that okay/accepted/intended?

No this is an error `@Component(property = Constants.SERVICE_RANKING + "=100")`Should actually be `@Component(property = Constants.SERVICE_RANKING + ":Integer=100")` see https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-property.properties.elements do you like to propose a PR?

PR #974, opened 2025-11-17
Fix NPE in some rare cases

laeubi:
RegistryFactory.getRegistry() is documented to return null and when running inside Tycho it can in some rare cases be null and then user gets an ugly NPE.


iloveeclipse:
@laeubi : please note that we are in RC1 phase.

> "All fixes submitted must have a project lead vote on the bug report, and the fix must be reviewed by an additional committer (any committer other than the one who made the fix). A positive review from a project lead means implicit approval and no vote is needed on the bug report. Ongoing changes to documentation, tests or examples do not require approval.”

https://www.eclipse.org/lists/platform-releng-dev/msg41155.html

laeubi:
+1 from me as PL

PR #975, opened 2025-11-24
A service ranking should be specified explicitly to be of type Integer

merks:
https://github.com/eclipse-equinox/p2/pull/968


merks:
@HeikoKlare

Odd that this doesn't appear in the log and that the service also did kick when I tested that via breakpoints on the SSL context processing.

PR #1001, opened 2026-02-13
Add current and new plugin versions while update

### LINKED ISSUE #935: Display the currently installed version of a feature in the update wizard
When running an update, the UI only shows me the new/future version of the elements to update:

Image

It would be nice to have both "Current version" and "New version" to be able to decide if I want to select an update or not (e.g. it a minor/service release compared to what I already use?)

subyssurendran666:
This change enhances the update wizard to display both the currently installed version and the available update version for each feature. This allows users to easily compare versions and make more informed decisions when selecting updates.
image

Fix: https://github.com/eclipse-equinox/p2/issues/935


subyssurendran666:
There are two areas where changes were required for this fix. I have added a unit test for `IUDetailsLabelProvider`, but I’m unsure how to write a unit test for `ResolutionResultsWizardPage`. I also reviewed the revision history of the modified area, and it appears that no unit tests were added there previously. Could someone please guide me on how to approach testing this part?

laeubi:
I would say display both versions in two colums (From, To)

If you are concerned about the length I think one can shorten the version for display, e.g. only Major.Minor.Micro the qualifier is almost useless in all cases.

I think a good algorithm would be to find the first differing segment and reduce it to that e.g. lets say the Major version is different then just show Major version, if minor is different show major+minor and so on... that's the actual relevant information and the full version details can be shown when hovering in a tooltip or elsewhere.

merks:
One useful bit of information in the qualifier is the date which is typically the only information that's meaningful to the user who is generally not familiar with the versioning scheme of the unit.  I agree though that it looks a bit like "too much information" and that two columns would allow one to squeeze down the column to truncation the qualifier.    (I wonder, are the columns packed or is a column layout used.)

subyssurendran666:
Thanks for the feedbacks, I will come up with a two columns approach.

subyssurendran666:
@laeubi Could you please trigger the GitHub Copilot review for this PR?

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
```diff
@@ -125,8 +125,33 @@ public String getToolTipText(Object element) {
 				return super.getToolTipText(element);
 			}
 		});
+		// check the operation is for update or installation
+		boolean isUpdate = (resolvedOperation instanceof org.eclipse.equinox.p2.operations.UpdateOperation) ? true : false;
```
merks: We don't need a conditional expression for this.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
```diff
@@ -125,8 +125,33 @@ public String getToolTipText(Object element) {
...
+		if (isUpdate) {
+			TreeViewerColumn versionColumnOld = new TreeViewerColumn(treeViewer, SWT.LEFT);
+			versionColumnOld.getColumn().setText(ProvUIMessages.ProvUI_VersionColumnTitle_Old);
+			versionColumnOld.getColumn().setWidth(200);
```
merks: I'm never so comfortable with magic numbers.  I'm not sure what to suggest though.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
```diff
@@ -125,8 +125,33 @@ public String getToolTipText(Object element) {
...
+			versionColumnOld.setLabelProvider(new ColumnLabelProvider() {
+				@Override
+				public String getText(Object element) {
+					IInstallableUnit iu = ProvUI.getAdapter(element, IInstallableUnit.class);
+					if (element instanceof IIUElement) {
```
merks: Use the pattern syntax to declare a variable to hold the cast value.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
```diff
@@ -125,8 +125,33 @@ public String getToolTipText(Object element) {
...
+						if (((IIUElement) element).shouldShowVersion()) {
+							String toBeUpdateId = null;
+							if (element instanceof AvailableUpdateElement elm) {
+								toBeUpdateId = elm.getIUToBeUpdated().getVersion().toString();
+							}
+							return toBeUpdateId;
```
merks: I have a feeling that return null is invalid as the return value for getText.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/viewers/IUDetailsLabelProvider.java
```diff
@@ -128,6 +128,15 @@ public String getColumnText(Object element, int columnIndex) {
 				}
 				// It's a raw IU, return the version
 				return iu.getVersion().toString();
+			case IUColumnConfig.OLD_COLUMN_VERSION :
+				if (element instanceof IIUElement) {
```
merks: Assign to a variable.

I think the entire set of nested branches could be reduces to a list of `&&` expressions.

merks (changes requested):

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
subyssurendran666: Done.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
subyssurendran666: A new variable for column width had been added.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
subyssurendran666: Done.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
subyssurendran666: Done.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/viewers/IUDetailsLabelProvider.java
subyssurendran666: Done.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
```diff
@@ -125,9 +125,31 @@ public String getToolTipText(Object element) {
 				return super.getToolTipText(element);
 			}
 		});
+		final int DEFAULT_COLUMN_WIDTH = 200;
+		// check the operation is for update or installation
+		boolean isUpdate = (resolvedOperation instanceof org.eclipse.equinox.p2.operations.UpdateOperation);
```
merks: Is there a reason not to import the UpdateOperation class?

merks:
Generally it looks good, but I’m on my phone. Are the screen shots up to date? Probably you will want to write an N&N entry with nice screenshots…

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
subyssurendran666: Sorry, there is no specific reason. I have pushed the latest changes.

subyssurendran666:
> Generally it looks good, but I’m on my phone. Are the screen shots up to date? Probably you will want to write an N&N entry with nice screenshots…


Here are the latest screenshots.

image
image

merks (approved):

merks:
@subyssurendran666 

Sorry, I message up trying to do a rebase but instead did a merge commit.  Sorry sorry.  Could you please force push a rebased commit?

subyssurendran666:
> @subyssurendran666
> 
> Sorry, I message up trying to do a rebase but instead did a merge commit. Sorry sorry. Could you please force push a rebased commit?

Done.

PR #1004, opened 2026-02-19
Plugin section via file import does not work as expected

subyssurendran666:
image

Changes the behavior so that no plugins are selected by default when importing from a p2f file. This allows users to explicitly select only the plugins they want to install.

Fix: https://github.com/eclipse-equinox/p2/issues/1003

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui.importexport/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.ui.importexport/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.ui.importexport;singleton:=true
-Bundle-Version: 1.5.100.qualifier
+Bundle-Version: 1.5.200.qualifier
 Bundle-Vendor: %providerName
 Require-Bundle: org.eclipse.ui,
  org.eclipse.core.runtime;bundle-version="[3.34.200,4.0.0)",
--- a/bundles/org.eclipse.equinox.p2.ui.importexport/src/org/eclipse/equinox/internal/p2/importexport/internal/wizard/AbstractImportPage.java
+++ b/bundles/org.eclipse.equinox.p2.ui.importexport/src/org/eclipse/equinox/internal/p2/importexport/internal/wizard/AbstractImportPage.java
@@ -111,7 +111,7 @@ public abstract class AbstractImportPage extends AbstractPage {
 					IInstallableUnit iu = ProvUI.getAdapter(element, IInstallableUnit.class);
 					IQueryResult<IInstallableUnit> collector = profile.query(QueryUtil.createIUQuery(iu.getId(), new VersionRange(iu.getVersion(), true, null, false)), new NullProgressMonitor());
 					if (collector.isEmpty()) {
-						return true;
+						return false;
 					}
 				}
 				return false;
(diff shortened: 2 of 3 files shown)


subyssurendran666:
@laeubi  Could you please trigger the GitHub Copilot review for this PR?

iloveeclipse:
> Changes the behavior so that no plugins are selected by default when importing from a p2f file. This allows users to explicitly select only the plugins they want to install.

This is bad proposal & wrong solution. I want everything to be installed, I don't want to explicitly select anything, because the file in question has all what I need, and I do this literally every day.

iloveeclipse (changes requested):
As explained, it is not a solution for the problem.

subyssurendran666:
> > Changes the behavior so that no plugins are selected by default when importing from a p2f file. This allows users to explicitly select only the plugins they want to install.
> 
> This is bad proposal & wrong solution. I want everything to be installed, I don't want to explicitly select anything, because the file in question has all what I need, and I do this literally every day.

One situation I faced while setting up JDT.UI was that I only needed AssertJ and Mockito. However, the p2f file contains many other plugins as well. In that case, I had to manually deselect the unwanted plugins, or click “Deselect All” and then select only the required plugins. Because of this, the auto-selection process irritated me multiple times while setting up JDT.UI.

The actual problem was slightly different. In the existing implementation, all plugins are selected by default. If the user clicks “Deselect All” and then tries to select the required plugin, all the plugins get auto-selected again, and the plugin that the user manually selected becomes deselected.

To avoid this issue, I set all plugins to be deselected by default, and the problem was resolved. These are the reasons why I implemented it this way.

I am open to implementing this in a different way, but auto-selecting all plugins by default can definitely cause inconvenience for users who do not need all the plugins.

iloveeclipse:
> If the user clicks “Deselect All” and then tries to select the required plugin, all the plugins get auto-selected again,

**This** is the bug.

subyssurendran666:
Hi @iloveeclipse , I have incorporated the requested changes. Could you please review this PR when you have some time? Thanks in advance.

iloveeclipse:
> Hi @iloveeclipse , I have incorporated the requested changes. Could you please review this PR when you have some time? Thanks in advance.

Going back to your original issue, I've tried to reproduce on master but failed.
Here is what we you are trying to fix:

> If I click the Deselect All button to clear the default selections and then try to select only the plugin I need, the behavior is not as expected. When I click on the required plugin, instead of selecting that plugin, it selects all the other plugins and does not allow the intended plugin to be selected.

I'm using this file:
```
<?xml version='1.0' encoding='UTF-8'?>
<?p2f version='1.0.0'?>
<p2f  version='1.0.0'>
  <ius size='28'>
    <iu id='AnyEditTools.feature.group' name='AnyEditTools' version='0.0.0'>
        <repositories size='1'>
            <repository location='https://raw.githubusercontent.com/iloveeclipse/plugins/latest'/>
        </repositories>
    </iu>
    <iu id='org.eclipse.eclemma.feature.feature.group' name='EclEmma Java Code Coverage' version='0.0.0'>
        <repositories size='1'>
            <repository location='https://download.eclipse.org/eclemma/updates-master/'/>
        </repositories>
    </iu>
    <iu id='org.eclipse.jgit.lfs.feature.group' name='Eclipse Git LFS support' >
      <repositories size='1'>
        <repository location='https://download.eclipse.org/egit/updates-nightly'/>
      </repositories>
    </iu>
    <iu id='org.eclipse.egit.feature.group' name='Eclipse Git Team Provider' >
      <repositories size='1'>
        <repository location='https://download.eclipse.org/egit/updates-nightly'/>
      </repositories>
    </iu>
    <iu id='org.eclipse.equinox.executable.feature.group' name='Eclipse platform launcher executables' version='0.0.0'>
        <repositories size='1'>
            <repository location='https://download.eclipse.org/eclipse/updates/I-builds'/>
        </repositories>
    </iu>
    <iu id='com.microsoft.copilot.eclipse.feature.feature.group' name='Github Copilot' version='0.0.0'>
        <repositories size='1'>
            <repository location='https://azuredownloads-g3ahgwb5b8bkbxhd.b01.azurefd.net/github-copilot-eclipse-nightly/'/>
        </repositories>
    </iu>
  </ius>
</p2f>
```

1) If I open the import page, everything is checked.
2) After "delelect all" everything is unchecked.
3) Selecting any entry afterwards simply does what it is supposed to do, **without** this patch.

What is different? I'm using Linux and you are testing on Mac. So I assume the bug is not in the fixed line but in the Mac table/checkbox handling code. Please investigate that.

subyssurendran666:
Thank you for testing this and for sharing your insights. I will investigate it further based on your suggestions.

PR #1048, opened 2026-03-10
NextButton not disabled toggling "Hide items that are already installed"

subyssurendran666:
Toggling "Hide items that are already installed" refreshes the available software view, but the wizard does not update the Next button state. This change ensures the Next button is disabled when no installable units are selected.

fix: https://github.com/eclipse-equinox/p2/issues/1036

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/AvailableIUsPage.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/AvailableIUsPage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2007, 2018 IBM Corporation and others.
+ * Copyright (c) 2007, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -277,6 +277,14 @@ public class AvailableIUsPage extends ProvisioningWizardPage implements ISelecta
 		updateSelection();
 	}
 
+	private void updatePageCompleteAsync() {
+		Display displayPage = availableIUGroup.getCheckboxTreeViewer().getControl().getDisplay();
+		displayPage.asyncExec(() -> {
+			Object[] checked = availableIUGroup.getCheckboxTreeViewer().getCheckedElements();
+			setPageComplete(checked.length > 0);
+		});
+	}
+
 	private void createViewControlsArea(Composite parent) {
 		showLatestVersionsCheckbox = new Button(parent, SWT.CHECK);
 		showLatestVersionsCheckbox.setText(ProvUIMessages.AvailableIUsPage_ShowLatestVersions);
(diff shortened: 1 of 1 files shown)


subyssurendran666:
@laeubi Could you please trigger the GitHub Copilot review for this PR?

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/AvailableIUsPage.java
```diff
@@ -285,12 +285,20 @@ private void createViewControlsArea(Composite parent) {
 			public void widgetDefaultSelected(SelectionEvent e) {
 				updateQueryContext();
 				availableIUGroup.updateAvailableViewState();
+				Display.getDefault().asyncExec(() -> {
```
merks: Get the display from one of the widgets.

merks (changes requested):
I wonder, is it really necessary to use `asyncExec`?   Better in general to get the display from a widget in the context.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/AvailableIUsPage.java
subyssurendran666: Instead of using Display.getDefault(), I’m now using the display associated with the availableIUGroup widget. Thanks for the review comment.

merks (changes requested):
Wow, this code is kind of terrible to begin with before you improved it.  Sorry I didn't look closer before.

There appear to be three identical listeners each with two methods that have identical bodies, so like 6 copies in all!   Better would be to create one listener (adding it 3 times to the separate controls) and to de-duplicate the identical bodies.  Sorry, I know you didn't make it ugly like this, but it would sure be nice to improve it.

subyssurendran666:
> I wonder, is it really necessary to use `asyncExec`? Better in general to get the display from a widget in the context.

I had a reason for using asyncExec here. In this case it is still required because the viewer state (checked elements) is updated after update available ViewState, so a synchronous call reads stale data. Using asyncExec ensures the UI refresh is complete before evaluating the selection state. Could you please suggest an alternative approach to handle this more cleanly?

merks (approved):
This looks so much better❣️

merks:
@HannesWell 

Is the failure

> 08:10:33  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:5.0.3-SNAPSHOT:test (default-test) on project org.eclipse.equinox.p2.tests.verifier: No tests found -> [Help 1]

perhaps related to this change

https://github.com/eclipse-equinox/p2/commit/4b7608b9e656605e5fd3c27a7c244eecfb64bc07

merks:
FYI, I have this in the works hopefully to fix PR build failures:

https://github.com/eclipse-equinox/p2/pull/1061

merks:
I'll rebase to get a cleaner build.

PR #1053, opened 2026-03-15
Delete unused and deprecated MD5ArtifactComparator

manandbytes:
...as there are no usages in production code since Jan'2021 [1], when MirrorApplication switched to use ArtifactChecksumComparator.

[1] c173ee728bbb877ffc25790f2019d6dabd06aed2


akurtakov:
I believe this one should finally be pushed.

PR #1066, opened 2026-05-04
Auto-trigger updates on software site changes

### LINKED ISSUE #1062: Add "Check for Updates" button in the "No updates found" dialouge
The idea behind this is to improve the user experience. Currently, when the user sees the “No updates found” dialog and clicks on “Available software sites”, they may modify the sites and save the changes. After that, they are taken back to the same “No updates found” dialog.

At this point, it would be helpful if the system provides an option in the same dialog to directly initiate “Check for updates”(Please see the attached image).

Image

 Otherwise, the user has to close the dialog and manually go to **Help → Check for Updates** again, which is not very convenient.

subyssurendran666:
This PR reworks the update flow to improve the overall user experience.

An automatic update is now triggered when the user makes changes (add/edit/enable) to the available software sites and saves them. This removes the need for users to manually initiate an update after modifying repositories.

How to test the feature
----------------------
1. Install a plugin from the software site(this is not required if there is any plugin)
2.Then modify the software sites(add/edit or enable) the software site and "Apply and save"
3. This will trigger an auto update.

Fix: https://github.com/eclipse-equinox/p2/issues/1062

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/MessageDialogWithLink.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/MessageDialogWithLink.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2015 Red Hat Inc.
+ *  Copyright (c) 2015, 2026 Red Hat Inc and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -19,6 +19,7 @@ import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.events.SelectionListener;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.widgets.*;
@@ -47,9 +48,25 @@ public class MessageDialogWithLink extends MessageDialog {
 			this.link = new Link(composite, getMessageLabelStyle());
 			this.link.setText(this.linkMessage);
 			GridDataFactory.fillDefaults().align(SWT.FILL, SWT.BEGINNING).grab(true, false).hint(convertHorizontalDLUsToPixels(IDialogConstants.MINIMUM_MESSAGE_AREA_WIDTH), SWT.DEFAULT).applyTo(this.link);
-			for (SelectionListener linkListener : this.linkListeners) {
-				this.link.addSelectionListener(linkListener);
-			}
+			this.link.addSelectionListener(new SelectionListener() {
+				@Override
+				public void widgetSelected(SelectionEvent e) {
+					// First notify existing listeners
+					for (SelectionListener listener : linkListeners) {
+						listener.widgetSelected(e);
+					}
+
+					Shell shell = getShell();
+					if (shell != null && !shell.isDisposed()) {
+						shell.close();
+					}
+				}
+
+				@Override
+				public void widgetDefaultSelected(SelectionEvent e) {
+					widgetSelected(e);
+				}
+			});
 		}
 		return composite;
 	}
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/p2/ui/RepositoryManipulationPage.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/p2/ui/RepositoryManipulationPage.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2007, 2018 IBM Corporation and others.
+ *  Copyright (c) 2007, 2026 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -47,6 +47,7 @@ import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.*;
 import org.eclipse.ui.*;
 import org.eclipse.ui.dialogs.PatternFilter;
+import org.eclipse.ui.handlers.IHandlerService;
 import org.eclipse.ui.statushandlers.StatusManager;
 
 /**
@@ -91,6 +92,7 @@ import org.eclipse.ui.statushandlers.StatusManager;
 public class RepositoryManipulationPage extends PreferencePage implements IWorkbenchPreferencePage, ICopyable {
 	final static String DEFAULT_FILTER_TEXT = ProvUIMessages.RepositoryManipulationPage_DefaultFilterString;
 	private final static int FILTER_DELAY = 200;
+	private static final String UPDATE = "org.eclipse.equinox.p2.ui.sdk.update"; //$NON-NLS-1$
 
 	StructuredViewerProvisioningListener listener;
 	CheckboxTableViewer repositoryViewer;
@@ -99,6 +101,7 @@ public class RepositoryManipulationPage extends PreferencePage implements IWorkb
 	Policy policy;
 	Display display;
 	boolean changed = false;
+	boolean repoAddedOrEdited;
 	MetadataRepositoryElementComparator comparator;
 	RepositoryDetailsLabelProvider labelProvider;
 	RepositoryTracker tracker;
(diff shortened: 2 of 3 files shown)


bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/ProvUI.java
```diff
@@ -111,8 +112,33 @@ public static void reportStatus(IStatus status, int style) {
 				|| (style & StatusManager.SHOW) == StatusManager.SHOW) {
 			if (status.getSeverity() == IStatus.INFO) {
 				final MessageDialogWithLink dialog = new MessageDialogWithLink(ProvUI.getDefaultParentShell(),
-						ProvUIMessages.ProvUI_InformationTitle, null, status.getMessage(), MessageDialog.INFORMATION, 0,
-						IDialogConstants.OK_LABEL);
+						ProvUIMessages.ProvUI_InformationTitle, null, status.getMessage(), MessageDialog.INFORMATION, 1,
+						"Check for Updates", IDialogConstants.OK_LABEL) { //$NON-NLS-1$
```
subyssurendran666: @merks should I add new label for the "Check for Updates" button in IDialogConstants(jFace).

merks: No, I don’t think it makes sense there.

merks:
Maybe a stupid question but if automatic check for updates fails, shouldn’t I expect a manual check also to fail?

subyssurendran666:
> Maybe a stupid question but if automatic check for updates fails, shouldn’t I expect a manual check also to fail?

Hmm.. that makes sense. In that case, it might be better to hide the “Check for Updates” button when automatic updates are enabled. Thanks for pointing that out.

merks:
Where does the button you consider to hide currently exist? I thought you are proposing to add a button. But the dialog has a link to configure the sites?  Does clicking that close the dialog and open the preference page?

subyssurendran666:
> Where does the button you consider to hide currently exist? I thought you are proposing to add a button. 

The new button is in the “No updates found” dialog. In your previous comment, you mentioned the behaviour of automatic vs manual updates. Based on that, I was thinking that if the user has already enabled automatic updates (from settings), then showing this newly proposed button may not be necessary.

> But the dialog has a link to configure the sites? Does clicking that close the dialog and open the preference page?

Yes, the dialog already has a link to configure the sites. As per the existing implementation, clicking this link does not close the dialog, it simply opens the “Available Software Sites” dialog on top of it. In the current flow, the user still has to go to Help → "Check for Updates" to trigger an update, which adds a few extra steps. With the proposed button, after configuring the sites, the user can directly trigger the update from the same “No updates found” dialog.

subyssurendran666:
I was thinking about this in a slightly different way, and I feel it might improve the user experience more than my current proposal. Instead of adding a new “Check for Updates” button in the “No updates found” dialog, it might be better to automatically trigger an update check once the user makes any changes in the “Available Software Sites”.
@merks could you please share your insights on this.

merks:
I see.  Yes, if one doesn't actually configure the sites, checking again makes no sense.  But if one does configure the sites by clicking the link and actually changing the sites then its highly likely one would want to check for updates using those changes.  It's a little tricky how to be more helpful (after configuration) without providing pointless helpfulness (where checking the same thing a second time would be expected to produce the same result).

subyssurendran666:
Hi Ed, I have reworked the functionality in a slightly different way. Now, whenever the user makes any changes (add/edit/enable) to the software sites, an automatic update is triggered when they save the changes.

The changes in MessageDialogWithLink.java are to ensure that the “No updates found” dialog closes when the user clicks on the “Available Software Sites” link.

I will update the PR description and the parent tickets to reflect this approach. The PR currently has three commits, and I will rebase them once the PR is approved.

merks:
Generally it’s best to amend your commit and force push it so the PR consists of a single commit (potentially with an auto-generated version increment commit).

merks:
That sounds better.

subyssurendran666:
> Generally it’s best to amend your commit and force push it so the PR consists of a single commit (potentially with an auto-generated version increment commit).

Done

subyssurendran666:
@merks could you please review this PR if time permits ? Thanks in advance 🙏

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/MessageDialogWithLink.java
```diff
@@ -47,9 +47,26 @@ protected Control createMessageArea(Composite composite) {
 			this.link = new Link(composite, getMessageLabelStyle());
 			this.link.setText(this.linkMessage);
 			GridDataFactory.fillDefaults().align(SWT.FILL, SWT.BEGINNING).grab(true, false).hint(convertHorizontalDLUsToPixels(IDialogConstants.MINIMUM_MESSAGE_AREA_WIDTH), SWT.DEFAULT).applyTo(this.link);
-			for (SelectionListener linkListener : this.linkListeners) {
-				this.link.addSelectionListener(linkListener);
-			}
+			this.link.addSelectionListener(new SelectionListener() {
+
+				@Override
+				public void widgetSelected(org.eclipse.swt.events.SelectionEvent e) {
```
merks: Please import the name unless there is a good reason not to import it.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/MessageDialogWithLink.java
```diff
@@ -47,9 +47,26 @@ protected Control createMessageArea(Composite composite) {
 			this.link = new Link(composite, getMessageLabelStyle());
 			this.link.setText(this.linkMessage);
 			GridDataFactory.fillDefaults().align(SWT.FILL, SWT.BEGINNING).grab(true, false).hint(convertHorizontalDLUsToPixels(IDialogConstants.MINIMUM_MESSAGE_AREA_WIDTH), SWT.DEFAULT).applyTo(this.link);
-			for (SelectionListener linkListener : this.linkListeners) {
-				this.link.addSelectionListener(linkListener);
-			}
+			this.link.addSelectionListener(new SelectionListener() {
+
```
merks: Please delete this blank line.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/p2/ui/RepositoryManipulationPage.java
```diff
@@ -504,6 +507,16 @@ public boolean performOk() {
 		}
 		originalNameCache.clear();
 		originalURICache.clear();
+		
+		if (repoAddedOrEdited) {
+			Display.getDefault().asyncExec(() -> {
+				try {
+					PlatformUI.getWorkbench().getService(IHandlerService.class).executeCommand(UPDATE, null);
+				} catch (Exception e) {
+					e.printStackTrace();
```
merks: Generally it's not a good idea to simply print a stack trace.  How is such a case handled elsewhere in the code?  Is it logged?

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/p2/ui/RepositoryManipulationPage.java
```diff
@@ -99,6 +101,7 @@ public class RepositoryManipulationPage extends PreferencePage implements IWorkb
 	Policy policy;
 	Display display;
 	boolean changed = false;
+	boolean repoAddedOrEdited = false;
```
merks: It's kind of pointless to initialize a primitive to the primitive type's default value.

merks (changes requested):
Some small suggestions.  Overall looks fine.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/MessageDialogWithLink.java
subyssurendran666: Done.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/MessageDialogWithLink.java
subyssurendran666: Removed the blank line.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/p2/ui/RepositoryManipulationPage.java
subyssurendran666: In several places, exceptions were not being logged. Hence removing the printStackTrace.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/p2/ui/RepositoryManipulationPage.java
subyssurendran666: Done. Thanks for the suggestion.

merks (approved):

PR #1069, opened 2026-05-07
Hide selection count during available software filter changes

### LINKED ISSUE #1068: Hide plugin selection count message when no plugins are selected
Image

The plugin selection count message is currently shown even when no plugins are selected. This message should be hidden when there are no selected plugins to avoid unnecessary or confusing UI feedback.

Additionally, when plugins are selected, toggling the following options while installing plugins should update the selection state and corresponding UI messages correctly:

- Show only the latest versions of available software
- Group items by category
- Show only software applicable to the target environment
- Hide items that are already installed

The behaviour should remain consistent across these filters to ensure an accurate and user-friendly installation experience.

subyssurendran666:
image

Hide the plugin selection count message when the available software view is refreshed through filter options such as “Show only the latest versions of available software”, “Group items by category”, “Show only software applicable to target environment”, and “Hide items that are already installed”.

This improves the UI behaviour by preventing stale or misleading selection count messages from being shown while the filtered content is being updated.

Fix: https://github.com/eclipse-equinox/p2/issues/1068


subyssurendran666:
@merks, could you please review this PR? This functionality was missed in #1048.

merks:
I'm trusting that you've fully/properly tested this.

subyssurendran666:
> I'm trusting that you've fully/properly tested this.

Yes, I've tested the change locally and confirmed that it behaves as expected. Thanks for reviewing and merging the PR.

PR #1072, opened 2026-05-16
Suppress API errors resulting from binary compatible ecj changes

merks:


laeubi:
@merks if it is really a binary compatible change we should fix the API tools instead of suppress the error.

merks:
Yes, I think this is effectively an API tools bug, i.e., `<? extends Object>` is semantically identical to `<?>`,  but API tools treats them as different, which was sort of reasonable until ecj started to literally treat them as identical even in terms of the generated byte code changing to omitting the details about `extends Object` which now API tools sees as an API change. 

See this for related activities:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3814

So yes, it would be good to address when someone has time to address that.  Sooner is better than later because downstream clients might well be affected the same way.  In the meantime, we really needed builds (hence the qualifier changes from 3814) and we really need an error-free workspace, hence this change.  For what it's worth, I discussed with @HannesWell  via slack whether to actually increment versions according to the error messages versus suppress the error as in this PR.

It's of course just a matter of who has time to fix API tools, and as usual that's a problem.  It would definitely be the best course of action though.  I think this is related which also references other related suppression activities:

https://github.com/eclipse-pde/eclipse.pde/issues/2332

PR #1073, opened 2026-05-18
Add filter to allow disabling an OSGi capability requirement

laeubi:
This allows to use generic require capability in builds that otherwise would create a cyclic dependency issue.

This is similar to what SWT does here:

https://github.com/eclipse-platform/eclipse.platform.swt/blob/7f9089ed7daaf630970fea7fa5bdb18cfba5cdcb/bundles/org.eclipse.swt/META-INF/p2.inf

1. It create a requirement to its fragment
2. The fragment has a requirement to the host
3. To break the cycle there is a `!(org.eclipse.swt.buildtime=true)` filter property that we then set here https://github.com/eclipse-platform/eclipse.platform.swt/blob/7f9089ed7daaf630970fea7fa5bdb18cfba5cdcb/bundles/org.eclipse.swt/pom.xml#L39-L41 to disable this only for building the host.

This change has to actually happen at Tycho (where we should have a integration test), but for consistency and to gather some agreement first I opened it here.


bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/BundlesAction.java
```diff
@@ -78,6 +79,43 @@ public class BundlesAction extends AbstractPublisherAction {
...
+	 * Returns the profile property name that disables <code>Require-Capability</code>
+	 * requirements in the given OSGi namespace. Setting this property to
+	 * <code>"true"</code> on a profile causes requirements in that namespace to be
+	 * ignored during installation.
+	 *
+	 * @param namespace the OSGi namespace (e.g. {@code "osgi.ee"}, {@code "my.cap"})
+	 * @return the full profile property name for the given namespace
+	 */
+	public static String getFilterPropertyForNamespace(String namespace) {
+		return FILTER_PROPERTY_DISABLE_REQUIRE_CAPABILITY + "." + namespace; //$NON-NLS-1$
```
merks: Nit suggestion, if you used `'.'` instead of `".`. you'd not need the NLS marker.
```suggestion
		return FILTER_PROPERTY_DISABLE_REQUIRE_CAPABILITY + '.' + namespace;
```

bundles/org.eclipse.equinox.p2.tests/src/org/eclipse/equinox/p2/tests/publisher/actions/BundlesActionRequireCapabilityFilterTest.java
```diff
@@ -0,0 +1,212 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Eclipse contributors and others.
```
merks:
```suggestion
 * Copyright (c) 2026 Eclipse contributors and others.
```

bundles/org.eclipse.equinox.p2.tests/src/org/eclipse/equinox/p2/tests/publisher/actions/BundlesActionRequireCapabilityFilterTest.java
```diff
@@ -0,0 +1,212 @@
...
+ * {@link BundlesAction} carry a namespace-specific environment filter that is
+ * active by default and can be disabled by setting the profile property
+ * {@code org.eclipse.equinox.p2.disable.require.capability.<namespace>} to
+ * {@code "true"}.
+ */
+public class BundlesActionRequireCapabilityFilterTest extends AbstractProvisioningTest {
+
+	private static final String TEST_BUNDLE = "testData/BundlesActionRequireCapabilityFilterTest/testRC";
+
+	private IInstallableUnit bundleIU;
```
merks, line 38:
```suggestion
	private final IInstallableUnit bundleIU;
```

merks (approved):
Seems like a good idea.  Some tiny little suggestions.

merks (approved):
Looks fine in general.

It would be nice if somewhere in all this is show an example of what's generated in the serialized p2 metadata for this.

merks (approved):
Looks reasonable.

laeubi:
I need to check why this breaks the build but have in the meanwhile prepared a symmetric change on the Tycho side as well:

- https://github.com/eclipse-tycho/tycho/pull/6215

laeubi:
Build is now green, I'm not sure if we should merge this right now or wait for the next release before M2. @merks (or others) WDYT?

merks:
@laeubi 

I think it's find to merge this for this cycle.  We've done similar types of changes in the past without a single complaint as a result.

PR #1075, opened 2026-05-21
Support plugin_customization.ini for p2 trust preferences

vogella:
p2's "trust always" preferences (`trustAlways` for artifacts, `trustAllAuthorities` for authorities) were read with `IEclipsePreferences.getBoolean()` on a single `ProfileScope` node, which by contract never falls through to other scopes. Values seeded into `DefaultScope` via a product's `plugin_customization.ini` were therefore unreachable, even though `docs/Trust_Settings.md` documents that as the configuration path.

Both `isTrustAlways()` methods now read through `IPreferencesService` with `{ ProfileScope, DefaultScope.INSTANCE }`, so user-set profile values still win and `plugin_customization.ini` defaults are honored as a fallback. Writes still go to ProfileScope and are unchanged.

Mirrors the fix in eclipse-platform/eclipse.platform.ui#1795 for the e4 workbench renderer preferences.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.engine/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.engine/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.engine;singleton:=true
-Bundle-Version: 2.11.100.qualifier
+Bundle-Version: 2.11.200.qualifier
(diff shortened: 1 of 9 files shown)


vogella:
@merks AFAIK you implement this, have a look if you have the time.

merks:
This seems reasonable.  

Note that `-Dp2.trustedAuthorities` can also be used by product to customize the list.

vogella:
> This seems reasonable.

> Note that `-Dp2.trustedAuthorities` can also be used by product to customize the list.

Thanks, AFAICS that's already wired up via p2.trustedAuthorities in AuthorityChecker and noted in Trust_Settings.md.

iloveeclipse:
This PR caused regression, see https://github.com/eclipse-equinox/p2/issues/1098

PR #1076, opened 2026-05-29
Revert "Use CI Friendly versions for the P2 maven build"

HannesWell:
This reverts the changes
- https://github.com/eclipse-equinox/p2/pull/109
- https://github.com/eclipse-equinox/p2/pull/686

Together they break the the automated update of versions:
- https://github.com/eclipse-tycho/tycho/pull/5933#discussion_r3325625599

Reverting both unblocks the preparation of the next development cycle.
And now that only almost all p2 projects are pomless and the version update is automated, the benefit of it is small anyway.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.tests/pom.xml
+++ b/bundles/org.eclipse.equinox.p2.tests/pom.xml
@@ -10,7 +10,7 @@
 	<parent>
 		<groupId>org.eclipse.platform</groupId>
 		<artifactId>rt.equinox.p2</artifactId>
-		<version>${releaseVersion}${qualifier}</version>
+		<version>4.40.0-SNAPSHOT</version>
 		<relativePath>../../</relativePath>
 	</parent>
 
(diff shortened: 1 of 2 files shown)


HannesWell:
This is a RelEng change that's not relevant for the upcoming release but necessary to unblock the preparation of the next development cycle.
The quality-gate failure is unrelated.

laeubi:
> And now that only almost all p2 projects are pomless and the version update is automated, the benefit of it is small anyway.

The main benefit is/was that we actually test this feature is working. So its a bit unfortunate to not fix it and use a snapshot of Tycho and instead revert the usage all together.

HannesWell:
> So its a bit unfortunate to not fix it and use a snapshot of Tycho and instead revert the usage all together.

That bug was blocking the preparation for the new development cycle and I didn't want to delay that by one or two days to fix that bug first, which would then also have to be done in a hurry.
Furthermore I personally didn't want to spend that time, when there is an easy workaround or pressure somebody to fix it fast.

> The main benefit is/was that we actually test this feature is working. 

Maybe it would then be better to have a corresponding integration test for it in Tycho? Then it's also easier to notice if the feature breaks, respectively it will be noticed in verification builds.

Issue #1083, opened 2026-06-08
Available version is irrelevant in the uninstall dialog


subyssurendran666:
<img width="1718" height="1004" alt="Image" src="https://github.com/user-attachments/assets/2af3042f-6490-43c0-bf85-eb163c75bb1d" />
In the plugin uninstall dialog, there is no point to display the "Available Version", also the plugin version should display in the "Installed version"

subyssurendran666:
<img width="1255" height="627" alt="Image" src="https://github.com/user-attachments/assets/1d3a7f5e-7d94-4f0d-a12b-0d2d9aeca0b2" />
This should fix here also.

PR #1085, opened 2026-06-09
Disable Back button on the initial uninstall selection page

### LINKED ISSUE #1080: Back button is irrelevant in the Uninstall plugin page
Image

Steps to navigate to Uninstall plugin details page.
1. Help -> Install new Software .
2. Click "already installed" link from the Available Software page. This will navigate to the "Eclipse SDK Installation Details" page.
3. Select any plugin -> Click "Uninstall...".
4. Now there is no point to enable the "< Back" button from the "Uninstall" dialoge.

subyssurendran666:
image

Depending on the provisioning status, the uninstall wizard may start on the selection page ("Check the items that you wish to uninstall."). In this scenario, the selection page is the first page presented to the user, but the Back button remains enabled even though there is no previous page available for navigation. Clicking the button has no effect. This change disables the Back button when the uninstall selection page is displayed as the initial page of the workflow, preventing a non-functional navigation option while preserving existing Back button behavior in workflows where a previous page is available.

Fix: https://github.com/eclipse-equinox/p2/issues/1080

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ProvisioningOperationWizard.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ProvisioningOperationWizard.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2008, 2018 IBM Corporation and others.
+ *  Copyright (c) 2008, 2026 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -117,7 +117,7 @@ public abstract class ProvisioningOperationWizard extends Wizard {
 
 	@Override
 	public IWizardPage getPreviousPage(IWizardPage page) {
-		if (page == errorPage) {
+		if (page == errorPage && page != mainPage) {
 			return mainPage;
 		}
 		return super.getPreviousPage(page);


subyssurendran666:
@merks could you please review this PR?

merks (commented):
I wonder why this method in the super class is not  working?

https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface/src/org/eclipse/jface/wizard/WizardDialog.java#L1345-L1351

Does the current page have a previous page that is somehow not null but also not really there/visible?

subyssurendran666:
Thanks for pointing this out. I investigated the JFace implementation further to understand why the Back button is being enabled instead of working around it in `updateButtons()`.

From my debugging, the current page in this scenario is always UninstallWizardPage, and WizardDialog enables the Back button because `getPreviousPage()` returns a non-null page. Specifically, `super.getPreviousPage(page)` resolves to SelectableIUsPage, so from JFace's perspective there is a valid previous page and the default logic enables the Back button.

I also considered overriding `canFlipToPreviousPage()` instead of `updateButtons()`, as it is a more appropriate extension point for controlling the Back button state. However, before taking that approach, I wanted to understand why the default implementation believes a previous page exists in the first place. It seems there is a mismatch between the wizard navigation model and the actual user-visible flow.

I will continue investigating whether the underlying navigation logic can be corrected so that getPreviousPage() and consequently `canFlipToPreviousPage()` naturally reflects the expected behavior. If that is not feasible, overriding `canFlipToPreviousPage()` would likely be a cleaner solution than manipulating the button state in `updateButtons()`.

merks:
Yes there’s something fishy here and probably best to fix the cause rather than treat the symptom.

subyssurendran666:
> Yes there’s something fishy here and probably best to fix the cause rather than treat the symptom.

You are absolutely right. The issue was that `SelectableIUsPage.getPreviousPage()` was returning a non-null value (itself), which made JFace think there was a previous page and enabled the Back button. I have updated `SelectableIUsPage.getPreviousPage()` to return null when used in an UninstallWizard, which correctly tells JFace there's no previous page and automatically disables the Back button through the standard navigation mechanism, no need for the updateButtons() override anymore.

merks (changes requested):
I think the more fundamental problem is in the wizard itself where we try get the previous page of page, but that page is the error page and also the main page:

image

I think it should be like this:

```
	@Override
	public IWizardPage getPreviousPage(IWizardPage page) {
		if (page == errorPage && page != mainPage) {
			return mainPage;
		}
		return super.getPreviousPage(page);
	}
```

Then the wizard comes up like this:

image

And if we hit back, we see the first page where we might make our selection smaller:

image

The version columns on this page are a little unhelpful, i.e., available versus installed..

subyssurendran666:
Thanks for the help Ed. I have pushed the latest changes. 

> The version columns on this page are a little unhelpful, i.e., available versus installed..

This behaviour is reported in this [github issue](https://github.com/eclipse-equinox/p2/issues/1083).

merks (approved):
This seems best.

PR #1086, opened 2026-06-19
Show the IU id column in the Available Software dialog

vogella:
The Available Software install dialog now shows an Id column after Version, so users can see the feature or plug-in id of each entry, not just its display name and version.
The id is hidden for category (group) rows, matching the existing behaviour of the Version column.
The filter field now also matches against the id, and IUs without a display name no longer duplicate their id in the Name column since it has its own column.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/viewers/IUDetailsLabelProvider.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/viewers/IUDetailsLabelProvider.java
@@ -86,8 +86,8 @@ public class IUDetailsLabelProvider extends ColumnLabelProvider implements ITabl
 		IInstallableUnit iu = ProvUI.getAdapter(element, IInstallableUnit.class);
 		if (iu == null) {
 			if (columnIndex == 0) {
-				if (element instanceof ProvElement) {
-					return ((ProvElement) element).getLabel(element);
+				if (element instanceof ProvElement provElement) {
+					return provElement.getLabel(element);
 				}
 				return element.toString();
 			}
@@ -96,6 +96,15 @@ public class IUDetailsLabelProvider extends ColumnLabelProvider implements ITabl
 
 		switch (columnContent) {
 			case IUColumnConfig.COLUMN_ID :
+				// If it's an element, determine if the id should be shown.
+				// Categories (groups) reuse the same rule as the version column.
+				if (element instanceof IIUElement iuElement) {
+					if (iuElement.shouldShowVersion()) {
+						return iu.getId();
+					}
+					return BLANK;
+				}
+				// It's a raw IU, return the id
 				return iu.getId();
 			case IUColumnConfig.COLUMN_NAME :
 				// Get the iu name in the current locale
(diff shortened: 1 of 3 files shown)


merks:
I can only imagine what it looks like.  A picture would seem helpful.

@subyssurendran666

Given you've been working in this area recently, would you be able to review these changes?

vogella:
image

merks:
Looks okay, and there seems to be generally enough real estate.  But I imagine for most users such technical details are meaningless, irrelevant, and maybe really just noise even... The details are available for anyone who is super interested in that, but this seem highly technical.

image

We do show it here too for what's already installed.

image

I do wonder what's the end-user motivation for seeing such technical details that seem likely of interest only to plug-in developers...

vogella:
In the default sized dialog the id is out of the visible area so the average user will not see them. 

And as you explained we show the id almost everywhere. So not showing it on the first installation page is inconsistent for a user who just checked this information in his about dialog or the other option.

(Also copy via right mouse click is broken on the second page of the wizard, so in the current setup it is impossible to copy the id, I plan to fix this in the future but).

And p2 is the technical installation dialog while marketplace is the non-technical installation means.

merks:
I can't judge what the default size dialog is like.  I have 4k 42" main monitor and it seem lots of stuff is always visible.

Anyway, it's no big deal to show more details but I do feel it's likely to be noise for 99% of the users which makes it not highly compelling.  So be it.

subyssurendran666:
I checked the PR on my laptop (16-inch screen) and found that there is enough space to add an additional Provider column, although spacing may become an issue when lengthy names, IDs, and provider values are present. However, Ed Merks's point is valid, since this information is irrelevant to most users. Nevertheless, we already show the Provider column in many other places, so it should be fine to add it to the Available Software dialog as well.

subyssurendran666:
> @subyssurendran666
> 
> Given you've been working in this area recently, would you be able to review these changes?

The PR looks good to me.

merks (approved):
I'll mark it as approved, since the effort has been invested.  But I remain doubtful about the noise-to-benefit ratio given the dialog is designed and intended to provide a simple user-facing outline of the available content, not complete view of the low-level technical details.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/viewers/IUDetailsLabelProvider.java
```diff
@@ -96,6 +96,15 @@ public String getColumnText(Object element, int columnIndex) {
 
 		switch (columnContent) {
 			case IUColumnConfig.COLUMN_ID :
+				// If it's an element, determine if the id should be shown.
+				// Categories (groups) reuse the same rule as the version column.
+				if (element instanceof IIUElement) {
```
gravemalte: Could we use here the `instanceof` pattern matching cast?  That way the "traditional cast" would be gone. Maybe a matter of taste but I would prefer it that way.

vogella: will update, also the second case in which the instanceof Patter in not used in this label provider

vogella:
Thanks for all the reviews

PR #1089, opened 2026-06-22
Validate repository URL reachability before saving

### LINKED ISSUE #1081: Show a warning if the repository is unreachable
Image

Currently, there is no way for users to validate whether the entered update site URL is valid. If the site is unreachable or inaccessible, the user receives no immediate feedback. It would be helpful to notify the user when the specified site cannot be reached, allowing them to verify and correct the URL before proceeding.

NB: This should be applicable for update the site URL.

subyssurendran666:
image

Repository URLs are now validated for reachability before being added or updated, preventing unreachable URLs from being saved to the system.

Fix: https://github.com/eclipse-equinox/p2/issues/1081

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/RepositoryNameAndLocationDialog.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/RepositoryNameAndLocationDialog.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2007, 2018 IBM Corporation and others.
+ * Copyright (c) 2007, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -18,7 +18,9 @@ import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.equinox.internal.p2.ui.ProvUIActivator;
 import org.eclipse.equinox.internal.p2.ui.ProvUIMessages;
+import org.eclipse.equinox.p2.core.ProvisionException;
 import org.eclipse.equinox.p2.operations.RepositoryTracker;
+import org.eclipse.equinox.p2.repository.metadata.IMetadataRepositoryManager;
 import org.eclipse.equinox.p2.ui.ProvisioningUI;
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.dialogs.IDialogConstants;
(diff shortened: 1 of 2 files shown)


subyssurendran666:
The Add Site dialog currently validates only the URL format, which allows invalid or unreachable repository URLs to be added. This can lead to a ProvisionException when the repository is accessed later. This PR adds reachability checks to both the Add Site and Edit Site dialogs to prevent saving unreachable repository URLs.

@merks Could you please review the PR when you have some time.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/RepositoryNameAndLocationDialog.java
```diff
@@ -168,6 +180,25 @@ protected IStatus validateRepositoryURL(final boolean contactRepositories) {
...
+	/**
+	 * Validate that the repository at the given location is reachable by attempting
+	 * to load it.
+	 *
+	 * @param uri the repository location to validate
+	 * @return a status indicating whether the repository is reachable
+	 */
+	protected IStatus validateRepositoryReachability(URI uri) {
+		try {
+			IMetadataRepositoryManager manager = org.eclipse.equinox.internal.p2.ui.ProvUI
```
gravemalte: May we import the `ProvUI` class?

gravemalte (commented):
And a meta question. Why do we allocate 

`final IStatus[] status = new IStatus[1];`  ?

Couldn't we simply the code without the array?

merks:
I need to find some time to actually try it out. 

I'm a little concerned that loading a repository into the manger will tend to leave it loaded in the manager.  But then it's highly like to be loaded soon anyway.  Also, p2 has a bad habit of recording/remembering repositories, so I want to double check that in fact this doesn't happen as a side effect of loading.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/RepositoryNameAndLocationDialog.java
subyssurendran666: You are right. I have imported the ProvUI class.

subyssurendran666:
> And a meta question. Why do we allocate
> 
> `final IStatus[] status = new IStatus[1];` ?
> 
> Couldn't we simply the code without the array?

The variables captured by a lambda must be final or effectively final.  Since the status needs to be updated from within the lambda, a mutable holder is required. The current implementation uses a single-element array for that purpose, `AtomicReference<IStatus> `would be another alternative.

subyssurendran666:
After reconsidering the approach, I think Eclipse should not restrict users from entering arbitrary update site URLs. Instead, it would provide a better user experience if unreachable repositories were preserved, allowing users to modify or remove them later, as proposed in #1102 .



Since this PR is superseded by #1102 , I am closing it.

PR #1090, opened 2026-06-23
Remove Available Version column from uninstall dialog

### LINKED ISSUE #1083: Available version is irrelevant in the uninstall dialog

subyssurendran666:
image

image

The uninstall dialog currently shows an "Available Version" column, which is not meaningful during uninstallation. This change removes that column and displays the plugin version under "Installed Version" instead.
The same behavior has been applied to the update page when the original request is modified.

Fix: https://github.com/eclipse-equinox/p2/issues/1083

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
@@ -17,8 +17,7 @@
 package org.eclipse.equinox.internal.p2.ui.dialogs;
 
 import java.util.Collection;
-import org.eclipse.core.runtime.IStatus;
-import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.*;
 import org.eclipse.equinox.internal.p2.ui.*;
 import org.eclipse.equinox.internal.p2.ui.model.*;
 import org.eclipse.equinox.internal.p2.ui.viewers.*;
@@ -127,7 +126,7 @@ public abstract class ResolutionResultsWizardPage extends ResolutionStatusPage {
 		});
 		final int DEFAULT_COLUMN_WIDTH = 200;
 		// check the operation is for update or installation
-		boolean isUpdate = (resolvedOperation instanceof UpdateOperation);
+		boolean isUpdate = shouldShowAvailableVersionColumn() && resolvedOperation instanceof UpdateOperation;
 		if (isUpdate) {
 			TreeViewerColumn versionColumnOld = new TreeViewerColumn(treeViewer, SWT.LEFT);
 			versionColumnOld.getColumn().setText(ProvUIMessages.ProvUI_VersionColumnTitle_Old);
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionStatusPage.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionStatusPage.java
@@ -22,6 +22,7 @@ import org.eclipse.equinox.internal.p2.ui.model.*;
 import org.eclipse.equinox.internal.p2.ui.viewers.IUColumnConfig;
 import org.eclipse.equinox.p2.metadata.IInstallableUnit;
 import org.eclipse.equinox.p2.operations.ProfileChangeOperation;
+import org.eclipse.equinox.p2.operations.UninstallOperation;
 import org.eclipse.equinox.p2.ui.ProvisioningUI;
 import org.eclipse.jface.dialogs.IDialogSettings;
 import org.eclipse.jface.dialogs.IMessageProvider;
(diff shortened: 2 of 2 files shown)


subyssurendran666:
@merks Could you please review this PR when you have a chance ?

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/SelectableIUsPage.java
```diff
@@ -57,6 +57,12 @@ public class SelectableIUsPage extends ResolutionStatusPage implements IResoluti
 	protected Policy policy;
 	SashForm sashForm;
 
+	private static final String NAME_COLUMN_WIDTH = "NameColumnWidth"; //$NON-NLS-1$
+	private static final String VERSION_COLUMN_WIDTH = "VersionColumnWidth"; //$NON-NLS-1$
+	private static final String ID_COLUMN_WIDTH = "IDColumnWidth"; //$NON-NLS-1$
```
merks: All the private statics should be grouped at the top.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/SelectableIUsPage.java
```diff
@@ -337,6 +343,39 @@ protected SashForm getSashForm() {
...
+		return nameColumn;
+	}
+
+	private IUColumnConfig getVersionColumn() {
+		return versionColumn;
+	}
+
+	private IUColumnConfig getIdColumn() {
+		return idColumn;
+	}
```
merks: I don't see the point of private methods to access the private fields.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/SelectableIUsPage.java
```diff
@@ -337,6 +343,39 @@ protected SashForm getSashForm() {
...
+
+	@Override
+	protected void getColumnWidthsFromSettings() {
```
merks: This is also really confusing me because it looks like a copy, and it's a copy because base class uses org.eclipse.equinox.internal.p2.ui.dialogs.ResolutionStatusPage.getNameColumn() which is private.

All this is really confusing and will make this less maintainable.

Might it not be better to have a protected method that one might override to filter out the field you want filtered out rather than so much duplication?

merks (changes requested):
I find it really confusing.  It would seem better to make the base class smarter that you can filter something out in a subclass is a very simple way rather than all this duplicate I see here.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/SelectableIUsPage.java
subyssurendran666: Logic has been moved to the ResolutionStatusPage. Hence private static fields are not required here.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/SelectableIUsPage.java
subyssurendran666: You are right, getter methods not are not required.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/SelectableIUsPage.java
subyssurendran666: Done.

subyssurendran666:
> I find it really confusing. It would seem better to make the base class smarter that you can filter something out in a subclass is a very simple way rather than all this duplicate I see here.

The logic has been moved to the parent `ResolutionStatusPage`, where the IUColumnConfig is configured conditionally. Thanks for the insight.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionStatusPage.java
```diff
@@ -334,4 +339,8 @@ private boolean isLocked(AvailableIUElement elementToBeUpdated) {
 		}
 		return false;
 	}
+
+	protected boolean shouldShowAvailableVersionColumn() {
+		return !(this.getProvisioningWizard().operation instanceof UninstallOperation);
```
merks: We don't need the `this.`

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
```diff
@@ -365,4 +364,24 @@ protected int getColumnWidth(int index) {
...
+	 * {@link UpdateWizardPage}.
+	 * <p>
+	 * If the wizard's starting page cannot be obtained due to an
+	 * {@link AssertionFailedException}, this method returns {@code false}.
+	 * </p>
+	 *
+	 * @return {@code true} if the wizard was started from an
+	 *         {@code UpdateWizardPage}; {@code false} otherwise
+	 */
+	private boolean isRequestModified() {
```
merks: I'm trying to wrap my head around how this guard is different from shouldShowAvailableVersionColumn in the other file because that guard is much simpler and doesn't depend on checking start pages...

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionStatusPage.java
subyssurendran666: Done.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/ResolutionResultsWizardPage.java
subyssurendran666: Now, reusing the shouldShowAvailableVersionColumn method from the ResolutionStatusPage.

merks:
We can’t merge merge commits so this needs to be rebased and force pushed.

subyssurendran666:
> We can’t merge merge commits so this needs to be rebased and force pushed.

It was added accidentally when I used the update branch option on Github. I have addressed the review comments. Could you please re-review when you have a moment ?

merks (approved):
That looks simpler.  I will trust your manual testing and your good judgement because I don't have time for manual testing.

PR #1094, opened 2026-07-07
Fix duplicate error dialog when an unreachable update site is added

### LINKED ISSUE #1093: Duplicate error dialog displayed when an update site is unreachable
https://github.com/user-attachments/assets/0e745a21-1106-45d8-8e97-5847bb745b1a

When an unreachable update site is accessed, a ProvisionException is raised. The error dialog is initially shown briefly and then displayed again, resulting in duplicate error dialogs.

Image

Image

subyssurendran666:
When an unreachable update site URL was added, the error dialog appeared twice because the repository was being loaded twice via two different code paths in MetadataRepositoryElement. Fixed by loading through the metadata repository manager directly so the ProvisionException propagates to the single correct catch block in fetchChildren().

https://github.com/user-attachments/assets/13b6ca4f-9f8d-4ec0-a317-ba140e41833f

Fix: https://github.com/eclipse-equinox/p2/issues/1093


subyssurendran666:
@merks could you please review this PR if you have a moment ? Thanks in advance 🙏

PR #1095, opened 2026-07-07
Trigger auto-update after importing sites from bookmark file

subyssurendran666:
Importing update sites via a bookmarks.xml file did not trigger the auto-update check after clicking Apply and Close, unlike the Add and Edit actions. Set repoAddedOrEdited = true in importRepositories() to fix this inconsistency.

OLD Implementation
-----------------------

https://github.com/user-attachments/assets/4892d8be-085b-4a61-968e-8a4d8a410bfa 

New implementation
--------------------

https://github.com/user-attachments/assets/5506ac4d-8deb-459d-9304-948f8906e183

Parent issue: https://github.com/eclipse-equinox/p2/issues/1062
NB: This was missed in #1066


merks:
@subyssurendran666 

I assume this is ready to merge?

subyssurendran666:
> @subyssurendran666
> 
> I assume this is ready to merge?

Yes, it should be ready to merge. Thank you for taking a look.

PR #1103, opened 2026-07-13
Honour eclipse.p2.checksums.disable as documented

### LINKED ISSUE #561: eclipse.p2.checksums.disable is not working correctly (true means false and false means true)
https://github.com/eclipse-equinox/p2/blob/master/bundles/org.eclipse.equinox.p2.artifact.repository/src/org/eclipse/equinox/internal/p2/artifact/repository/simple/SimpleArtifactRepository.java#L778
was refactored 9 months ago, but there a mistake was made.
!"true" was changed to !"FALSE" and now you have to set the property to false to disable the checksum checks (eclipse.p2.checksums.disable=false).

joaodinissf:
Fixes #561.

`isChecksumsEnabled()` reads the negatively-named `eclipse.p2.checksums.disable`
property with the `!FALSE.equals(...)` idiom, which is only correct for
positively-named properties. As a result `=true` (the documented way to disable
checksum verification) was a no-op, and `=false` was the only value that
disabled it.

Parse the property as a boolean instead: verification stays on by default and
is disabled only when the flag is `true`. `Boolean.parseBoolean` also makes the
check null-safe and case-insensitive.

### DIFF
--- a/bundles/org.eclipse.equinox.p2.artifact.repository/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.artifact.repository/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.p2.artifact.repository;singleton:=true
-Bundle-Version: 1.5.900.qualifier
+Bundle-Version: 1.5.1000.qualifier
 Bundle-Activator: org.eclipse.equinox.internal.p2.artifact.repository.Activator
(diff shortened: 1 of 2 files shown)


merks:
Thanks for your PR!  The change looks good.   The bundle version incrementer should have kicked, but something is broken in that regard.  Maybe it will get fixed quickly.  

This version needs to be incremented by 100

image

I.e., to `1.5.1000.qualifier`.

To speed the process up, you can make that change with a second commit; usually the increments are separate from the change just in case the change needs to be reverted...

joaodinissf:
Thanks for the quick feedback, hopefully the update does what you requested.
If you need anything else from my side, just let me know!

PR #1104, opened 2026-07-13
Fix o.e.e.internal.p2.repository.Activator.getParserFactory

merks:
- Eliminate the use of a service tracker to create a SAXParserFactory. and instead use o.e.e.internal.p2.core.helpers.SecureXMLUtil.newSecureSAXParserFactory.


akurtakov (approved):
Change looks good overall (requires version bump to build). This is totally not needed nowadays.

tjwatson (approved):

PR #1106, opened 2026-07-18
Fix #990: Grouping Items in Install dialog gives different results than without grouping

subyssurendran666:
The flat view (grouping disabled) used createIUGroupQuery() which only matched IUs with type.group=true, while the category view surfaced everything reachable through category requirements, causing the two modes to show different content.

Fixed in QueryProvider by collecting all category requirements recursively and OR-ing them with the group query, so the flat view includes the same installable content as the category view.

Fix: https://github.com/eclipse-equinox/p2/issues/990

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2010 IBM Corporation and others.
+ * Copyright (c) 2008, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -129,13 +129,18 @@ public class QueryProvider {
 				if (element instanceof MetadataRepositories || element instanceof MetadataRepositoryElement) {
 					if (context.getViewType() == IUViewQueryContext.AVAILABLE_VIEW_FLAT || !context.getUseCategories()) {
 						AvailableIUWrapper wrapper = new AvailableIUWrapper(queryable, element, false, context.getShowAvailableChildren());
-						if (showLatest) {
-							topLevelQuery = QueryUtil.createLatestQuery(topLevelQuery);
-						}
 						if (targetProfile != null) {
 							wrapper.markInstalledIUs(targetProfile, hideInstalled);
 						}
-						return new ElementQueryDescriptor(queryable, topLevelQuery, new Collector<>(), wrapper);
+						// Include category members in flat view to avoid missing items.
+						Collector<IInstallableUnit> collector = new Collector<>();
+						@SuppressWarnings("unchecked")
+						IQueryable<IInstallableUnit> iuQueryable = (IQueryable<IInstallableUnit>) queryable;
+						Collection<IInstallableUnit> allCategoryMembers = collectAllCategoryMembers(iuQueryable);
+						collector.addAll(new CollectionResult<>(allCategoryMembers));
+
+						IQuery<IInstallableUnit> flatQuery = showLatest ? QueryUtil.createLatestQuery(topLevelQuery) : topLevelQuery;
+						return new ElementQueryDescriptor(queryable, flatQuery, collector, wrapper);
 					}
 					// Installed content not a concern for collecting categories
 					return new ElementQueryDescriptor(queryable, categoryQuery, new Collector<>(), new CategoryElementWrapper(queryable, element));
@@ -247,5 +252,33 @@ public class QueryProvider {
 			default :
 				return null;
 		}
+	}
+
+	/**
+	 * Collects all IUs reachable through categories, including nested ones,
+	 * so that flat view shows same content as category view.
+	 */
+	private static Collection<IInstallableUnit> collectAllCategoryMembers(IQueryable<IInstallableUnit> queryable) {
+		Collection<IInstallableUnit> allCategories = queryable.query(QueryUtil.createIUCategoryQuery(), null)
+				.toUnmodifiableSet();
+
+		Set<IInstallableUnit> visitedCategories = new HashSet<>();
+		Set<IInstallableUnit> members = new LinkedHashSet<>(); // set: an IU can belong to more than one category
+		Deque<IInstallableUnit> toProcess = new ArrayDeque<>(allCategories);
+
+		while (!toProcess.isEmpty()) {
+			IInstallableUnit category = toProcess.pop();
+			if (!visitedCategories.add(category)) {
+				continue; // already processed, avoid infinite loops
+			}
+			for (IInstallableUnit member : queryable.query(QueryUtil.createIUCategoryMemberQuery(category), null)) {
+				if (QueryUtil.isCategory(member)) {
+					toProcess.push(member); // nested category — walk its members too
+				} else {
+					members.add(member);
+				}
+			}
+		}
+		return members;
 	}
 }
--- a/bundles/org.eclipse.equinox.p2.tests.ui/src/org/eclipse/equinox/p2/tests/ui/query/QueryProviderTests.java
+++ b/bundles/org.eclipse.equinox.p2.tests.ui/src/org/eclipse/equinox/p2/tests/ui/query/QueryProviderTests.java
@@ -16,8 +16,8 @@ package org.eclipse.equinox.p2.tests.ui.query;
 import java.util.ArrayList;
 import java.util.HashMap;
 import org.eclipse.equinox.internal.p2.ui.model.*;
-import org.eclipse.equinox.p2.metadata.IInstallableUnit;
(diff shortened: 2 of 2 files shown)


merks:
So previously it was showing only features but it's possible for a category to contain a bundle and those would not be shown in the un-groupd view.  Is that correct?

subyssurendran666:
Yes, that is correct. Previously the ungrouped view was only showing IUs with org.eclipse.equinox.p2.type.group=true (features), so if a category contained a plain bundle without that property, it would not show up in the ungrouped view at all.

The fix collects all requirements reachable through every category recursively and OR-s them with the group query, so now the ungrouped view will show everything that is visible when browsing by category.

Existing implementation
=================
image
image

New implementation
===============
image
image

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
```diff
@@ -248,4 +261,38 @@ public ElementQueryDescriptor getQueryDescriptor(final QueriedElement element) {
...
+	 * Collects all IU requirements reachable through categories,
+	 * including nested ones, so that flat view shows same content
+	 * as category view.
+	 */
+	private static Collection<IRequirement> collectAllCategoryRequirements(IQueryable<IInstallableUnit> queryable) {
+		Map<String, IInstallableUnit> categoryById = new HashMap<>();
+		queryable.query(QueryUtil.createIUCategoryQuery(), null)
+				.forEach(cat -> categoryById.put(cat.getId(), cat));
+
+		Collection<IRequirement> result = new ArrayList<>();
```
merks: Given that an IU can be in multiple categories, maybe it's better to make this a set not a list?

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
```diff
@@ -248,4 +261,38 @@ public ElementQueryDescriptor getQueryDescriptor(final QueriedElement element) {
...
+		Set<String> visited = new HashSet<>();
+		Deque<IInstallableUnit> stack = new ArrayDeque<>(categoryById.values());
+
+		while (!stack.isEmpty()) {
+			IInstallableUnit cat = stack.pop();
+			if (!visited.add(cat.getId())) {
+				continue; // already processed, avoid infinite loops
+			}
+			for (IRequirement req : cat.getRequirements()) {
+				result.add(req);
```
merks: If this were already in the set, we'd not need to do anything.

Also, if we determine this to be a category before, then we'd need not need to add it at all, right?

merks (commented):
Overall I get the sense that this would be simpler if it used this to collect the category members.

org.eclipse.equinox.p2.query.QueryUtil.createIUCategoryMemberQuery(IInstallableUnit)

One can filter out the categories themselves with 

org.eclipse.equinox.p2.query.QueryUtil.isCategory(IInstallableUnit)

Then you'd never need to visit the detailed requirements of the IUs themselves...

I don't think the current implementation handles multiple versions  of the same ID very well.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
subyssurendran666: Thanks for pointing it out, changed it to a LinkedHashSet.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
subyssurendran666: Refactored to use QueryUtil.createIUCategoryMemberQuery() and QueryUtil.isCategory() directly. This ensures that category IUs are never added to the result, while the visited check handles duplicates cleanly.

subyssurendran666:
> Overall I get the sense that this would be simpler if it used this to collect the category members.
> 
> org.eclipse.equinox.p2.query.QueryUtil.createIUCategoryMemberQuery(IInstallableUnit)
> 
> One can filter out the categories themselves with
> 
> org.eclipse.equinox.p2.query.QueryUtil.isCategory(IInstallableUnit)
> 
> Then you'd never need to visit the detailed requirements of the IUs themselves...
> 
> I don't think the current implementation handles multiple versions of the same ID very well.

Thank you for the suggestion. Refactored to use `QueryUtil.createIUCategoryMemberQuery()` and `QueryUtil.isCategory() `as suggested, and fixed the multiple versions issue as well.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
```diff
@@ -248,4 +260,32 @@ public ElementQueryDescriptor getQueryDescriptor(final QueriedElement element) {
...
+		Set<IInstallableUnit> members = new LinkedHashSet<>(); // set: an IU can belong to more than one category
+		Deque<IInstallableUnit> toProcess = new ArrayDeque<>(allCategories);
+
+		while (!toProcess.isEmpty()) {
+			IInstallableUnit category = toProcess.pop();
+			if (!visitedCategories.add(category)) {
+				continue; // already processed, avoid infinite loops
+			}
+			for (IInstallableUnit member : queryable.query(QueryUtil.createIUCategoryMemberQuery(category), null)) {
+				members.add(member); // every direct member belongs in the flat view
```
merks: Should we avoid add a category into the members?

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
```diff
@@ -248,4 +260,32 @@ public ElementQueryDescriptor getQueryDescriptor(final QueriedElement element) {
...
+				if (QueryUtil.isCategory(member)) {
+					toProcess.push(member); // nested category — walk its members too
```
merks, line 276: I was going to say, don't push it if it's already visited, but that would involve repeating the guard at like 278, so that seems okay.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
```diff
@@ -129,13 +129,25 @@ public ElementQueryDescriptor getQueryDescriptor(final QueriedElement element) {
...
-						return new ElementQueryDescriptor(queryable, topLevelQuery, new Collector<>(), wrapper);
+						// Include category members in flat view to avoid missing items.
+						@SuppressWarnings("unchecked")
+						IQueryable<IInstallableUnit> iuQueryable = (IQueryable<IInstallableUnit>) queryable;
+						Collection<IInstallableUnit> allCategoryMembers = collectAllCategoryMembers(iuQueryable);
+						IQuery<IInstallableUnit> flatQuery;
+						if (!allCategoryMembers.isEmpty()) {
+							IExpression isKnownMember = ExpressionUtil.parse("$0.exists(iu | this.id == iu.id && this.version == iu.version)"); //$NON-NLS-1$
+							IQuery<IInstallableUnit> categoryMembersQuery = QueryUtil.createMatchQuery(isKnownMember, allCategoryMembers);
+							flatQuery = QueryUtil.createCompoundQuery(topLevelQuery, categoryMembersQuery, false);
```
merks: I'm trying to wrap my head around this.  We already have the list of IUs we want, but somehow we need to turn it into a query again because that's what ElementQueryDescriptor needs?

merks (commented):
I need some time to understand this properly; particularly the part where we turn the result we want back into a query...

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
subyssurendran666: Categories are now only pushed for further traversal and never added to the members set.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
subyssurendran666: Yes, agreed. The guard already takes care of it, so there is no need to repeat the check. Thank you for confirming.

bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/QueryProvider.java
subyssurendran666: Simplified it by pre-loading the category members into the collector before running the query. Thank you.

merks (approved):
That's **way** nicer!

PR #1107, opened 2026-07-21
Requirements could not be serialised as LDAP filter strings, which blocked bridging p2 to the OSGi Repository Service.

### LINKED ISSUE #211: LDAP Filter support for requirements
I'm currently working on a bridge from P2 to [Repository Service Specification](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.repository.html)

This seems mostly match quite well except one thing:
The `IExpression.toLDAPString(StringBuffer)` actually throws `UnsupportedOperationException` for the interesting case of `IRequirement.getMatches()`

The reason is that generic expressions do not support LDAP and in 

https://github.com/eclipse-equinox/p2/blob/0dfc4b4b4e0bb917841b967da2ac68ba45e51ad8/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/RequiredCapability.java#L139-L169

there is an "Exits" Expression created. Looking at the passed parameters there we have:

- namespace (e.g. `org.eclipse.equinox.p2.iu`)
- name (e.g. `my.bundle.name`)
- range (e.g. `[0.0.0,1.0.0)`)

what could be transformed into an LDAP as:

`(&(org.eclipse.equinox.p2.iu=my.bundle.name)(version<1.0.0))`

I'm not sure where to handle it best, but it would be great if P2 can support LDAP Filters at least for the common cases and only fails if something can not be expressed (e.g. greedy, min/max), so at least for `RequiredCapability`, but probably  `RequiredPropertiesMatch` as well.

subyssurendran666:
Requirements could not be serialised as LDAP filter strings, which blocked bridging p2 to the OSGi Repository Service. The root cause was that RequiredCapability match expressions store bound values as Parameter placeholder nodes, which the existing Equals/Compare serialisers could not handle.

The fix adds toLDAPString overrides to MatchExpression, Exists, and LambdaExpression. MatchExpression substitutes each Parameter with a Literal before delegating to the existing serialisers; Exists and LambdaExpression simply drop their wrappers.

Fixes: https://github.com/eclipse-equinox/p2/issues/211

### DIFF
--- a/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/expression/Exists.java
+++ b/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/expression/Exists.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2017 Cloudsmith Inc. and others.
+ * Copyright (c) 2009, 2026 Cloudsmith Inc. and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -23,6 +23,11 @@ import org.eclipse.equinox.p2.metadata.expression.IEvaluationContext;
 final class Exists extends CollectionFilter {
 	Exists(Expression collection, LambdaExpression lambda) {
 		super(collection, lambda);
+	}
+
+	@Override
+	public void toLDAPString(StringBuilder buf) {
+		lambda.toLDAPString(buf);
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


merks:
What does `org.eclipse.equinox.p2.metadata.expression.IExpression.toLDAPString(StringBuilder)` return?  If that doesn't work, might it be better to make that work?

subyssurendran666:
You are right. Adding toLDAPString overrides to `Exists`, `LambdaExpression`, and `Parameter` is the cleaner fix. It completes what was already designed without a separate utility method. I'll update the PR accordingly.

subyssurendran666:
@merks could you please re-review whenever this you have a moment? Thanks.

bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/RequiredCapability.java
```diff
@@ -213,4 +213,90 @@ private static void assertVersionRangeRequirement(IMatchExpression<IInstallableU
 			throw new IllegalArgumentException();
 		}
 	}
+
```
merks: This change is supposed to be removed from the PR isn't it?

merks (changes requested):

bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/RequiredCapability.java
subyssurendran666: Done.

merks (approved):
It looks sensible.

PR #1109, opened 2026-07-29
Fix: DownloadStatus should carry information about its URI

### LINKED ISSUE #178: DownloadStatus should carry information about its URL/URI
Currently we have last change, file size and transfer speed, but there is not data that says what URL this actually was recorded for!

subyssurendran666:
DownloadStatus already tracked file size, last-modified timestamp, and transfer rate, but had no field to identify which URL the status was recorded for, making it impossible for callers to correlate a status object back to the download that produced it. A uri field is added with `getUri()` / `setUri(URI)` accessors and included in `toString()`, following the same mutable-setter pattern as the existing fields to avoid any constructor API break. In `RepositoryTransport`, the URI was already flowing through as a parameter at every `DownloadStatus` construction site but was never attached to the resulting object; this fix wires it in across all exit paths — success, cancellation, authentication failure, and retry exhaustion — via thin public wrappers around internal helpers, with the same pattern applied symmetrically to RepositoryStatus. Nine unit tests are added in `DownloadStatusTest` covering the new field, the `toString()` null fallback, and all four factory methods across both classes.

Fix: https://github.com/eclipse-equinox/p2/issues/178

### DIFF
--- a/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/DownloadStatus.java
+++ b/bundles/org.eclipse.equinox.p2.repository/src/org/eclipse/equinox/internal/p2/repository/DownloadStatus.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.equinox.internal.p2.repository;
 
+import java.net.URI;
 import org.eclipse.core.runtime.Status;
 
 /**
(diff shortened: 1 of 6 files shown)


merks (approved):

merks:
@laeubi 

Given you seem to have asked for something like this, do you want to review this?

PR #1110, opened 2026-07-30
Fix #862: Once-a-month auto-update schedule fails with IllegalArgumentException due to int overflow

### LINKED ISSUE #862: Once a month schedule not working due to int overflow in AutomaticUpdateScheduler.computeFuzzyPoll()
We have a catch-all exception handler in our RCP and have had a few reports of an `IllegalArgumentException` on setting the auto-update check schedule, e.g.

```
java.lang.IllegalArgumentException
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor.scheduleAtFixedRate(Unknown Source)
	at java.base/java.util.concurrent.Executors$DelegatedScheduledExecutorService.scheduleAtFixedRate(Unknown Source)
	at a.b.v.b.d.addUpdateCheck(d.java:34)
	at org.eclipse.equinox.internal.p2.ui.sdk.scheduler.AutomaticUpdateScheduler.scheduleUpdate(AutomaticUpdateScheduler.java:156)
	at org.eclipse.equinox.internal.p2.ui.sdk.scheduler.AutomaticUpdateScheduler.rescheduleUpdate(AutomaticUpdateScheduler.java:121)
	at a.b.v.c.c.performOk(c.java:101)
	at org.eclipse.jface.preference.PreferencePage.performApply(PreferencePage.java:454)
```

It's caused by being passed values <= 0. I've finally spotted what the issue is, it's in the calculation for the monthly schedule here:

- https://github.com/eclipse-equinox/p2/blob/1a09efdb39b2946efbacc281baaee76da8f1c323/bundles/org.eclipse.equinox.p2.ui.sdk.scheduler/src/org/eclipse/equinox/internal/p2/ui/sdk/scheduler/AutomaticUpdateScheduler.java#L234

```
	private static final int ONE_HOUR_IN_MS = 60 * 60 * 1000;
	private static final int ONE_DAY_IN_MS = 24 * ONE_HOUR_IN_MS;
```

The method is returning a `long` but the calc is done with `int`s where

```
30 * 24 * 60 * 60 * 1000 = 2,592,000,000
```
which is actually bigger than the int max value of `2,147,483,647` so overflow occurs and a negative result it actually returned.

Perhaps the constants should be changed to `long`s or the calc changed to use `30L`...

subyssurendran666:
Changed `ONE_HOUR_IN_MS` and `ONE_DAY_IN_MS` from `int` to `long` so all time  arithmetic in `computeFuzzyPoll()` and `getMaxDelay()` operates in `long`,  eliminating the overflow for any multiplier. Updated the two `Random.nextInt()`  call sites to `Random.nextLong()` accordingly.

fix: https://github.com/eclipse-equinox/p2/issues/862

### DIFF
--- a/bundles/org.eclipse.equinox.p2.tests.ui/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.p2.tests.ui/META-INF/MANIFEST.MF
@@ -20,6 +20,7 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="3.29.0",
  org.eclipse.equinox.p2.ui.importexport;bundle-version="1.0.1"
 Bundle-RequiredExecutionEnvironment: JavaSE-21
 Import-Package: org.eclipse.equinox.internal.p2.operations,
+ org.eclipse.equinox.internal.p2.ui.sdk.scheduler,
  org.eclipse.equinox.p2.operations;version="[2.0.0,3.0.0)",
  org.mockito,
  org.mockito.stubbing
--- a/bundles/org.eclipse.equinox.p2.tests.ui/src/org/eclipse/equinox/p2/tests/ui/misc/AllTests.java
+++ b/bundles/org.eclipse.equinox.p2.tests.ui/src/org/eclipse/equinox/p2/tests/ui/misc/AllTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2008, 2018 IBM Corporation and others.
+ *  Copyright (c) 2008, 2026 IBM Corporation and others.
  *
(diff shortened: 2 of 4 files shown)


merks (changes requested):
This looks fine, but it links to an unrelated issue.  Better to update the commit message to remove the issue reference or to correct it to the matching issue if there is one.

subyssurendran666:
> This looks fine, but it links to an unrelated issue. Better to update the commit message to remove the issue reference or to correct it to the matching issue if there is one.

Thank you, Ed Merks, for pointing that out. I have corrected the incorrect URL in both the commit message and the PR description.

merks (approved):

PR #1112, opened 2026-08-01
Make IU list columns sortable in p2 provisioning UI

### LINKED ISSUE #1111: Add column sorting to the Available Software and Installed Software plugin lists
The Available Software and Installed Software dialogs currently do not support sorting by column headers. Enabling column sorting would make it easier for users to browse, locate, and compare plugins, especially when working with a large number of installed or available entries.

Similar functionality is already available in the Import Software Configuration and Available Software Sites dialogs. Providing the same capability here would offer a more consistent user experience across the p2 UI.

Expected behavior
----------------

- Users should be able to sort the plugin list by clicking on a column header.
- Clicking the same column header again should toggle the sort order (ascending/descending).
- The sorting behavior should be consistent with other p2 dialogs that already support column sorting.

Image

Image

subyssurendran666:
Column headers in the Available Software, Installed Software, and Install/Update resolution results pages are now clickable. Clicking a header sorts by that column, and clicking again toggles between ascending and descending order.

Fix: https://github.com/eclipse-equinox/p2/issues/1111

image

image

image

https://github.com/user-attachments/assets/dc589841-2ec8-48ae-ab7f-98a275d64ca4

### DIFF
--- a/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/AvailableIUGroup.java
+++ b/bundles/org.eclipse.equinox.p2.ui/src/org/eclipse/equinox/internal/p2/ui/dialogs/AvailableIUGroup.java
@@ -172,7 +172,7 @@ public class AvailableIUGroup extends StructuredIUGroup {
 		availableIUViewer.setContentProvider(contentProvider);
 
 		// Now the presentation, columns before label provider.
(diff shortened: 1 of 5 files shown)


merks (approved):
It's cool.  I like it.

subyssurendran666:
Thanks for reviewing the PR.

PR #1114, opened 2026-08-02
Repository Location field should suggest previously known sites

### LINKED ISSUE #1113: Add autocomplete support for repository URLs in the Add/Edit Repository dialogs
The Location field in the Add Repository and Edit Repository dialogs does not provide autocomplete suggestions for previously used repository URLs.

In contrast, the Work with: field on the Available Software page already suggests known repository URLs, including repositories that have been disabled.

It would be helpful if the Location field in the Add and Edit Repository dialogs offered the same autocomplete experience. This would save users from having to remember or retype long repository URLs when adding a new repository or editing an existing one.

Implementing this in the common RepositoryNameAndLocationDialog would automatically improve the Add and Edit Repository dialogs used from both the Available Software page and the Available Software Sites page, providing a more consistent user experience across the p2 UI.

subyssurendran666:
When adding or editing a repository, the Location field offered no suggestions, requiring users to remember and retype repository URLs manually. This change adds content assist to RepositoryNameAndLocationDialog.createLocationField() using a new TextAutoCompleteField helper.
The content assist suggests previously disabled repository URLs, making it easier to re-add repositories that are no longer discoverable through the existing UI. Implementing this in RepositoryNameAndLocationDialog automatically covers the Add Repository, Edit Repository.

Fixes: https://github.com/eclipse-equinox/p2/issues/1113

image

image

### DIFF
--- a/bundles/org.eclipse.equinox.p2.tests.ui/src/org/eclipse/equinox/p2/tests/ui/dialogs/AddRepositoryDialogTest.java
+++ b/bundles/org.eclipse.equinox.p2.tests.ui/src/org/eclipse/equinox/p2/tests/ui/dialogs/AddRepositoryDialogTest.java
@@ -1,0 +1,103 @@
+/*******************************************************************************
+ *  Copyright (c) 2026 IBM Corporation and others.
+ *
+ *  This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License 2.0
+ *  which accompanies this distribution, and is available at
+ *  https://www.eclipse.org/legal/epl-2.0/
+ *
+ *  SPDX-License-Identifier: EPL-2.0
+ *
+ *  Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.equinox.p2.tests.ui.dialogs;
+
+import java.net.URI;
+import java.util.Arrays;
+import org.eclipse.core.runtime.URIUtil;
+import org.eclipse.equinox.internal.p2.ui.dialogs.RepositoryNameAndLocationDialog;
+import org.eclipse.equinox.internal.p2.ui.dialogs.TextAutoCompleteField;
+import org.eclipse.equinox.p2.tests.ui.AbstractProvisioningUITest;
+import org.eclipse.jface.fieldassist.IContentProposal;
+import org.eclipse.swt.widgets.Shell;
+
+/**
+ * Tests that {@link RepositoryNameAndLocationDialog} wires autocomplete
+ * proposals from the repository manager into its Location field.
+ */
+public class AddRepositoryDialogTest extends AbstractProvisioningUITest {
+
+	/**
+	 * Dialog subclass that exposes the protected autocomplete field for testing.
+	 */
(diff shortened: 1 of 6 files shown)


merks:
Just reading on my phone, it sounds like add will suggest repositories that are already present and don’t need to be added because they are already present.  Maybe I overlooked something?

subyssurendran666:
> Just reading on my phone, it sounds like add will suggest repositories that are already present and don’t need to be added because they are already present. Maybe I overlooked something?

That is a good point. You are right that re-adding an already enabled repository would effectively be a no-op.

The primary use case I had in mind was adding a new repository URL that's similar to an existing one. For example, if `releases/2024-06` is already known, and the user wants to add `releases/2026-06`, they can select the existing URL from the suggestions and simply update the version suffix instead of typing the entire URL again.

I was thinking of it as being similar to a browser's address bar, where previously used URLs are suggested even if the user intends to navigate to a slightly different one.

merks:
I find the use case not really so strong.  Here the combo box already includes the available repositories.

image

And I can do my editing in place after picking one from the list:

image

That Add... button doesn't help need helpful additions. 

So in general this seems marginally useful.

---

Of course it doesn't help in the Eclipse SDK, but the Repository Explorer is in every EPP package and it provides access to the indexed repositories available on the download server:

image

subyssurendran666:
Thank you for the detailed explanation. I hadn't considered that the "Work with:" combo already supports editing the selected URL in place. You are right that it already covers the URL variation scenario I had in mind.

I agree that including enabled repositories does not seem to add much value. I will update the implementation to suggest only disabled repositories, since those are not discoverable through the existing UI. That keeps the feature focused on the remaining gap. I will update the patch accordingly.

subyssurendran666:
I have updated the implementation. Now, only disabled repository URLs are shown in the autocomplete suggestions.

HannesWell:
> Now, only disabled repository URLs are shown in the autocomplete suggestions.

I have to admit that a I'm not sure if this relatively large change carries it's own weight by providing assistance for a very specific use-case.

Furthermore did you write the tests yourself or are they AI generated? In case of the latter, please check if they are really useful. AI often generates tests that are basically getter-tests.

subyssurendran666:
Thank you for the feedback. I revisited the tests and removed a few that did not add much value.

I should also mention that I used AI to help write the tests by providing it with the exact scenarios I wanted to cover. I reviewed the generated tests before including them, but your comment made me take another look, and I realised a few of them were indeed unnecessary. I have removed those and retained only the ones that validate the intended behaviour.

You are right that the use case is quite specific. The intention was to help users who had previously added a repository, later disabled it, and now want to add it again. Since disabled repositories are no longer available through the "Work with:" combo, and the Repository Explorer is intended for discovering repositories rather than recalling previously disabled ones, I thought this would help close that small gap.

That said, I agree it's a fairly narrow use case. If you feel the feature does not justify the additional code, I'm perfectly happy to drop the change.

merks:
I would prefer we don't add complexity to support this.  If one disables an update site, it's even less likely one wants to use it, and of course one can easily enable it.

subyssurendran666:
Thank you for the review and the feedback. I understand your reasoning. I will close the PR.

